PR #4, opened 2020-01-28
Add send_file warning

ulziibay:
Adds a check when there might be a dataflow from user input to a sink.
"How to serve static files in Flask" is the most asked question in Stackoverflow for Flask. Hence, this will be a popular check if we do it right.

Pending `def $X(filename: $Y, ...):` support integration in https://github.com/returntocorp/sgrep. 

Testing: 
1) Run rule 
```
sgrep-lint ~/Workspace/sgrep-rules/python/flask/secure-static-file-serve.yaml ~/Workspace/sgrep-rules/tests/python/flask/send_static_file/test_send_file_without_path_sanitization.py| jq
```
2) See it fire on 
```
@app.route("/<path:filename>")
def download_file(filename):
  return send_file(filename) .  ## CATCHES THIS
```

### DIFF
--- a/python/flask/secure-static-file-serve.yaml
+++ b/python/flask/secure-static-file-serve.yaml
@@ -3,9 +3,12 @@
 # https://stackoverflow.com/questions/20646822/how-to-serve-static-files-in-flask
 rules:
   - id: avoid_send_file_without_path_sanitization
-    pattern: |
-      def $X():
-        ...
+    patterns:
+      - pattern-inside: |
+          def $X(filename):
+            ...
+      - pattern-either:
+        - pattern: flask.send_file(filename, ...)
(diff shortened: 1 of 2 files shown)


python/flask/secure-static-file-serve.yaml
```diff
@@ -0,0 +1,14 @@
+# For Flask app to serve file securely, it should use `send_from_directory` method
+# The other methods like `send_file` or `send_static_file` are not recommended per
+# https://stackoverflow.com/questions/20646822/how-to-serve-static-files-in-flask
+rules:
+  - id: avoid_send_file_without_path_sanitization
+    patterns:
+      - pattern-inside: |
+          def $X(filename):
```
ievans: @ulziibay shouldn't filename be a metavariable, eg `$F`?

ievans: Ideally we would have decorator support here https://github.com/returntocorp/sgrep/issues/29

PR #25, opened 2020-02-03
javascript tests

ievans:
Initial version of javascript support in the test suite and one javascript test.


javascript/smells/leftover_debugging.yaml
```diff
@@ -0,0 +1,29 @@
+rules:
+  - id: javascript-alert
+    patterns: 
+        - pattern-either: 
+          - pattern: alert()
+          - pattern: alert($X)
```
aryx: you can use alert(...)
will match 0 or n times.

aryx: Or do you just want to check for call arities of 0 and 1?

ievans: Yup, arities 0 and 1

ievans:
Fixed sgrep in https://github.com/returntocorp/sgrep/pull/80/

PR #28, opened 2020-02-04
Disable bad rules with return pattern

ulziibay:
Due to https://github.com/returntocorp/sgrep/issues/73, disabling the following two rules!

With these disabled, sgrep able to run without erroring out on sgrep-rules repo
```
ulziibayarotgonbaatar@ulziibayars-MacBook-Pro sgrep % sgrep-lint -f ~/Workspace/sgrep-rules/ ~/Workspace/pyyaml 

/Users/ulziibayarotgonbaatar/Workspace/sgrep-rules/.github/workflows/sgrep-rules-test.yml is missing `rules` as top-level key
/Users/ulziibayarotgonbaatar/Workspace/sgrep-rules/.circleci/config.yml is missing `rules` as top-level key
running 59 rules from 25 yaml files (2 yaml files were invalid)
{"results" : OMITTED}
```


python/deadcode/return.py
```diff
@@ -1,6 +1,5 @@
 
-# ruleid: return-not-in-function
```
ievans: instead of commenting the code out, you can changes this to # todoruleid:  return-not-in-function

PR #29, opened 2020-02-04
use the new decorator support to pass more tests

ievans:


python/flask/secure-static-file-serve.py
```diff
@@ -6,3 +6,7 @@
 def download_file(filename):
   # ruleid:avoid_send_file_without_path_sanitization
```
ulziibay: You will need to disable this or add decorator

ievans: @ulziibay there is a decorator -- just expand the diff :)

ievans:
It is expected that sgrep master will fail as these changes are only on sgrep develop.

PR #30, opened 2020-02-04
Check for Flask request access outside in GET requests

ulziibay:
Adds sgrep checks for accessing request.form, request.json, request.data inside GET handler
This could lead to badreques exception if not caught.


python/flask/access-request-in-wrong-handler.yaml
```diff
@@ -0,0 +1,43 @@
+rules:
+  - id: avoid-accessing-request-in-wrong-handler
+    patterns:
+      - pattern-either:
+        - pattern: |
+            @app.route(..., method="GET")
+            def $X(...):
+              $Y = flask.request.json
+              ...
+              return
```
ievans: why do you want to match on `return` (also I think you want `return ...`)? Would you rather have:

```
def $X(...):
    ...
    $Y = flask.request.json
    ...
```

python/flask/access-request-in-wrong-handler.yaml
```diff
@@ -0,0 +1,43 @@
+rules:
+  - id: avoid-accessing-request-in-wrong-handler
+    patterns:
+      - pattern-either:
```
ievans: I think this would be simplified greatly by using 
```
- pattern-inside: |
   @app.route(..., method="GET")
    def $X(...):
        ...
```
followed by `pattern-either`

ulziibay: Sounds good

python/flask/access-request-in-wrong-handler.yaml
ulziibay: I believe that with 
```
    $Y = flask.request.json
    ...
```
the pattern was matching the entire python file. return was like a stop canary.

ulziibay: This is addressed with the following comment

PR #36, opened 2020-02-04
Add wtf-python string comparison using 'is'

DrewDennison:
Triaged in: https://dev.massive.ret2.co/jobs/550


ievans:
I'm surprised this is working as it uses the deprecated `ok` severity. @aryx did you change that?

minusworld:
Thanks! 🙏

PR #48, opened 2020-02-06
check for pass in for loop or body of a function

DrewDennison:


python/smells/pass-body.yaml
```diff
@@ -0,0 +1,15 @@
+rules:
+  - id: pass-body-fn
+    pattern: |
+      def $X(...):
+        pass
+    message: "`pass` is the body of function $X. Consider removing this"
```
ievans: maybe suggest "raise NotImplementedError()" if the method is a TODO

PR #52, opened 2020-02-07
Add rule for detecting file objects opened without corresponding close

mschwager:


python/smells/open-never-closed.yml
```diff
@@ -0,0 +1,47 @@
...
+            $F.close()
+        - pattern-not-inside: |
+            $F = tempfile.SpooledTemporaryFile(...)
+            ...
+            $F.close()
+        - pattern-either:
+            - pattern: $F = open(...)
+            - pattern: $F = io.open(...)
+            - pattern: $F = os.open(...)
+            - pattern: $F = tarfile.open(...)
```
ievans: I think you are missing a  `pattern-not-inside` for `tarfile.open(...)` case

mschwager: Hmm, I see it above as the 4th `pattern-not-inside`

PR #69, opened 2020-02-14
add useless comparison check to strings

ievans:


aryx:
I think it would have been good to also add
if "..." or ... in ...:
  ...
pattern so that we can give a more precise error message "The priority of or is higher than the priority of in, this expression probably does not do what you mean"

PR #179, opened 2020-02-25
Django Field setting null=True checks

minusworld:
Three checks:
- per [documentation](https://docs.djangoproject.com/en/3.0/ref/models/fields/#django.db.models.Field.null), discourage null=True in string fields
- per [documentation](https://docs.djangoproject.com/en/3.0/ref/models/fields/#django.db.models.Field.null), require null=True when blank=True and unique=True in string fields
- per check recommendation by Django creator, require null=True when blank=True on non-text fields


DrewDennison:
another one @jacobian

python/django/string-field-null-checks-import.py
```diff
@@ -0,0 +1,32 @@
+from django.db import models
+from django.db.models import Model, CharField, IntegerField
+
+class FakeModel(Model):
```
nbrahms: I'd expect tests for `TextField` too

python/django/string-field-null-checks.yml
```diff
@@ -0,0 +1,28 @@
+rules:
+  - id: no-null-string-field
+    patterns:
```
nbrahms: Do you want to add `pattern-inside` and assignment checks here?

minusworld: 👍

python/django/string-field-null-checks-import.py
minusworld: 👍 added

PR #186, opened 2020-02-25
add simple csrf_exempt example

### LINKED ISSUE #82: avoid csrf_exempt decorator
only very occasionally is this ok

DrewDennison:
Fixes #82 (simplistic)

### DIFF
--- a/python/django/security/csrf_exempt.yaml
+++ b/python/django/security/csrf_exempt.yaml
@@ -1,0 +1,12 @@
+rules:
+  - id: django-csrf_exempt
+    pattern: |
+      from django.views.decorators.csrf import csrf_exempt
+      ...
+
+      @csrf_exempt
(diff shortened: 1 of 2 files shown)


python/django/security/csrf_exempt.yaml
```diff
@@ -0,0 +1,12 @@
+rules:
+  - id: django-csrf_exempt
+    pattern: |
+      from django.views.decorators.csrf import csrf_exempt
+      ...
+
+      @csrf_exempt
```
minusworld: Would import resolution work here, too?

e.g., 
`@django.views.decorators.csrf.csrf_exempt`

DrewDennison: I tried that in the rule but let me make this an or case

PR #187, opened 2020-02-25
fixup: csrf_exempt for full-qualified import names

DrewDennison:


DrewDennison:
@minusworld I think this works a bit better and addresses your concern but I think import resolution in sgrep is still pretty brittle in general

PR #189, opened 2020-02-25
use strict parsing flag

ievans:
enable flags merged in https://github.com/returntocorp/sgrep/pulls/194

### DIFF
--- a/test.py
+++ b/test.py
@@ -158,7 +158,7 @@ def generate_file_pairs(
             cmd = (
                 ["sgrep-lint"]
                 + extra_args
-                + ["--strict", "--json", "--no-rewrite-rule-ids", "-f", str(filename)]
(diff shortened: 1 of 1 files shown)


test.py
```diff
@@ -158,7 +158,7 @@ def generate_file_pairs(
             cmd = (
                 ["sgrep-lint"]
                 + extra_args
-                + ["--strict", "--json", "--no-rewrite-rule-ids", "-f", str(filename)]
+                + ["--strict", "--strict-parsing", "--json", "--no-rewrite-rule-ids", "-f", str(filename)]
```
nbrahms: what does this change?

ievans: If sgrep has a parse error on a code file, under this flag it will fail rather than silently skipping that file. This means our test suite will now fail if someone tries to commit a file sgrep can't parse to this repo.

PR #191, opened 2020-02-25
python is-not-is-not check

brendongo:
In python `X is not ...` is different from `X is (not ...)`
in the latter the not converts the ... to boolean.


python/wtf-python/is-not-is-not.yml
```diff
@@ -0,0 +1,7 @@
+rules:
+  - id: is-not-is-not
+    patterns:
+      - pattern: $S is (not ...)
+    message: "is (not ...) probably not what you want to do"
```
minusworld: The description in the commit PR is so good! Maybe you could add that explanation here? 😄

nbrahms: I second this :)

minusworld:
https://dev.massive.ret2.co/triager/filter/631 - no findings

PR #192, opened 2020-02-26
Upsell count() method

ulziibay:
Addresses https://github.com/returntocorp/sgrep-rules/issues/97
Test with:
```
docker run --rm -v $(pwd):/home/repo returntocorp/sgrep:develop --config python/django/db/upsell_count.yaml python/django/db/upsell_count.py
```

### DIFF
--- a/python/django/db/upsell_count.yaml
+++ b/python/django/db/upsell_count.yaml
@@ -5,6 +5,6 @@ rules:
         - pattern: $X.objects.$FUNC(...).len()
         - pattern: $X.objects.$FUNC(...).$FILTER().len()
         - pattern: $X.objects.$FUNC(...).$FILTER().$UPDATE(...).len()
-    message: "Looks like you need to determine the number of records. Django provides efficient `count()` method for precisely this reason. See https://docs.djangoproject.com/en/3.0/ref/models/querysets/"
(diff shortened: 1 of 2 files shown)


python/django/db/upsell_count.yaml
```diff
@@ -0,0 +1,10 @@
+rules:
+  - id: use-count-method
+    patterns:
+      - pattern-either:
+        - pattern: $X.objects.$FUNC(...).len()
+        - pattern: $X.objects.$FUNC(...).$FILTER().len()
+        - pattern: $X.objects.$FUNC(...).$FILTER().$UPDATE(...).len()
+    message: "Looks like you need to determine the number of records. Django provides efficient `count()` method for precisely this reason. See https://docs.djangoproject.com/en/3.0/ref/models/querysets/"
```
minusworld: Suggestion: "Looks like you need to determine the number of records. Django provides the `count()` method which is more efficient than `.len()`.  See https://docs.djangoproject.com/en/3.0/ref/models/querysets/"

ulziibay: ok addressed

minusworld:
cf. https://stackoverflow.com/questions/14327036/count-vs-len-on-a-django-queryset

PR #196, opened 2020-02-26
check for connections over HTTP or disabling certs with HTTPS (superset of DUO132)

ievans:
Requires https://github.com/returntocorp/sgrep/pull/198

### DIFF
--- a/python/certificates/disabled-cert-validation.yaml
+++ b/python/certificates/disabled-cert-validation.yaml
@@ -1,0 +1,22 @@
+rules:
+    - id: disabled-cert-validation
+      patterns:
+        - pattern-either:
+            - pattern: urllib3.PoolManager(..., cert_reqs=$REQS, ...)
+            - pattern: urllib3.ProxyManager(..., cert_reqs=$REQS, ...)
+            - pattern: urllib3.HTTPSConnectionPool(..., cert_reqs=$REQS, ...)
+            - pattern: urllib3.connection_from_url(..., cert_reqs=$REQS, ...)
+            - pattern: urllib3.proxy_from_url(..., cert_reqs=$REQS, ...)                
+            - pattern: ssl.wrap_socket(..., cert_reqs=$REQS, ...)
+        - pattern-where-python: "vars['$REQS'] in ['NONE', 'CERT_NONE', 'ssl.CERT_NONE']"
+      message: "certificate verification explicitly disabled, insecure connections possible"
+      languages: [python]
+      severity: WARNING
+    - id: disabled-cert-validation-requests
+      patterns:
+        - pattern-either:
+            - pattern: requests.get(..., verify=False, ...)
+            - pattern: requests.post(..., verify=False, ...)
+      message: "certificate verification explicitly disabled, insecure connections possible"
+      languages: [python]
+      severity: WARNING
--- a/python/certificates/disabled-cert-validation.yaml
+++ b/python/certificates/disabled-cert-validation.yaml
@@ -13,8 +13,7 @@ rules:
         - pattern-where-python: |-
             vars['$REQS'] in ['NONE', 'CERT_NONE', 'ssl.CERT_NONE', 'ssl.CERT_OPTIONAL', 'CERT_OPTIONAL',
                               "'NONE'", "'CERT_NONE'", "'ssl.CERT_NONE'", "'ssl.CERT_OPTIONAL'", "'CERT_OPTIONAL'",
-                              '"NONE"', '"CERT_NONE"', '"ssl.CERT_NONE"', '"ssl.CERT_OPTIONAL"', '"CERT_OPTIONAL"']
-          
+                              '"NONE"', '"CERT_NONE"', '"ssl.CERT_NONE"', '"ssl.CERT_OPTIONAL"', '"CERT_OPTIONAL"']          
       message: "certificate verification explicitly disabled, insecure connections possible"
       languages: [python]
       severity: WARNING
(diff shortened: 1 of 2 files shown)


python/certificates/disabled-cert-validation.yaml
```diff
@@ -0,0 +1,35 @@
...
+            - pattern: urllib3.ProxyManager(..., cert_reqs=$REQS, ...)
+            - pattern: urllib3.HTTPSConnectionPool(..., cert_reqs=$REQS, ...)
+            - pattern: urllib3.connectionpool.HTTPSConnectionPool(..., cert_reqs=$REQS, ...)
+            - pattern: urllib3.connection_from_url(..., cert_reqs=$REQS, ...)
+            - pattern: urllib3.proxy_from_url(..., cert_reqs=$REQS, ...)
+            - pattern: $CONTEXT.wrap_socket(..., cert_reqs=$REQS, ...)
+            - pattern: ssl.wrap_socket(..., cert_reqs=$REQS, ...)
+        - pattern-where-python: |-
+            vars['$REQS'] in ['NONE', 'CERT_NONE', 'ssl.CERT_NONE', 'ssl.CERT_OPTIONAL', 'CERT_OPTIONAL',
+                              "'NONE'", "'CERT_NONE'", "'ssl.CERT_NONE'", "'ssl.CERT_OPTIONAL'", "'CERT_OPTIONAL'",
```
mschwager: I believe the string literal module paths aren't valid arguments to pass, i.e. `"'ssl.CERT_NONE'"`, `"'ssl.CERT_OPTIONAL'"`, `'"ssl.CERT_NONE"'`, and `'"ssl.CERT_OPTIONAL"'`. The implementation uses `getattr` if it's a string, which would fail on a module path:

```python
In [10]: urllib3.util.ssl_.resolve_cert_reqs??                                                          
Signature: urllib3.util.ssl_.resolve_cert_reqs(candidate)
Source:   
def resolve_cert_reqs(candidate):
    """
    Resolves the argument to a numeric constant, which can be passed to
    the wrap_socket function/method from the ssl module.
    Defaults to :data:`ssl.CERT_NONE`.
    If given a string it is assumed to be the name of the constant in the
    :mod:`ssl` module or its abbreviation.
    (So you can specify `REQUIRED` instead of `CERT_REQUIRED`.
    If it's neither `None` nor a string we assume it is already the numeric
    constant which can directly be passed to wrap_socket.
    """
    if candidate is None:
        return CERT_NONE

    if isinstance(candidate, str):
        res = getattr(ssl, candidate, None)
        if res is None:
            res = getattr(ssl, 'CERT_' + candidate)
        return res

    return candidate
File:      ~/.local/lib/python3.6/site-packages/urllib3/util/ssl_.py
Type:      function
```

And dot resolution doesn't work in `getattr`:

```python
In [15]: class Foo:
    ...:     pass 
    ...:                                                                                                

In [16]: class Bar:
    ...:     pass 
    ...:                                                                                                

In [17]: class Baz:
    ...:     pass 
    ...:                                                                                                

In [18]: baz = Baz()

In [19]: baz.foo = Foo()

In [20]: baz.foo.bar = Bar()

In [22]: getattr(baz, 'foo', 'NONE')
Out[22]: <__main__.Foo at 0x7f8fb7d72fd0>

In [23]: getattr(baz, 'foo.bar', 'NONE')
Out[23]: 'NONE'
```

However, we can also search for the string literal `'OPTIONAL'` like we do `'NONE'`.

python/certificates/disabled-cert-validation.yaml
```diff
@@ -0,0 +1,35 @@
...
+      patterns:
+        - pattern-either:
+            - pattern: urllib3.HTTPConnectionPool(...)
+            - pattern: urllib3.connectionpool.HTTPConnectionPool(...)
+      message: "suggest HTTPS over HTTP connection"
+      languages: [python]
+      severity: WARNING
+    - id: disabled-cert-validation-requests
+      patterns:
+        - pattern-either:
```
mschwager: There are a few more HTTP methods `requests` makes available (although they're likely less common) if you'd like to cover all of them: https://github.com/dlint-py/dlint/blob/master/dlint/linters/bad_requests_use.py

python/certificates/disabled-cert-validation.yaml
ievans: Great point @mschwager  -- addressed!

python/certificates/disabled-cert-validation.yaml
ievans: Thanks for the link! I added them as well

python/certificates/disabled-cert-validation.yaml
```diff
@@ -0,0 +1,40 @@
...
+            - pattern: urllib3.PoolManager(..., cert_reqs=$REQS, ...)
+            - pattern: urllib3.ProxyManager(..., cert_reqs=$REQS, ...)
+            - pattern: urllib3.HTTPSConnectionPool(..., cert_reqs=$REQS, ...)
+            - pattern: urllib3.connectionpool.HTTPSConnectionPool(..., cert_reqs=$REQS, ...)
+            - pattern: urllib3.connection_from_url(..., cert_reqs=$REQS, ...)
+            - pattern: urllib3.proxy_from_url(..., cert_reqs=$REQS, ...)
+            - pattern: $CONTEXT.wrap_socket(..., cert_reqs=$REQS, ...)
+            - pattern: ssl.wrap_socket(..., cert_reqs=$REQS, ...)
+        - pattern-where-python: |-
+            vars['$REQS'] in ['NONE', 'CERT_NONE', 'ssl.CERT_NONE', 'ssl.CERT_OPTIONAL', 'CERT_OPTIONAL',
```
mschwager: I don't think the `'NONE'` check is valid. I believe that's looking for `ssl.NONE`, which isn't an attribute:

```python
----> 1 ssl.NONE

AttributeError: module 'ssl' has no attribute 'NONE'
```

ievans: It's *technically* not, but I left it in in case someone makes a variable named NONE and passes it in. Maybe a wrong decision.

PR #197, opened 2020-02-26
Upsell latest/earliest

ulziibay:
Addresses https://github.com/returntocorp/sgrep-rules/issues/92

Test with:

```
docker run --rm -v $(pwd):/home/repo returntocorp/sgrep:develop --config python/django/performance/upsell_earliest_latest.yaml python/django/performance 
```

### DIFF
--- a/python/django/performance/upsell_earliest_latest.py
+++ b/python/django/performance/upsell_earliest_latest.py
@@ -1,0 +1,8 @@
+# ruleid:use-earliest-or-latest
+Entry.objects.order_by()[0]
+
+# ruleid:use-earliest-or-latest
+Entry.objects.all().order_by('foo')[0]
+
+# ruleid:use-earliest-or-latest
+Entry.objects.all().filter().order_by('foo')[0]
--- a/python/django/performance/upsell_earliest_latest.py
+++ b/python/django/performance/upsell_earliest_latest.py
@@ -1,8 +1,18 @@
 # ruleid:use-earliest-or-latest
 Entry.objects.order_by()[0]
+# OK
+Entry.objects.order_by()[1]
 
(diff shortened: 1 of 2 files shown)


python/django/performance/upsell_earliest_latest.py
```diff
@@ -0,0 +1,8 @@
+# ruleid:use-earliest-or-latest
```
nbrahms: Please add negative test cases as well

python/django/performance/upsell_earliest_latest.yaml
```diff
@@ -0,0 +1,10 @@
+rules:
+  - id: use-earliest-or-latest
+    patterns:
+      - pattern-either:
+        - pattern: $X.objects.order_by(...)[0]
+        - pattern: $X.objects.$FUNC(...).order_by(...)[0]
+        - pattern: $X.objects.$FUNC(...).$FILTER(...).order_by(...)[0]
+    message: "Looks like you are only accessing first element of ordered QuerySet. Use `latest()` or `earliest()` instead. See https://docs.djangoproject.com/en/3.0/ref/models/querysets/#django.db.models.query.QuerySet.latest"
```
nbrahms:
```suggestion
    message: "Looks like you are only accessing first element of an ordered QuerySet. Use `latest()` or `earliest()` instead. See https://docs.djangoproject.com/en/3.0/ref/models/querysets/#django.db.models.query.QuerySet.latest"
```

python/django/performance/upsell_earliest_latest.py
ulziibay: Added a few. LMK if you think of more!

PR #202, opened 2020-02-27
Open Redirects in Flask

minusworld:
This check looks for data going from `requests.*` into `flask.redirect` implying user control of where to redirect.

https://dev.massive.ret2.co/jobs/587?page=1&sort_on=repo_url&sort_by=asc

### DIFF
--- a/python/flask/open-redirect/open-redirect.yml
+++ b/python/flask/open-redirect/open-redirect.yml
@@ -13,6 +13,10 @@ rules:
         - pattern: |
             $V = flask.request.$W.get($X)
             ...
+            $Z = flask.redirect($V)
+        - pattern: |
+            $V = flask.request.$W.get($X)
+            ...
             return flask.redirect($V)
         - pattern: |
             $V = flask.request.$W[$X]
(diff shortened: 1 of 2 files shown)


python/flask/open-redirect/open-redirect.yml
```diff
@@ -0,0 +1,59 @@
...
+            flask.redirect($V)
+        - pattern: |
+            $V = flask.request.$W
+            ...
+            $Z = flask.redirect($V)
+        - pattern: |
+            $V = flask.request.$W
+            ...
+            return flask.redirect($V)
+    message: "Found data from flask.request passed in to flask.redirect. This is an open redirect and could be exploited"
```
ulziibay: It took me second to decypher the flask.request passed in to flask.redirect sentence. Does it make sense to just say:
```
Data from request is passed to redirect(). This is an open redirect and could be exploited. See <LINK>
```

minusworld: That's much better! Thanks!

PR #203, opened 2020-02-27
New Check for Flask: Rendering string-formatted templates

minusworld:
Added check for rendering template strings that have been string formatted or appended to.


minusworld:
https://dev.massive.ret2.co/triager/filter/585

PR #209, opened 2020-02-28
Add python mutable default datastructure checks

brendongo:
In python, function definitions with parameters set to have a default
value of `[]` or `{}` actually have this datastructure instantiated at
start up time and all calls to the function with the default value used
all point to the same location in memory as the startup instantiated object.

This usually results in unintuitive behavior of changes to the default
object persisting over multiple calls to the function. i.e.:

```
    def f(x=[]):
            x.append(1)
            print(x)
    f()  # [1]
    f()  # [1, 1]
```

This adds test cases of this phenomenon and instances when the mutable
structure is appropriately handled/copied. Also does a first pass effort
to include some patterns for sgrep to find instances of default datastructures
being mutated.


brendongo:
Waiting on https://github.com/returntocorp/sgrep/issues/229

python/wtf-python/default-mutable-dict.yaml
```diff
@@ -0,0 +1,184 @@
...
+                $D = $D or {}
+                ...
+                $D.setdefault(...)
+          - pattern-not-inside: |
+              def $A(...):
+                ...
+                def $F(..., $D={}, ...):
+                  ...
+                  $D.setdefault(...)
+    message: "Function $F mutates default dict $D"
```
nbrahms: I'm not sure if people will immediately see _why_ this is bad. I might suggest adding:

> Default function arguments in Python are shared between function calls. Mutating a default function argument changes the result for all function consumers. This can cause unexpected results, or lead to security vulnerabilities whereby one function consumer can either view or modify the data of another function consumer.

brendongo: Even this is not enough IMO. We need to link to a blog post with examples. Do we want to wordsmith error messages here or once a check has been brought into the golden set?

minusworld: Typically what I'll do is get the check message close to good enough, because we might forget to edit the check messages before they get shipped 🤷‍♂️ . No need to have any back-and-forths at this stage though

brendongo: Made the suggested change.

ulziibay:
This seems to be one of the most verbose and well written sgrep checks to date! Huge ❤️for raising the bar on sgrep check quality!

PR #212, opened 2020-03-03
Django/open redirect

minusworld:
addresses https://github.com/returntocorp/sgrep-rules/issues/83


minusworld:
cf. https://stackoverflow.com/questions/23233683/what-is-the-best-way-to-avoid-open-redirects-in-django

PR #217, opened 2020-03-04
Copy secure-set-cookie to django

nbrahms:
This was a relatively successful check in our flask corpus. I've ported
it to Django here.

Interestingly, there's not a good way to identify that the set_cookie
method is being called on an HttpResponse object without some
rudimentary typing or dataflow. So I just use the common identifier
"response" as the invocation target. Note that this means that this
check fires on Flask projects.

In order to have this not fire on random Django middleware vendored
code, I excluded expressions from common Django middlewares.

After doing this, the fire frequency on non-test, non-vendored code is
1 finding in the entire Django-500 corpus. It's a true positive.

For details, see:
https://dev.massive.ret2.co/jobs/618?page=1&sort_on=repo_url&sort_by=asc

### DIFF
--- a/python/django/security/secure-cookies.py
+++ b/python/django/security/secure-cookies.py
@@ -1,0 +1,57 @@
+from django.shortcuts import render
+
+from django.conf import settings
+from django.contrib.messages.storage.base import BaseStorage, Message
+
+class CookieStorage(BaseStorage):
+    def _update_cookie(self, encoded_data, response):
+        """
+        Either sets the cookie with the encoded data if there is any data to
+        store, or deletes the cookie.
+        """
+        if encoded_data:
+            # ok
+            response.set_cookie(
+                self.cookie_name, encoded_data,
+                domain=settings.SESSION_COOKIE_DOMAIN,
+                secure=settings.SESSION_COOKIE_SECURE or None,
+                httponly=settings.SESSION_COOKIE_HTTPONLY or None,
+            )
+        else:
+            response.delete_cookie(self.cookie_name, domain=settings.SESSION_COOKIE_DOMAIN)
+
+
+def index(request, template):
+    response = render(request, template)
+
+    # ok
+    response.set_cookie("hello", "world", secure=True, httponly=True, samesite="Lax")
+
+    # ok
+    response.set_cookie("hello", "world", **kwargs)
+
+    # ok
+    response.set_cookie(
+        settings.SESSION_COOKIE_NAME,
+        request.session.session_key, max_age=max_age,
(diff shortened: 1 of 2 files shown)


nbrahms:
Closes https://github.com/returntocorp/sgrep-rules/issues/216

python/django/security/secure-cookies.py
```diff
@@ -0,0 +1,57 @@
...
+def index(request, template):
+    response = render(request, template)
+
+    # ok
+    response.set_cookie("hello", "world", secure=True, httponly=True, samesite="Lax")
+
+    # ok
+    response.set_cookie("hello", "world", **kwargs)
+
+    # ok
```
minusworld: Should this one fire because `samesite` is not set?

python/django/security/secure-cookies.py
```diff
@@ -0,0 +1,57 @@
...
+from django.contrib.messages.storage.base import BaseStorage, Message
+
+class CookieStorage(BaseStorage):
+    def _update_cookie(self, encoded_data, response):
+        """
+        Either sets the cookie with the encoded data if there is any data to
+        store, or deletes the cookie.
+        """
+        if encoded_data:
+            # ok
```
minusworld: Should this one fire because samesite is not set?

python/django/security/secure-cookies.py
```diff
@@ -0,0 +1,57 @@
...
+    response.set_cookie(
+        settings.SESSION_COOKIE_NAME,
+        request.session.session_key, max_age=max_age,
+        expires=expires, domain=settings.SESSION_COOKIE_DOMAIN,
+        path=settings.SESSION_COOKIE_PATH,
+        secure=settings.SESSION_COOKIE_SECURE or None,
+        httponly=settings.SESSION_COOKIE_HTTPONLY or None,
+    )
+
+    # ok
```
minusworld: Should this one fire because samesite is not set?

python/django/security/secure-cookies.py
minusworld: This is excluded b/c of vendored code. See comment in yaml file

python/django/security/secure-cookies.py
minusworld: See comment above

python/django/security/secure-cookies.py
minusworld: See comment above

nbrahms:
> There are a few test cases that appear as though they should fire because `samesite` is not set. Could you look at them and make sure they are annotated correctly? 🙏

These all appear to be test cases within django itself. Probably the Django project should be excluded from the input set. I don't consider them positives or negatives either way.

PR #218, opened 2020-03-04
Fix #159, add rule for deprecated Flask APIs

### LINKED ISSUE #159: Flask Deprecated APIs
Needs list

mschwager:
This information was mostly gleaned by searching [Flask's CHANGELOG](https://github.com/pallets/flask/blob/master/CHANGES.rst) for "deprecated." There were a few more, but it's probably not necessary to search for things deprecated/removed in 2011. Once https://github.com/returntocorp/sgrep/issues/197 is in we'll be able to search for `flask.ext` use as well.

Fixes https://github.com/returntocorp/sgrep-rules/issues/159

### DIFF
--- a/python/flask/deprecated-apis.yaml
+++ b/python/flask/deprecated-apis.yaml
@@ -1,0 +1,41 @@
+rules:
+  - id: flask-deprecated-apis
+    patterns:
+      - pattern-either:
+        - pattern: |
+              $F = Flask(...)
+              ...
+              $F.open_session(...)
+        - pattern: |
+              $F = Flask(...)
+              ...
+              $F.save_session(...)
+        - pattern: |
+              $F = Flask(...)
+              ...
+              $F.make_null_session(...)
+        - pattern: |
+              $F = Flask(...)
+              ...
+              $F.init_jinja_globals(...)
+        - pattern: |
+              $F = Flask(...)
+              ...
+              $F.request_globals_class(...)
+        - pattern: |
+              $F = Flask(...)
+              ...
+              $F.static_path(...)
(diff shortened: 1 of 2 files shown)


mschwager:
This rule works pretty well: https://dev.massive.ret2.co/triager/score/633?check_id=tmp.flask-deprecated-apis

Using these deprecated APIs seems uncommon, but the rule correctly detects them when they are used.

python/flask/deprecated-apis.yaml
```diff
@@ -0,0 +1,41 @@
...
+        - pattern: app.save_session(...)
+        - pattern: app.make_null_session(...)
+        - pattern: app.init_jinja_globals(...)
+        - pattern: app.request_globals_class(...)
+        - pattern: app.static_path(...)
+        - pattern: app.config.from_json(...)
+        - pattern: flask.json_available
+        - pattern: flask.request.module
+        - pattern: flask.testing.make_test_environ_builder(...)
+    message: "deprecated Flask API"
```
nbrahms: Just a note that, if we ship this check, we'll probably need to update this message.

mschwager: Yup, I agree. What should we add to make it better?

nbrahms: Probably a question for @minusworld ?

mschwager: That all sounds good to me! I'll push this through now, and if we decide this is a golden check in the future we can revisit the message :+1:

minusworld:
https://dev.massive.ret2.co/jobs/664?page=1&sort_on=repo_url&sort_by=asc

PR #223, opened 2020-03-05
WIP: Add rules to detect when mutable default list/dict are shared

brendongo:


brendongo:
Waiting on a way to select the offending default assignment instead of firing on the entire class. This is very noisy and would benefit from more fine grained where/contains operators.

ulziibay:
@brendongo what do you want to do with this check?

ievans:
@brendongo do you want to update this or should we close out? closing and letting you re-open if you like,

PR #225, opened 2020-03-05
move test.py

ievans:
cc @DrewDennison https://github.com/returntocorp/sgrep/pull/256/


DrewDennison:
Appears there is an actual test failure in `deadcode/baseclass-attribute-override.py`

PR #247, opened 2020-03-17
Airflow/bash operator

minusworld:
This is a new check for the Apache Airflow BashOperator class.

As the name implies, BashOperator will run a bash command on an airflow instance. The same principles apply here as with any command string: formatted or concatenated strings are dangerous.

This check will look for these cases, alerting when formatted or concatenated strings are passed into BashOperator.

https://airflow.apache.org/docs/stable/_api/airflow/operators/bash_operator/index.html

This was inspired by a review of Etsy's `boundary-layer` tool, which transforms YAML files into Apache Airflow Python code. This code is autogenerated, and in looking for invariants to express about Airflow, this one for BashOperator seemed obvious.

I tested doing command injection via this method and was successful.

### DIFF
--- a/python/airflow/formatted-string-bashoperator.py
+++ b/python/airflow/formatted-string-bashoperator.py
@@ -1,0 +1,98 @@
+import requests
+from datetime import timedelta
+from airflow import DAG
+from airflow.operators.bash_operator import BashOperator
+from airflow.utils.dates import days_ago
+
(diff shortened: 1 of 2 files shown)


minusworld:
https://dev.massive.ret2.co/jobs/731?page=1&sort_on=repo_url&sort_by=asc

PR #252, opened 2020-03-17
Improve java.eqeq-is-bad message

nbrahms:
x != x is often used to test if x is a numeric value. Since Java
1.early, one can just use Double.isNaN, so recommend this.

### DIFF
--- a/java/basic-rules.yaml
+++ b/java/basic-rules.yaml
@@ -16,6 +16,6 @@ rules:
           - pattern: $X == $X
           - pattern: $X != $X
       - pattern-not: 1 == 1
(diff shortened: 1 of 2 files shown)


java/basic-rules.yaml
```diff
@@ -16,6 +16,6 @@ rules:
           - pattern: $X == $X
           - pattern: $X != $X
       - pattern-not: 1 == 1
-    message: "useless comparison operation `$X == $X` or `$X != $X`"
+    message: "`$X == $X` or `$X != $X` is often a useless comparison. To test if `$X` is not-a-number, use `Double.isNaN($X)`."
```
ievans:
```suggestion
    message: "`$X == $X` or `$X != $X` is a useless comparison unless the value compared is a float or double. To test if `$X` is not-a-number, use `Double.isNaN($X)`."
```

PR #254, opened 2020-03-17
Fix two false positives in flask-view-func-match-route-params

nbrahms:
1. Using single instead of double quotes in string literals (IMO sgrep
   should deal with this for you)
2. Defining converters in the route (see
   https://flask.palletsprojects.com/en/1.1.x/quickstart/#variable-rules)


DrewDennison:
@minusworld Nathan fixed your FP findings from the original one

PR #255, opened 2020-03-18
Added golang check for ListenAndServe without TLS.

minusworld:


minusworld:
https://dev.massive.ret2.co/jobs/730?page=1&sort_on=repo_url&sort_by=asc

PR #256, opened 2020-03-18
Added a new check for formatted/concatenated strings into template.HTML

minusworld:
Added a new check for formatted/concatenated strings into template.HTML which does not escape input.

### DIFF
--- a/go/net/formatted-template-string.go
+++ b/go/net/formatted-template-string.go
@@ -4,6 +4,7 @@ import (
     "fmt"
     "html/template"
     "net/http"
+    "strconv"
 )
 
 func Fine(r *http.Request) template.HTML {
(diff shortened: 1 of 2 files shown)


minusworld:
https://dev.massive.ret2.co/triager/filter/716

go/net/formatted-template-string.go
```diff
@@ -0,0 +1,52 @@
...
+        return template.HTML("")
+    }
+    return template.HTML(tmpl)
+}
+
+func FormattedInline(r *http.Request) template.HTML {
+    customerId := r.URL.Query().Get("id")
+    // ruleid: formatted-template-string
+    return template.HTML(fmt.Sprintf("<html><body><h1>%s</h1></body></html>", customerId))
+}
```
nbrahms: Might be worth adding a (presumably currently broken) test case with control flow

go/net/formatted-template-string.go
```diff
@@ -0,0 +1,52 @@
...
+
+func main() {}
```
nbrahms: This seems like a useful negative test case:

```go
return template.HTML("foo" + "bar")
```

go/net/formatted-template-string.go
```diff
@@ -0,0 +1,52 @@
...
+    if (err != nil) {
+        return template.HTML("")
+    }
+    return template.HTML(tmpl)
+}
+
+func FormattedInline(r *http.Request) template.HTML {
+    customerId := r.URL.Query().Get("id")
+    // ruleid: formatted-template-string
+    return template.HTML(fmt.Sprintf("<html><body><h1>%s</h1></body></html>", customerId))
```
nbrahms: Is there no good way to detect that `customerId` was assigned out of the `Request` object?

minusworld: In other checks, I used `pattern-inside` to limit checking when `$R *http.Request` was a function argument. I wanted to see what else was out there first: https://dev.massive.ret2.co/triager/filter/716

go/net/formatted-template-string.go
minusworld: Good idea

go/net/formatted-template-string.go
minusworld: Also a good idea

PR #261, opened 2020-03-24
Check for unsanitized reflection into flask.make_response

### LINKED ISSUE #210: User input format string into flask.make_response
This issue provides a great example

https://github.com/kylewm/silo.pub/issues/57

https://github.com/kylewm/silo.pub/blob/46aece85f8918f56ed75f1e11b544c10f70a17fc/silopub/micropub.py#L107

nbrahms:
Looks for calls to make_response using unsanitized request parameters.

Closes #210 .

### DIFF
--- a/python/flask/unsanitized_input.yaml
+++ b/python/flask/unsanitized_input.yaml
@@ -2,38 +2,38 @@ rules:
   - id: response_contains_unsanitized_input
     patterns:
       - pattern-either:
-        - pattern: |
-            $X = request.args.get(...)
-            ...
-            return flask.make_response("...".format($X))
-        - pattern: |
-            $X = request.args.get(...)
-            ...
-            $R = flask.make_response("...".format($X))
-        - pattern: |
-            $X = request.args.get(...)
-            ...
-            return flask.make_response(f"...{$X}...")
-        - pattern: |
-            $X = request.args.get(...)
-            ...
-            $R = flask.make_response(f"...{$X}...")
-        - pattern: |
-            $X = request.args.get(...)
-            ...
-            return flask.make_response(f"...{$X}")
-        - pattern: |
-            $X = request.args.get(...)
-            ...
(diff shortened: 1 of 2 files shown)


nbrahms:
Blocked on https://github.com/returntocorp/sgrep/issues/326

nbrahms:
> Blocked on [returntocorp/sgrep#326](https://github.com/returntocorp/sgrep/issues/326)

Thanks @ulziibay for pointing out I just needed to update my pattern.

python/flask/unsanitized_input.yaml
```diff
@@ -0,0 +1,39 @@
+rules:
+  - id: response_contains_unsanitized_input
+    patterns:
+      - pattern-either:
```
ulziibay: I would probably add `pattern-inside` so that you are only catching on function scope

nbrahms: Unfortunately this doesn't work, because metavariables aren't the same between different patterns :(

python/flask/unsanitized_input.yaml
```diff
@@ -0,0 +1,39 @@
...
+              $R = flask.make_response(f"...{$X}")
+          - pattern: |
+              $X = request.args.get(...)
+              ...
+              return flask.make_response(f"{$X}...")
+          - pattern: |
+              $X = request.args.get(...)
+              ...
+              $R = flask.make_response(f"{$X}...")
+    message: Flask response reflects unsanitized user input.
```
ulziibay: Need to add more verbose messaging. How about we give more details, similar to https://github.com/returntocorp/sgrep-rules/issues/210
```
 Flask response reflects unsanitized user input. This introduces possible [XSS attack](https://www.owasp.org/index.php/Cross-site_Scripting_(XSS) ) as user can supply input like "<script>some_code</script>" and execute arbitrary code in user's browser.
```

PR #271, opened 2020-03-26
Django/unvalidated passwords

minusworld:
Two new checks:
1. https://github.com/returntocorp/sgrep-rules/issues/270, which looks for missing calls to `validate_password`
2. https://github.com/returntocorp/sgrep-rules/issues/267, which looks for default arguments set as the empty string flowing into `set_password`.


python/django/security/passwords/use-none-for-password-default.yml
```diff
@@ -0,0 +1,20 @@
...
+              ...
+              $MODEL.set_password($VAR)
+              ...
+              $MODEL.save(...)
+          - pattern: |
+              def $F(..., $VAR="", ...):
+                ...
+                $MODEL.set_password($VAR)
+    message: |
+      '$VAR' is using the empty string as its default and is being used to set
```
ulziibay: Does this message work with 2 metavars?

minusworld: Yup!

PR #272, opened 2020-03-26
Added checks for user data going into eval and exec.

minusworld:
Addresses https://github.com/returntocorp/sgrep-rules/issues/260

### DIFF
--- a/python/django/security/user-eval-format-string.yaml
+++ b/python/django/security/user-eval-format-string.yaml
@@ -1,0 +1,133 @@
+rules:
+  - id: user-eval-format-string
+    patterns:
+      - pattern-inside: |
+          def $F(...):
+            ...
+      - pattern-either:
+          - pattern: eval(..., $STR % request.$W.get(...), ...)
+          - pattern: |
+              $V = request.$W.get(...)
+              ...
+              eval(..., $STR % $V, ...)
+          - pattern: |
+              $V = request.$W.get(...)
+              ...
+              $S = $STR % $V
+              ...
+              eval(..., $S, ...)
+          - pattern: |
+              $V = request.$W.get(...)
+              ...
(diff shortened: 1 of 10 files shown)


python/django/security/user-eval-format-string.yaml
```diff
@@ -0,0 +1,133 @@
+rules:
+  - id: user-eval-format-string
+    patterns:
+      - pattern-inside: |
+          def $F(...):
+            ...
+      - pattern-either:
```
ulziibay: you can add the f-string variants of these checks in one go

python/django/security/user-exec.yaml
```diff
@@ -0,0 +1,50 @@
...
+          - pattern: |
+              $V = request.$W[...]
+              ...
+              $A = exec(..., $V, ...)
+          - pattern: |
+              $V = request.$W[...]
+              ...
+              return exec(..., $V, ...)
+    message: |
+      Found user data in a call to 'exec'. This is extremely dangerous.
```
ulziibay: Add maybe a link about RCE

python/django/security/user-eval-format-string.yaml
```diff
@@ -0,0 +1,133 @@
...
+              ...
+              $A = eval(..., $S, ...)
+          - pattern: |
+              $V = request.$W[...]
+              ...
+              $S = $STR.format(..., $V, ...)
+              ...
+              return eval(..., $S, ...)
+    message: |
+      Found user data in a call to 'eval'. This is extremely dangerous.
```
ulziibay: Maybe add a link about RCE

python/django/security/user-exec.yaml
minusworld: Roger. That's a good idea.

PR #273, opened 2020-03-26
Added check for user data going into open()

minusworld:
Addresses https://github.com/returntocorp/sgrep-rules/issues/268

### DIFF
--- a/python/django/security/user-open.yaml
+++ b/python/django/security/user-open.yaml
@@ -60,7 +60,7 @@ rules:
               with open($V, ...) as $FD:
                 ...
     message: |
-      Found user data in a call to 'open'. Make absolutely sure the user data
+      Found request data in a call to 'open'. Ensure the request data
       is validated or sanitized, otherwise it could result in path traversal attacks.
     languages: [python]
     severity: WARNING
--- a/python/django/security/user-open.yaml
+++ b/python/django/security/user-open.yaml
@@ -1,0 +1,66 @@
+rules:
+  - id: user-open
+    patterns:
+      - pattern-inside: |
+          def $F(...):
+            ...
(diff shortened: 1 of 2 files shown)


python/django/security/user-open.yaml
```diff
@@ -0,0 +1,66 @@
...
+              open(..., $V, ...)
+          - pattern: |
+              $V = request.$W.get(...)
+              ...
+              $A = open(..., $V, ...)
+          - pattern: |
+              $V = request.$W.get(...)
+              ...
+              return open(..., $V, ...)
+          - pattern: open(..., request.$W(...), ...)
```
mschwager: What is this pattern getting at? Looks like it could be `request.GET(...)`? Maybe this is too general, if there's specific function calls we're concerned about we should list them.

python/django/security/user-open.yaml
```diff
@@ -0,0 +1,66 @@
...
+              $V = request.$W(...)
+              ...
+              with open($V, ...) as $FD:
+                ...
+          - pattern: |
+              $V = request.$W[...]
+              ...
+              with open($V, ...) as $FD:
+                ...
+    message: |
```
mschwager: Slight nit here: I would change `Found user data ...` -> `Found request data ...` and `Make absolutely sure ...` -> `Please ensure ...`.

python/django/security/user-open.yaml
minusworld: There are a couple of methods on the request objects that are conveniences for getting certain attributes, as an example: `request.get_host()`

cf. https://docs.djangoproject.com/en/3.0/ref/request-response/#django.http.HttpRequest.get_host

PR #281, opened 2020-03-26
SQLi check for extra()

ulziibay:
Adds SQLi checks for extra() using const propagation.
Addresses last piece of https://github.com/returntocorp/sgrep-rules/issues/263

### DIFF
--- a/python/django/security/sql_injection_extra.yaml
+++ b/python/django/security/sql_injection_extra.yaml
@@ -40,9 +40,9 @@ rules:
             ...
             $MODEL.objects.extra(..., where=[$S % $V], ...)
     message: |
-      Data from request is passed to RawSQL(). 
+      Data from request is passed to RawSQL().
       This is a SQL injection and could be exploited.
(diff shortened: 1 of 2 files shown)


ulziibay:
job: https://dev.massive.ret2.co/jobs/759?page=1&sort_on=repo_url&sort_by=asc

python/django/security/sql_injection_extra.yaml
```diff
@@ -0,0 +1,48 @@
...
+    patterns:
+      - pattern-inside: |
+          def $F(...):
+            ...
+      - pattern-either:
+        # match when `format` is used
+        - pattern: |
+            $V = request.$W.get($X)
+            ...
+            $MODEL.objects.extra(..., where=[$S.format($V)], ...)
```
minusworld: I think doing `$S.format(..., $V, ...)` will catch more instances

ulziibay: addressed. lmk what you think. Also, there where clause expects a list, but i cannot do 
```
where=[..., $S.format($V), ...]
```
in the list at the moment in sgrep

PR #285, opened 2020-03-30
Add SQLi detection for execute() sink

ulziibay:
Addresses https://github.com/returntocorp/sgrep-rules/issues/283

### DIFF
--- a/python/django/security/sql-injection-using-db-cursor-execute.yaml
+++ b/python/django/security/sql-injection-using-db-cursor-execute.yaml
@@ -5,7 +5,7 @@ rules:
           def $F(...):
             ...
       - pattern-either:
-        # match when direct flow
+        # match when direct flow format func
         - pattern: |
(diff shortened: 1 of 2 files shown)


python/django/security/sql-injection-using-db-cursor-execute.yaml
```diff
@@ -0,0 +1,49 @@
+rules:
+  - id: sql-injection-db-cursor-execute
+    patterns:
+      - pattern-inside: |
+          def $F(...):
+            ...
+      - pattern-either:
+        # match when `format` is used
+        - pattern: |
+            $V = request.$W.get($X)
```
minusworld: Can you add in the single-line case?
```yaml
$CURSOR.execute(request.$W.get(...), ...)
```

ulziibay: Will do

PR #290, opened 2020-03-31
Add python/django/security/ssrf-injection-requests

nbrahms:
Detects SSRF injection attacks into the requests API.

Neither of these checks actually fire on anything in platform, so
they're pretty speculative.

For completeness, I had to write 48 different patterns for each sink. :(

This cardinality is the product space of
- Django import identification (import HttpResponse and import render)
- Request-object data extraction (request[], request.get,
  request.body[], request.body.get)
- Possible ways of embedding the injection expression (bare, assignment, return) -- note that this is incomplete, as there is also call site
- Injection methods (method call, e.g. by string format, and f-strint)

In order to streamline this generation in the future, I add jinja
templating.

For instance, to generate ssrf-injection-requests.yaml, I ran:

```bash
./generate-injection.py \
  --id ssrf-injection-requests \
  --sink 'requests.$METHOD' \
  --message 'Data from request object is passed to a new server-side request. This could lead to a server-side request forgery (SSRF). To mitigate, ensure that schemes and hosts are validated against a whitelist, do not forward the response to the user, and ensure proper authentication and transport-layer security in the proxied request. See https://owasp.org/www-community/attacks/Server_Side_Request_Forgery to learn more about SSRF vulnerabilities.' \ 
  > ssrf-injection-requests.yaml
```

### DIFF
--- a/Pipfile
+++ b/Pipfile
@@ -1,0 +1,11 @@
+[[source]]
+name = "pypi"
+url = "https://pypi.org/simple"
+verify_ssl = true
+
+[dev-packages]
+
+[packages]
+
+[requires]
+python_version = "3.7"
--- a/Pipfile
+++ b/Pipfile
@@ -4,6 +4,7 @@ url = "https://pypi.org/simple"
 verify_ssl = true
 
 [dev-packages]
+jinja2 ~= "2.11.1"
 
 [packages]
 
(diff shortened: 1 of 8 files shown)


minusworld:
https://github.com/returntocorp/sgrep-rules/issues/278

PR #291, opened 2020-04-01
Path traversal detection

ulziibay:
Path traversal detection when data flows from `request` to  `os.path.join` to `open()` to `response`
Addresses https://github.com/returntocorp/sgrep-rules/issues/269


python/django/security/path-traversal-join.yaml
```diff
@@ -0,0 +1,50 @@
...
+          ...
+      - pattern-either:
+        - pattern: |
+            $V = request.$W.get($X)
+            ...
+            $P = os.path.join(..., $V)
+            ...
+            open($P, ...)
+            ...
+            return django.http.HttpResponse(...)
```
minusworld, line 23: What happens if we remove this? This seems like it might limit the findings unnecessarily.

python/django/security/path-traversal-join.yaml
```diff
@@ -0,0 +1,50 @@
...
+        - pattern: |
+            $V = request.$W
+            ...
+            $P = os.path.join(..., $V)
+            ...
+            open($P, ...)
+            ...
+            return django.http.HttpResponse(...)
+    message: ""
```
minusworld, line 48: Could you add in a message?

python/django/security/path-traversal-join.yaml
ulziibay: Sure thing. Broader is fine with me!

python/django/security/path-traversal-join.yaml
```diff
@@ -0,0 +1,45 @@
...
+          ...
+      - pattern-not: |
+          ...
+          os.path.realpath(...)
+          ...
+      - pattern-either:
+        - pattern: |
+            $V = request.$W.get($X)
+            ...
+            $P = os.path.join(..., $V)
```
minusworld, line 19: Sorry! Last nit... would we catch more by doing `(..., $V, ...)` in case they put it in the middle of `os.path.join`?

PR #293, opened 2020-04-02
New check for request data going into .write

minusworld:
This wasn't a ticket. Just an experiment. After review of findings I don't think it's actually a good "injection" check.
Findings: https://dev.massive.ret2.co/triager/filter/812

### DIFF
--- a/python/django/security/request-data-write.py
+++ b/python/django/security/request-data-write.py
@@ -19,7 +19,7 @@ def save_scrawl_file(request, filename):
 def save_file(request):
     # ok
     user = User.objects.get(username=request.session.get('user'))
-    content = request.POST.get("file-contents", "")
+    content = "user logged in at {}".format(time.time())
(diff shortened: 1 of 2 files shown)


python/django/security/request-data-write.yaml
```diff
@@ -0,0 +1,580 @@
+rules:
```
ulziibay, line 1: there is a ton of extra lines in the rule. PLease clean up

python/django/security/request-data-write.py
```diff
@@ -0,0 +1,15 @@
+from django.http import HttpResponse
+from . import settings as USettings
+
+def save_scrawl_file(request, filename):
+    import base64
+    try:
+        # ruleid: request-data-write
```
ulziibay, line 7: add True negative(s)

minusworld:
I've addressed your comment -- I removed all the weird spacing and added a true negative test case.

PR #295, opened 2020-04-02
Django/email

minusworld:
Addresses https://github.com/returntocorp/sgrep-rules/issues/266


python/django/security/templates/injection-template.yaml.j2
```diff
@@ -1,217 +1,27 @@
...
-              ...
-              return {{ sink }}(f"...{$DATA}...", ...)
-    message: |
-      {{ message | indent(width=8)  }}
-    languages: [python]
-    severity: ERROR
+            {{ left_pattern }}{{ sink }}(..., $INTERM, ...)
+    {%- endfor -%}
+  {%- endfor -%}
+{%- endfor -%}
```
ulziibay, line 27: can you make sure there is newline at the end?

minusworld: Roger o7

PR #296, opened 2020-04-02
Django/reflected httpresponse

minusworld:
Addresses https://github.com/returntocorp/sgrep-rules/issues/215


python/django/security/reflected-data-httpresponse.yaml
```diff
@@ -0,0 +1,622 @@
+rules:
+  - id: reflected-data-httpresponse
+    message: Found request data reflected into HttpResponse. This could be vulnerable to XSS. Ensure the request data is properly escaped or sanitzed.
```
ulziibay, line 3: good idea to have the messsage up top!

PR #297, opened 2020-04-02
Injection to file name

ulziibay:
Add checks for detecting when request data flows into formatted file names.
Addresses https://github.com/returntocorp/sgrep-rules/issues/282

### DIFF
--- a/python/django/security/path-traversal-file-name.yaml
+++ b/python/django/security/path-traversal-file-name.yaml
@@ -48,9 +48,9 @@ rules:
             ...
             $FILE.format(..., $V, ...)
         - pattern-where-python: |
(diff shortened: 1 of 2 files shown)


python/django/security/path-traversal-file-name.yaml
```diff
@@ -0,0 +1,57 @@
...
+        - pattern: |
+            $V = request.$W($X)
+            ...
+            $FILE.format(..., $V, ...)
+        - pattern: |
+            $V = request.$W
+            ...
+            $FILE.format(..., $V, ...)
+        - pattern-where-python: |
+            vars.get('$FILE', '').split('.')[-1] in ['log', 'zip','txt', 'csv']
```
minusworld, line 51: Nice, that's really clever 👍

DrewDennison: Nice! Could use a set for negligible pref gains

PR #298, opened 2020-04-03
New check for mass assignment.

minusworld:
Addresses: https://github.com/returntocorp/sgrep-rules/issues/289

### DIFF
--- a/python/django/security/mass-assignment.py
+++ b/python/django/security/mass-assignment.py
@@ -1,6 +1,8 @@
 from django.shortcuts import render
 from myapp.models import Whatzit
 
+# Test cases borrowed from https://gist.github.com/jsocol/3217262
+
 def create_whatzit(request):
     # ruleid: mass-assignment
     Whatzit.objects.create(**request.POST)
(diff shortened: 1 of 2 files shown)


python/django/security/mass-assignment.py
```diff
@@ -0,0 +1,22 @@
+from django.shortcuts import render
+from myapp.models import Whatzit
+
+def create_whatzit(request):
+    # ruleid: mass-assignment
+    Whatzit.objects.create(**request.POST)
+    return render(request, 'created.html')
+
+def update_whatzit(request, id):
```
ulziibay, line 9: haha liking the func name :)

python/django/security/mass-assignment.yaml
```diff
@@ -0,0 +1,16 @@
+rules:
+- id: mass-assignment
+  languages: [python]
+  severity: WARNING
+  message: |
+    Mass assignment detected. This can result in assignment to
```
ulziibay, line 6: A link would be nice to learn about why it matters

PR #306, opened 2020-04-08
Add math.rand (G404) checks

ulziibay:
This alerts on usage of Int() and Read() functions of math.rand module.
It acts the same as [G404](https://github.com/securego/gosec/blob/master/rules/rand.go)
~BLOCKED on Sgrep engine improvement in https://github.com/returntocorp/sgrep/issues/411~


mschwager (approved):
LGTM

PR #307, opened 2020-04-09
Hardcoded credential check without entrophy (G101)

ulziibay:
Tries to mimic [G101](https://github.com/securego/gosec/blob/master/rules/hardcoded_credentials.go) without
the entropy check on the value of the string literals. Blocked on https://github.com/returntocorp/sgrep/issues/412

### DIFF
--- a/go/gosec/hardcoded_credentials/hardcoded_credentials.yaml
+++ b/go/gosec/hardcoded_credentials/hardcoded_credentials.yaml
@@ -1,0 +1,16 @@
+rules:
+  - id: hardcoded-credentials
+    message: Potential hardcoded credentials
+    severity: WARNING
+    languages: [go]
+    patterns:
+      # len([cred_name == vars['$VALUE'] for cred_name in ['"secret"', 'passwd','pass','password','pwd', 'secret', 'token']]) > 1
+      - pattern-either: 
+          - pattern: $VAR = "..."
+          - pattern: $VAR := "..."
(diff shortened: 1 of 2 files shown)


go/gosec/hardcoded_credentials/hardcoded_credentials.yaml
```diff
@@ -0,0 +1,16 @@
...
+    severity: WARNING
+    languages: [go]
+    patterns:
+      # len([cred_name == vars['$VALUE'] for cred_name in ['"secret"', 'passwd','pass','password','pwd', 'secret', 'token']]) > 1
+      - pattern-either:
+          - pattern: $VAR = "..."
+          - pattern: $VAR := "..."
+          - pattern: var $VAR = "..."
+          - pattern: const $VAR = "..."
+      - pattern-where-python: |
```
minusworld, line 13: This isn't a big deal for right now, but anecdotally we know that `token` and `secret` are prone to high FP. This was the case with the Bandit check which did the same thing: https://bento.dev/blog/2019/three-things-your-linter-shouldnt-tell-you/

ulziibay: Yeah ideally, i want to add the entropy check to the literal so as to reduce the FP rate.

ulziibay:
In absence of where-python support for libraries, I would not want to push this check to registry. I will close this out!

PR #324, opened 2020-04-10
Fix/update injection checks

minusworld:
Rearranges all the checks under `python/django/security` into categories. There are now three new folders:
- audit
- injection
- passwords

Audit is the only one which requires some explanation. These are checks that are not explicitly aimed at finding an issue -- these are stricter rules that help one identify areas to audit. An example is any call using `.extra()`.

Additionally, this PR updates many of the injection checks to use as many patterns as we now know about. Most of them need to be updated with "inline formatting" patterns and with "intermediate formatting" patterns.


DrewDennison:
@minusworld I merged in develop so most of the test pass. One failing test in your modified file. I believe you can fix it by just adding a `ruleid:` to https://github.com/returntocorp/sgrep-rules/pull/324/files#diff-4531088e10b49e090b3d6fa5b6770c16R27 but didn't want to do that without the domain knowledge

PR #330, opened 2020-04-14
* javascript/jwt-hardcode/jwt-hardcode.js: Fix todoruleid

aryx:
Really deep stmt matching probably allows to report this new error

test plan:
~/github/sgrep/sgrep_lint/sgrep.py --dangerously-allow-arbitrary-code-execution-from-rules --strict --test --test-ignore-todo .


inkz:
yep, it works!

PR #370, opened 2020-04-29
SSLv3 blacklist rule

ulziibay:
Lets discourage usage of sslv3. G402 is very strict as
it whitelists Mozilla's https://wiki.mozilla.org/Security/Server_Side_TLS
and anything outside of it as bad!
 
Per discussion, we will only blacklist SSLv3 and other broken TSL ciphers.

```
ulziibayarotgonbaatar@ulziibayars-MacBook-Pro sgrep-rules % semgrep --dangerously-allow-arbitrary-code-execution-from-rules -f go/gosec/tls_ssl_blacklist/ssl.yaml go/gosec/tls_ssl_blacklist
/Users/ulziibayarotgonbaatar/Workspace/sgrep-rules/go/gosec/tls_ssl_blacklist/ssl.go
rule:go.gosec.tls_ssl_blacklist.do-not-use-ssl-v3: SSLv3 is known to be broken. Use it with caution. https://www.us-cert.gov/ncas/alerts/TA14-290A for more information and mitigation.
36:	client := &http.Client{
37:		Transport: &http.Transport{
38:			TLSClientConfig: &tls.Config{
39:				KeyLogWriter: w,
40:				// ruleid: do-not-use-ssl-v3
41:				MinVersion:         tls.VersionSSL30,
42:				Rand:               zeroSource{}, // for reproducible output; don't do this.
43:				InsecureSkipVerify: true,         // test server certificate is not trusted.
```


go/gosec/tls_ssl_blacklist/ssl.yaml
```diff
@@ -0,0 +1,13 @@
...
+      SSLv3 is known to be broken. Use it with caution.
+      https://www.us-cert.gov/ncas/alerts/TA14-290A
+      for more information and mitigation.
+    languages: [go]
+    severity: WARNING
+    patterns:
+      - pattern: |
+          tls.Config{..., MinVersion: $MINV, ...}
+      - pattern-where-python: |
+          'VersionSSL30' in vars['$MINV']
```
mschwager, line 13: Can we just use the pattern `MinVersion: 'VersionSSL30'` instead of `where-python`?

go/gosec/tls_ssl_blacklist/ssl.yaml
```diff
@@ -0,0 +1,13 @@
+rules:
+  - id: do-not-use-ssl-v3
```
mschwager, line 2: I tend to prefer saying something like `ssl-v3-is-insecure`. Telling users to blindly not use something doesn't really give them any idea *why* they shouldn't use it.

ulziibay: ok addressed!

go/gosec/tls_ssl_blacklist/ssl.yaml
ulziibay: ok simplified!

PR #381, opened 2020-05-01
Port over G203

ulziibay:
Port Port G301, G302, G306, which are the filesystem  permissons checks. Uses 0o600 as threshold.


ulziibay:
@minusworld gosec uses 0700 for some of this and 0600 for others. I did not run this on platform, but would be trivial to do!

minusworld:
https://dev.massive.ret2.co/triager/filter/963

ulziibay:
@minusworld it is not too restrictive nor noisy based on 9 out of ~1300 repos have this check fired. Should we ship it as is?

minusworld:
> @minusworld it is not too restrictive nor noisy based on 9 out of ~1300 repos have this check fired. Should we ship it as is?

sg! 👍

PR #407, opened 2020-05-05
Port over G305

ulziibay:
The issue is discussed in https://github.com/securego/gosec/issues/205 and 
it seems to detect a real path traversal.

### DIFF
--- a/go/gosec/traversal/zip.go
+++ b/go/gosec/traversal/zip.go
@@ -1,0 +1,45 @@
+package unzip
+
+import (
+	"archive/zip"
+	"io"
+	"os"
+	"path/filepath"
+)
+
+func unzip(archive, target string) error {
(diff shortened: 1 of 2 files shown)


go/gosec/traversal/zip.go
```diff
@@ -0,0 +1,46 @@
...
+
+import (
+	"archive/zip"
+	"io"
+	"os"
+	"path/filepath"
+)
+
+func unzip(archive, target string) error {
+	// ruleid: path-traversal-inside-zip-extraction
```
minusworld, line 11: Can we add a True Negative test as well?

minusworld:
https://dev.massive.ret2.co/triager/filter/964

go/gosec/traversal/zip.go
ulziibay: Will do!

PR #409, opened 2020-05-05
Port over G303

ulziibay:
Bad temp file creatiojn is CWE: 377


go/gosec/bad_temp_file/bad_tmp.go
```diff
@@ -0,0 +1,14 @@
+package samples
+
+import (
+	"fmt"
+	"io/ioutil"
+)
+
+func main() {
```
minusworld, line 8: Can we add a True Negative test case as well?

ulziibay: Will do!

PR #411, opened 2020-05-05
Port over G102

ulziibay:
Detection for bind all for '0.0.0.0'. This is similar to python/socket


go/gosec/socket/bind_all.go
```diff
@@ -0,0 +1,15 @@
+package main
+
+import (
+	"log"
+	"net"
+)
+
+func main() {
+	// ruleid: avoid-bind-to-all-interfaces
```
minusworld, line 9: Can we add a True Negative test case as well?

PR #415, opened 2020-05-05
Port over G110

ulziibay:
Potential DOS via decomp. bomb. Used bomb.codes as a reference point.


minusworld:
https://dev.massive.ret2.co/triager/filter/962

ulziibay:
It is not a noisy check based on the platfom job!

PR #416, opened 2020-05-05
Port over G103

ulziibay:
Unsafe pkg is potentially dangerous. Lets warn about the usage.

### DIFF
--- a/go/gosec/bad_imports/unsafe.yaml
+++ b/go/gosec/bad_imports/unsafe.yaml
@@ -1,0 +1,14 @@
+rules:
+  - id: avoid-using-unsafe-pkg
+    message:
+      Package `unsafe` contains operations that step around the type safety of Go programs.
+      Packages that import unsafe may be non-portable and are not protected by the Go 1 compatibility guidelines.
+    metadata:
+      cwe: 'CWE-242: Use of Inherently Dangerous Function'
+      source_rule_url: https://github.com/securego/gosec
+    languages: [go]
+    severity: WARNING
+    pattern: |
+      import "unsafe"
+      ...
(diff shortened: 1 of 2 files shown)


go/gosec/bad_imports/unsafe.yaml
```diff
@@ -0,0 +1,14 @@
+rules:
+  - id: avoid-using-unsafe-pkg
+    message:
+      Package `unsafe` contains operations that step around the type safety of Go programs.
+      Packages that import unsafe may be non-portable and are not protected by the Go 1 compatibility guidelines.
+    metadata:
+      cwe: 'CWE-242: Use of Inherently Dangerous Function'
+      source_rule_url: https://github.com/securego/gosec
```
minusworld, line 8: Can we standardize on whether we use hyphens '-' or underscores '_'?

I think we use hyphens everywhere else. Is there a parsing reason we use underscores here?

minusworld:
https://dev.massive.ret2.co/triager/filter/961

go/gosec/bad_imports/unsafe.yaml
ulziibay: Thanks for the flag. I wish we can enforce it during the rule writing process. I will address all the renaming of `source_rule_url ` in a separate PR. Sounds good?

ulziibay:
@minusworld it is the same! I just happened to have done it and used it to pair program with a client!

go/gosec/bad_imports/unsafe.yaml
minusworld: Sounds good!

ulziibay:
Closing this in favor of  #418

PR #420, opened 2020-05-06
Updated the open-redirect check 

minusworld:
Updated the open-redirect check to filter out cases with 'if is_safe_url' with the new <... ...> syntax.


minusworld:
Superceded by https://github.com/returntocorp/semgrep-rules/pull/542

PR #467, opened 2020-05-09
Add check for java: blowfish-insufficient-key-size.

minusworld:


minusworld:
Addresses https://github.com/returntocorp/semgrep-rules/issues/448

Issue #457, opened 2020-05-08
[Check] Java: CORS wildcard


minusworld:
**Check Description**

cf. https://find-sec-bugs.github.io/bugs.htm#PERMISSIVE_CORS

- [x] This ticket has links, references, or examples.
- [x] Your check has true positive and true negative test cases.
- [x] Your check has been labeled with true positive and false positive findings on at least 10 repositories, and the link is pasted in the ticket. If there are no findings, paste the triager link anyway.
- [x] Your check PR has been reviewed and merged.
Now, close the ticket!

inkz:
code examples:
https://www.baeldung.com/spring-response-header
https://www.programcreek.com/java-api-examples/?class=javax.servlet.http.HttpServletResponse&method=addHeader

inkz:
blocked by https://github.com/returntocorp/semgrep/issues/743
the issue itself is resolved, waiting for semgrep update

inkz:
unblocked

inkz:
https://dev.massive.ret2.co/triager/filter/1038

PR #470, opened 2020-05-09
add permissive-cors rule

inkz:
rule for: https://github.com/returntocorp/semgrep-rules/issues/457


inkz:
the rule tackles both Spring framework and general Servlet cases, so I do not know what section it suits better

minusworld (approved):
yesssss!!

inkz:
updated the rule to use regex in order to find `Access-Control-Allow-Origin` case insensitively and also tackle it's value if it contains `*` or is `Null` (case insensitively also)

minusworld (approved):
😂 It works!

I don't know if we have any new features available with `pattern-regex` but it might be worth looking in to for future rules: https://github.com/returntocorp/semgrep/issues/688

PR #481, opened 2020-05-12
nodejsscan rules 2020-05-12

ajinabraham:
nodejsscan rules and tests


minusworld:
Taking a look today

ajinabraham:
I moved the rules and tests to same directory. Now the tests are running. I think the failing test and this bug is https://github.com/returntocorp/semgrep/issues/738

javascript/nodejsscan/archive_path_overwrite.yaml
```diff
@@ -0,0 +1,154 @@
...
+      owasp: 'A5: Broken Access Control'
+      cwe: >-
+        CWE-22: Improper Limitation of a Pathname to a Restricted Directory
+        ('Path Traversal')
+  - id: tar_path_overwrite
+    patterns:
+      - pattern-inside: |
+          var $X = require('tar-stream');
+          ...
+      - pattern-inside: |
```
minusworld, line 124: Regarding https://github.com/returntocorp/semgrep/issues/738:
One thing we could do to make tests pass is add

```
      - pattern-not-inside: |
          $Y.pipe($UNZIP.Parse(...)).on('entry', function $FUNC(...) {
              ...
          }, ...);
```
here.  `tar_path_overwrite` is triggering on the `unzipper` test at the bottom of the file because it meets the requirements for 

```
- pattern-inside: |
          $EXTRACT.on('entry', function $FUNC(...) {
            ...
          }, ...);
```
because `$EXTRACT` can include an expression that matches `$SOMETHING.Parse(...)`.

However, the easier thing to do is just move the unzipper test up in the file. :)

In order to keep moving with these, I'm gonna move the test. If changes need to be made later, we can always follow up with new PRs. :)

ajinabraham: Cool. I fixed that.

PR #488, opened 2020-05-14
Label Security Rules

minusworld:
Add CWE and OWASP labels to all security rules currently in the repository.


go/lang/security/audit/net/wip-xss-using-responsewriter-and-printf.yaml
```diff
@@ -49,6 +49,9 @@ rules:
     Found data going from url query parameters into formatted data written to ResponseWriter.
     This could be XSS and should not be done. If you must do this, ensure your data is
     sanitized or escaped.
+  metadata:
+    cwe: "CWE-79: Improper Neutralization of Input During Web Page Generation ('Cross-site Scripting')"
+    owasp: "A1: Injection"
```
ulziibay, line 54: this is A7: Cross-Site Scripting XSS

java/lang/security/audit/bad-hexa-conversion.yaml
```diff
@@ -1,7 +1,10 @@
 rules:
 - id: bad-hexa-conversion
+  metadata:
+    cwe: "CWE-704: Incorrect Type Conversion or Cast"
+    owasp: "A3: Sensitive Data Exposure"
```
ulziibay, line 5: Maybe it is A6: Security Misconfiguration

python/boto3/security/hardcoded-token.yaml
```diff
@@ -1,5 +1,16 @@
 rules:
 - id: hardcoded-token
+  message: |
+    Hardcoded AWS access token detected. Use environment variables 
+    to access tokens (e.g., os.environ.get(...)) or use non version-controlled 
+    configuration files.
+  metadata:
+    cwe: "CWE-798: Use of Hard-coded Credentials"
+    owasp: "A3: Sensitive Data Exposure"
```
ulziibay, line 9: Lets map all CWE-798 to A2: Broken Authentication

python/flask/security/audit/hardcoded-config.yaml
```diff
@@ -4,68 +4,159 @@ rules:
...
+      - https://flask.palletsprojects.com/en/1.1.x/config/?highlight=configuration#environment-and-debug-features
     languages: [python]
     severity: WARNING
   # Rules for `SECRET_KEY` config
   - id: avoid_hardcoded_config_SECRET_KEY
     pattern: $M.config['SECRET_KEY'] = "=~/.*/"
     message: Hardcoded variable `SECRET_KEY` detected. Use environment variables or config files instead
+    metadata:
+      cwe: "CWE-798: Use of Hard-coded Credentials"
+      owasp: "A6: Security Misconfiguration"
```
ulziibay, line 70: Lets map all CWE-798 to A2: Broken Authentication

python/flask/security/audit/hardcoded-config.yaml
```diff
@@ -4,68 +4,159 @@ rules:
...
+      - https://flask.palletsprojects.com/en/1.1.x/config/?highlight=configuration#builtin-configuration-values
+      - https://flask.palletsprojects.com/en/1.1.x/config/?highlight=configuration#environment-and-debug-features
     languages: [python]
     severity: ERROR
   - id: avoid_hardcoded_config_SECRET_KEY
     pattern: $M.update(SECRET_KEY="=~/.*/")
     message: Hardcoded variable `SECRET_KEY` detected. Use environment variables or config files instead
+    metadata:
+      cwe: "CWE-798: Use of Hard-coded Credentials"
+      owasp: "A6: Security Misconfiguration"
```
ulziibay, line 82: Lets map all CWE-798 to A2: Broken Authentication

java/lang/security/audit/bad-hexa-conversion.yaml
minusworld: This one is hard to map, but I went with A3 because this isn't really a "configuration" issue. It weakens the crypto, so I binned it with other crypto checks. 🤷‍♂️

go/gorilla/security/audit/handler-assignment-from-multiple-sources.yaml
```diff
@@ -1,5 +1,9 @@
 rules:
   - id: handler-assignment-from-multiple-sources
+    metadata:
+      cwe: "CWE-289: Authentication Bypass by Alternate Name"
+      owasp: "A2: Broken Authentication"
+      owaspapi: "API1: Broken Object Level Authorization"
```
dlukeomalley, line 6: nit: thoughts on `owasp-api` vs. `owaspapi` :D

go/lang/security/audit/crypto/tls.yaml
```diff
@@ -6,7 +6,9 @@ rules:
     metadata:
       cwe: 'CWE-327: Use of a Broken or Risky Cryptographic Algorithm'
       owasp: 'A9: Using Components with Known Vulnerabilities'
-      source_rule_url: https://github.com/securego/gosec/blob/master/rules/tls.go
+      source-rule-url: https://github.com/securego/gosec/blob/master/rules/tls.go
```
dlukeomalley, line 9: Thoughts on `source-rule-url` being listed as a reference since these rules serve as inspiration and offer further reading?

go/gorilla/security/audit/handler-assignment-from-multiple-sources.yaml
minusworld: Thoughts are: Let's get this PR in and if it's important later, we can make a ticket to change

dlukeomalley: Ack. Once we start to use these fields (which will be immediately) it will be a pain to change. Agree with the general philosophy though.

go/lang/security/audit/crypto/tls.yaml
minusworld: Thoughts are: Let's get this PR in and if it's important later, we can make a ticket to change.  
We originally "standardized" on this as a key since we were porting rules over.
I figure we could do either of these later: 1) write a script to add these to references or 2) have code in the viz layer that adds these to references

dlukeomalley:
@minusworld Meta question - How do we ensure that the strings match for any givenn CWE or OWASP string? e.g. `CWE-289` has the same descriptor across all patterns that are CWE-289?

This could be something we handle in the frontend if say we took the convention `cwe: 289` verus the full string.

We'll use these values to generate our histograms. Want to make sure they're reliable.

minusworld:
> @minusworld Meta question - How do we ensure that the strings match for any givenn CWE or OWASP number? e.g. `CWE-289` has the same descriptor across all patterns that are CWE-289?
> 
> This could be something we handle in the frontend if say we took the convention `cwe: 289` versus the full string.
> 
> We'll use these values to generate our histograms. Want to make sure they're reliable.

Yeah, I think that's a good idea since it was a problem I ran into myself while annotating everything. Sometimes things would have quotes or a character would be missing due to a copy-pasta mistake.  I'll make a ticket to transition over to that.

PR #494, opened 2020-05-15
Java LDAP poisoning check

ulziibay:
Addresses https://github.com/returntocorp/semgrep-rules/issues/459
This add check for https://find-sec-bugs.github.io/bugs.htm#LDAP_ENTRY_POISONING

### DIFF
--- a/java/security/audit/ldap-entry-poisoning.yaml
+++ b/java/security/audit/ldap-entry-poisoning.yaml
@@ -1,13 +1,15 @@
 rules:
   - id: ldap-entry-poisoning
     metadata:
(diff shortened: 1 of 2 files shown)


java/security/audit/ldap-entry-poisoning.yaml
```diff
@@ -0,0 +1,21 @@
+rules:
+  - id: ldap-entry-poisoning
+    metadata:
+      cwe: 'A1: Injection'
```
minusworld, line 4: This is owasp 😅 
Can we add in the CWE as well?

PR #495, opened 2020-05-15
Tweak matrixify script

minusworld:
Normalizes OWASP output


java/lang/security/audit/el-injection.yaml
```diff
@@ -1,9 +1,9 @@
 rules:
 - id: el-injection
   metadata:
-    - cwe: "CWE-94: Improper Control of Generation of Code ('Code Injection')"
-    - owasp: "A1: Injection"
-    - source-rule-url: https://find-sec-bugs.github.io/bugs.htm#EL_INJECTION
```
ulziibay, line 6: I noticed the semgrep.live PR-s have dash under metadata. We should maybe still support `-` operator until those PR-s adhere to this.

minusworld: Maybe we can make a future PR for it when it becomes a problem. I think ultimately @dlukeomalley was right about standardizing on just numbers and converting to the full representation when consumed.

PR #504, opened 2020-05-16
(Linter assisted) Prune equivalent rules from httpresponse.yaml

rcoh:


python/django/security/injection/reflected-data-httpresponse.yaml
```diff
@@ -68,107 +68,7 @@ rules:
             ...
             django.http.HttpResponse(..., $INTERM, ...)
         - pattern: $A = django.http.HttpResponse(..., request.$W.get(...), ...)
```
minusworld, line 70: Looks like some of the one-liners got missed; either way, this still looks good!

rcoh: yeah the linter doesn't find those yet.

Issue #511, opened 2020-05-18
[Check] Insecure Deserialization checks for Python


minusworld:
**Check Description**
Insecure deserialization checks for Python, in target frameworks:
# Lang
Starting point: cf. https://bandit.readthedocs.io/en/latest/blacklists/blacklist_imports.html
## Pickling
- [x] dill https://github.com/returntocorp/semgrep-rules/pull/516
- [x] pickle https://github.com/returntocorp/semgrep-rules/pull/516
- [x] shelve https://github.com/returntocorp/semgrep-rules/pull/516
- [x] cPickle https://github.com/returntocorp/semgrep-rules/pull/516

## Yaml
- [x] PyYAML https://github.com/returntocorp/semgrep-rules/pull/525

# Django
- [x] https://github.com/returntocorp/semgrep-rules/issues/537 https://github.com/returntocorp/semgrep-rules/pull/544
# Flask
- [x] https://github.com/returntocorp/semgrep-rules/pull/532

# Per check
- [x] This ticket has links, references, or examples.
- [x] Your check has true positive and true negative test cases.
- [x] Your check has been labeled with true positive and false positive findings on at least 10 repositories, and the link is pasted in the ticket. If there are no findings, paste the triager link anyway.
- [x] Your check PR has been reviewed and merged.
Now, close the ticket!

ulziibay:
Job: https://dev.massive.ret2.co/jobs/1021?page=1&sort_on=repo_url&sort_by=asc

ulziibay:
Another one: https://dev.massive.ret2.co/jobs/1023?page=1&sort_on=repo_url&sort_by=asc

ulziibay:
More, https://dev.massive.ret2.co/jobs/1027?page=1&sort_on=repo_url&sort_by=asc,  https://dev.massive.ret2.co/jobs/1026?page=1&sort_on=repo_url&sort_by=asc, https://dev.massive.ret2.co/jobs/1025?page=1&sort_on=repo_url&sort_by=asc

PR #525, opened 2020-05-19
PyYAML check

ulziibay:
Add warning against yaml.load and recommend safe_load instead.

Addresses part of https://github.com/returntocorp/semgrep-rules/issues/511


minusworld (approved):
lgtm

PR #534, opened 2020-05-21
Add script for generating coverage graphs.

minusworld:
Script requires `pandas` and `matplotlib`.
The script doesn't belong here long term, but I'm hosting here now so it lives somewhere.


matrixify.py
```diff
@@ -13,25 +15,41 @@
...
-LANGS = {
-    "python",
-    "javascript",
-    "java",
-    "go",
-}
+class ArchList(list):
+    """
+    A list with a .get method that works like dict.get.
+    It's also very ancient and has dark magical powers.
```
mschwager, line 21: Ahaha, this is amazing

Time for a patch to the Python stdlib?

mschwager (commented):
Awesome, this is for generating the OWASP coverage by language graph right?

minusworld:
> Awesome, this is for generating the OWASP coverage by language graph right?

Yup! It's a script that will dump out the graph in the READMEs. It uses `pandas` and `matplotlib`. I thought I might make a little microservice for generating the graph automatically, but for now the code exists in this script.

matrixify.py
minusworld: 🙃

matrixify_graph.py
```diff
@@ -0,0 +1,79 @@
...
+    # Add arguments here
+
+    parser.add_argument("uri")
+    parser.add_argument("--save")
+    parser.add_argument("--img-tag", action="store_true")
+
+    args = parser.parse_args()
+
+
+    uri = sys.argv[1]
```
mschwager: I think you want `args.uri` here.

mschwager (approved):
LGTM

Issue #508, opened 2020-05-18
[Check] XXE checks for JavaScript


minusworld:
**Check Description**
XXE checks for JavaScript, if applicable.

- [x] This ticket has links, references, or examples.
- [x] Your check has true positive and true negative test cases.
- [x] Your check has been labeled with true positive and false positive findings on at least 10 repositories, and the link is pasted in the ticket. If there are no findings, paste the triager link anyway.
- [x] Your check PR has been reviewed and merged.
Now, close the ticket!

inkz:
`node-expat` module and it's dependent `xml2json` allow XXE in some way, (e.g. Billion Laugh Attack)
so the patterns to cover this situation were created

also the most popular module for xml parsing - `sax` was covered with the set of patterns in order to identify weak part wich is reading DTD definitions, which if handled irresponsibly may lead to XXE

https://www.npmjs.com/package/node-expat
https://www.npmjs.com/package/xml2json
https://www.npmjs.com/package/sax

inkz:
node-expat: https://dev.massive.ret2.co/jobs/1018

inkz:
sax: https://dev.massive.ret2.co/jobs/1017

inkz:
xml2json: https://dev.massive.ret2.co/jobs/1030

inkz:
@minusworld 
I guess that’s it for now, I'll do some triaging but checks are created, so we can close this task

minusworld:
@inkz Sounds good! 👍  These look legit 😎

inkz:
:+1:

PR #535, opened 2020-05-21
add expat-xxe rule

inkz:
Under https://github.com/returntocorp/semgrep-rules/issues/508


inkz:
the rule is intended to find input passed to `node-expat` library https://www.npmjs.com/package/node-expat
External Entity may not be a thing but it for sure allow "billion laughs" attack

minusworld (approved):
🤘

PR #539, opened 2020-05-22
add sax-xxe rule

inkz:
under https://github.com/returntocorp/semgrep-rules/issues/508

as stated in module description https://www.npmjs.com/package/sax#regarding-doctype-s-and-entity-s
developers have to parse Entities on their own, so the rule is intended to warn about possible weakness while parsing for DTDs


minusworld (approved):
lgtm!

PR #540, opened 2020-05-22
Normalize all rule YAML to the default ruamel output format

rcoh:
As requested, normalize YAML to:
1. Ruamel default indentation format
2. Line length of 120

If we want a slightly different format, lmk. Adds a pre-commit hook to do the formatting.


minusworld (approved):

ievans:
@DrewDennison can these rules also be applied to the defaults on semgrep.live and the default .semgrep.yml?

rcoh:
@ievans since these both use `ruamel` with the default settings, we _should_ be good. the default `.semgrep.yml` should be changed though

DrewDennison:
yes what Russell said

rcoh:
Just verified with an intentionally poorly formatted rule from semgrep.live -- pre-commit passes #543

ievans:
I meant if you go to semgrep.live in an incognito window, the flask-secure-set-cookie rule I don't think follows these formatting conventions. Minor nit.

PR #541, opened 2020-05-22
Autofix all the things

rcoh:
Run --autofix on semgrep-rules. This deletes from pattern-either:
- Patterns that are verbatim identical
- Patterns that have `return X` and `$A = X`

This also reduces the total time to run `semgrep --test` from 25s to 17s.


minusworld (approved):
\o/

PR #551, opened 2020-05-27
add xml2json-xxe rule

inkz:
For https://github.com/returntocorp/semgrep-rules/issues/508

this is intended to work for `xml2json` module which uses `node-expat` behind (https://github.com/returntocorp/semgrep-rules/tree/develop/javascript/node-expat/security/audit) so passing untrusted stuff in it has the same consequences


minusworld (approved):
Looks good

Issue #513, opened 2020-05-18
[Check] Insecure Deserialization checks for JavaScript


minusworld:
**Check Description**
Set of insecure deserialization checks for JavaScript;
Frameworks:

- Lang
- Express.js

- [x] This ticket has links, references, or examples.
- [x] Your check has true positive and true negative test cases.
- [x] Your check has been labeled with true positive and false positive findings on at least 10 repositories, and the link is pasted in the ticket. If there are no findings, paste the triager link anyway.
- [x] Your check PR has been reviewed and merged.
Now, close the ticket!

inkz:
presentation on deserialization vulnerabilities in general (Java, Node.js and Python)
https://www.slideshare.net/GreenD0g/deserialization-vulnerabilities
https://github.com/GrrrDog/ZeroNights-WebVillage-2017

blog on Insecure Deserialization in Node.js
https://opsecx.com/index.php/2017/02/08/exploiting-node-js-deserialization-bug-for-remote-code-execution/

modules that worth covering:
https://www.npmjs.com/package/node-serialize
https://github.com/yahoo/serialize-javascript

inkz:
job for grpc-nodejs-insecure-connection rule: https://dev.massive.ret2.co/jobs/1032

inkz:
* turned out that Ajin is the author of the research: https://opsecx.com/index.php/2017/02/08/exploiting-node-js-deserialization-bug-for-remote-code-execution/
small wonder that he had already created the rules
* possible `grpc` related issues are covered
* `serialize-javascript` does not have deserialization functionality, so nothing to cover
created a separate task for further research anyway https://github.com/returntocorp/semgrep-rules/issues/565

PR #557, opened 2020-05-30
grpc-nodejs-insecure-connection rule

inkz:
For: https://github.com/returntocorp/semgrep-rules/issues/513

with new semgrep js aliasing features I was able to tackle both client and server insecure credentials with one rule, so slightly updated the rule and added new test case


minusworld (approved):
🤘

PR #558, opened 2020-05-31
add express-xml2json-xxe-event rule

inkz:
`express` edition of the rule: https://github.com/returntocorp/semgrep-rules/blob/develop/javascript/xml2json/security/audit/xml2json-xxe.yaml

based on a few real findings


minusworld (approved):
👌

PR #566, opened 2020-06-02
Filter out a few Table Name cases using regular expressions.

minusworld:
This will reduce FP rate just slightly.


inkz (approved):
Awesome!

Issue #405, opened 2020-05-05
[Check] JavaScript: SSRF through headless browser request


inkz:
**Check Description**

Create set of rules to identify SSRF through headless browser libs like Puppeteer and PhantomJS

- [x] This ticket has links, references, or examples.
- [x] Your check has true positive and true negative test cases.
- [x] Your check has been labeled with true positive and false positive findings on at least 10 repositories, and the link is pasted in the ticket. If there are no findings, paste the triager link anyway.
- [x] Your check PR has been reviewed and merged.
Now, close the ticket!

inkz:
The rule is covering insecure usage of
https://www.npmjs.com/package/puppeteer
https://www.npmjs.com/package/wkhtmltopdf
https://www.npmjs.com/package/wkhtmltoimage
https://www.npmjs.com/package/phantom

and is based on previous reports
https://hackerone.com/reports/660565

PR #571, opened 2020-06-04
add puppeteer-injection rule

inkz:
rule for https://github.com/returntocorp/semgrep-rules/issues/405
covers general `puppeteer` usage


minusworld (approved):
🤘

PR #573, opened 2020-06-04
add phantom-injection rule

inkz:
for https://github.com/returntocorp/semgrep-rules/issues/405
covers `phantom` library which is the common wrapper for `phantomjs`


minusworld (approved):
so good!

PR #595, opened 2020-06-15
add jose-exposed-data rule

inkz:
the rule is intended to find when an object is converted to JWT token without explicitly striping sensitive data from it


inkz:
`jose` edition of https://github.com/returntocorp/semgrep-rules/pull/594

minusworld (approved):

PR #597, opened 2020-06-16
Fix jose exposed data.

minusworld:
Weird interactions between `pattern-inside` and `pattern-either` was causing Semgrep to go awry.

Included the Gatsby configuration file which caused this issue as a test case.


inkz (approved):
Oh, my bad with unnecessary `pattern-either`, thanks! :+1:

PR #603, opened 2020-06-16
Refactor python disabled verification rules.

minusworld:
Split up python disabled verification rules.

Previously there were three rules in one file. One of these rules used `pattern-where-python` which created complications for use of these rules in other rule packs.

This PR factors out these rules into their own files with their own tests.


minusworld:
Addresses https://github.com/returntocorp/enterprise/issues/240.

PR #614, opened 2020-06-18
add express-jwt-not-revoked rule

inkz:
rule for: https://github.com/returntocorp/semgrep-rules/issues/605


minusworld (approved):
Whoa, unrevokable tokens? Didn't even know to think about that 😬

PR #615, opened 2020-06-18
add rules for express default cookie settings

inkz:
rule for: https://github.com/returntocorp/semgrep-rules/issues/600


minusworld (approved):
nice 👍

PR #621, opened 2020-06-19
add require-request rule

inkz:
rule for: https://github.com/returntocorp/semgrep-rules/issues/606


minusworld:
I think that using `patter-either` with `pattern-inside` gets very weird... I am curious to see how these do running on platform. Could you try that out?

inkz:
@minusworld yep sure! I'll also try to extend examples to test different behaviour

inkz:
@minusworld 
platform: https://dev.massive.ret2.co/jobs/1086
updated test examples also

minusworld (approved):
Scan looks good, thanks!

Issue #421, opened 2020-05-06
[Check] JavaScript: Path Traversal


inkz:
**Check Description**
path traversal check inspired by
https://snyk.io/research/zip-slip-vulnerability
https://github.com/snyk/zip-slip-vulnerability

vulnerable code examples:
https://github.com/ZJONSSON/node-unzipper/pull/59
https://github.com/cthackers/adm-zip/pull/212

- [x] This ticket has links, references, or examples.
- [x] Your check has true positive and true negative test cases.
- [ ] Your check has been labeled with true positive and false positive findings on at least 10 repositories, and the link is pasted in the ticket. If there are no findings, paste the triager link anyway.
- [x] Your check PR has been reviewed and merged.
Now, close the ticket!

inkz:
https://dev.massive.ret2.co/jobs/1088

PR #645, opened 2020-06-20
update path-join-resolve-traversal rule

inkz:
update for: https://github.com/returntocorp/semgrep-rules/issues/421
I think the type of rule should be WARNING


minusworld (approved):
👍

Issue #422, opened 2020-05-06
[Check] JavaScript: RCE through vm.runInContext


inkz:
**Check Description**

check for insecure usage of https://nodejs.org/dist/latest-v12.x/docs/api/vm.html#vm_vm_runincontext_code_contextifiedobject_options

- [x] This ticket has links, references, or examples.
- [x] Your check has true positive and true negative test cases.
- [ ] Your check has been labeled with true positive and false positive findings on at least 10 repositories, and the link is pasted in the ticket. If there are no findings, paste the triager link anyway.
- [x] Your check PR has been reviewed and merged.
Now, close the ticket!

inkz:
https://www.npmjs.com/package/vm2
https://www.npmjs.com/package/sandbox
https://github.com/goldbergyoni/nodebestpractices/blob/master/sections/security/sandbox.md

inkz:
`vm-injection` rules: https://dev.massive.ret2.co/jobs/1094

inkz:
`express-vm-injection` rules: https://dev.massive.ret2.co/jobs/1095

inkz:
`sandbox-code-injection` rules: https://dev.massive.ret2.co/jobs/1096

inkz:
`express-sandbox-code-injection` rules: https://dev.massive.ret2.co/jobs/1100

inkz:
all rules together: https://dev.massive.ret2.co/jobs/1102

PR #648, opened 2020-06-22
add express-vm-injection rule

inkz:
express edition of https://github.com/returntocorp/semgrep-rules/issues/422


inkz:
https://dev.massive.ret2.co/jobs/1087

PR #655, opened 2020-06-24
Fix regression using latest semgrep

aryx:
I've recently added ';' in the generic AST, but this introduced
some regressions in semgrep-rules because the range changed
and if you're using pattern-not or pattern-inside, this has implications.
Indeed if one pattern use ';' and the other not, then the range are
different and the intersection or negation might not trigger.

I didn't anticipate this problem ... It might be better to remove
back those ';' in the generic AST, but at the same time it also
allows better autofixing in the futur and better matching display
(we will highlight the ';' in semgrep.live if your pattern uses one).

Test plan:
make test (with latest semgrep-core locally;
hopefully the docker image will catch up)

### DIFF
--- a/contrib/nodejsscan/header_cors_star.yaml
+++ b/contrib/nodejsscan/header_cors_star.yaml
@@ -16,7 +16,7 @@ rules:
 - id: express_cors
   patterns:
   - pattern-inside: |
-      $APP.$METHOD(..., function $FUNC($REQ, $RES, ...) { ... })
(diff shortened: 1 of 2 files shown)


ievans:
Thanks for the writeup @aryx . This is going to be a tricky issue. At least now we have step-by-step debugging so it's easier to see where the range filter is or isn't working.

aryx:
Yep, i actually used -v to trace the issue

aryx:
Which i learned about thx to the weekly demo video!

PR #682, opened 2020-06-29
Fix duplicate pattern / patterns key

ievans:


minusworld (changes requested):
There are two rules in that file; are you trying to remove one of the rules?

PR #685, opened 2020-07-02
Add basic regex fixes using fix-regex key

### LINKED ISSUE #596: Add trivial autofixes to checks that support them.
- [x] Identify fixes for which trivial autofixes will work.
- [x] Create a test suite of actual vs. expected for each.
- [x] Add this to CI.
- [x] Add in ability to do regex replace in semgrep autofixes >.>
- [x] Make PR for these; when accepted, close issue.
- [ ] Highlight some of them on the semgrep.live examples dropdown

minusworld:
Closes https://github.com/returntocorp/semgrep-rules/issues/596.

Note that this breaks compatibility with Semgrep < 0.12.0.


inkz (approved):
it works!

PR #710, opened 2020-07-10
Update rule messages.

### LINKED ISSUE #701: Update rule messages for rules that are in Packs to follow: What the Finding Is, Why It's a Problem, and How to Fix It
What the finding is. Describe the code that is an issue.
Why it's a problem. Describe why this code is a problem, including effects of not fixing it.
Solution: Propose a solution.

Example:
`python.django.security.injection.code.user-exec.user-exec`

> Found request data as an index to 'globals()'. This is extremely dangerous because it allows an attacker to execute arbirary code on the system. Refactor your code not to use 'globals()'.

minusworld:
Update rule messages for rules that are in featured rule packs.
Closes https://github.com/returntocorp/semgrep-rules/issues/701.


ievans (approved):
So so good. Are you going to do pack descriptions too? :D

python/lang/correctness/list-modify-iterating.yaml
```diff
@@ -1,5 +1,5 @@
 rules:
-- id: list-pop-while-iterate
+- id: list-modify-while-iterate
```
mschwager, line 2: Note that every time we change a rule ID we're breaking backwards compatibility. Usage of `nosem` with the old rule ID will now break. This is why I'd argue for rule IDs that don't mean anything like `PY-SEC-0001` or `JS-COMP-0025`. Rule IDs or error codes aren't necessarily meant to mean anything by themselves, that's what the message and corresponding documentation is for. The ID is more of a easily searchable string used to look up more information, ignore or select rules, inline ignore, etc.

dlukeomalley: Our research disagrees here @mschwager. From a technical perspective I agree and understand how painful this is for back compat (we should be slow to change IDs), but during Bento interviews we consistently heard that users are frustrated by tools using meaningless names. I thing this largely stems from having to Google names like `PY-SEC-0001` or `E101` rather than understanding the rule solely on its own.

Our takeaway is that Rule IDs *should* mean something by themselves.

minusworld: That's a good flag, though, @mschwager . I didn't really think of that. I just updated the rule to be slightly more correct.

Are we using it somewhere now with nosem? If not, changing things now might be better 🤷‍♂️

mschwager: > Our research disagrees here @mschwager. From a technical perspective I agree and understand how painful this is for back compat (we should be slow to change IDs), but during Bento interviews we consistently heard that users are frustrated by tools using meaningless names.

PR #712, opened 2020-07-10
Eslint plugin security

### LINKED ISSUE #683: Add equivalent eslint-plugin-security rules
_Don't complete this - reserving for new hire task._

We've recently begun porting common static analysis/linter tool rules over to `semgrep`. This makes for an easy transition if a user/customer is already using this tool. For example, we've already ported the following tools:

* Golang's [`gosec`](https://github.com/securego/gosec) - https://semgrep.live/p/gosec
* Java's [`find-sec-bugs`](https://github.com/find-sec-bugs/find-sec-bugs) - https://semgrep.live/p/findsecbugs
* Python's [`bandit`](https://github.com/openstack-archive/bandit) - https://semgrep.live/p/bandit

Similarly, let's create an equivalent checkpack for [`eslint-plugin-security`](https://github.com/nodesecurity/eslint-plugin-security). The goal of this task is to port over rules from `eslint-plugin-security` to `semgrep`. This can be accomplished by creating the appropriate rules in Javascript security directory: https://github.com/returntocorp/semgrep-rules/tree/develop/javascript/lang/security, similar to `find-sec-bugs` in the Java security directory: https://github.com/returntocorp/semgrep-rules/tree/develop/java/lang/security.

Briefly looking through the rules, we probably don't need all of them. We want to balance coverage vs. potential for false positives. I.e. people don't like incorrect, noisy rules. Looking through the list I would say we can avoid the following because they may be too noisy:

* detect-unsafe-regex
* detect-non-literal-fs-filename
* detect-non-literal-regexp
* detect-possible-timing-attacks

sabrinabrogren:
Implemented the following rules from [eslint-plugin-security](https://github.com/nodesecurity/eslint-plugin-security):

- detect-buffer-noassert
- detect-child-process
- detect-disable-mustache-escape
- detect-eval-with-expression
- detect-no-csrf-before-method-override
- detect-non-literal-require
- detect-pseudoRandomBytes

Others omitted to minimize false positives. (Very) basic tests included.


mschwager:
LGTM! What do you think @minusworld?

minusworld (approved):
these look great!

mschwager (approved):

PR #730, opened 2020-07-16
Add Ruby check secrets

clintgibler:


minusworld (approved):

minusworld:
Oh, looks like `pattern-where-python` needs to explicitly return a boolean

clintgibler:
```
python where expression needs boolean output but got: <re.Match object; span=(0, 8), match='password'> for import re
re.match(r"password|secret|(rest_auth_site|api)_key$", vars['$X'].lower())
```

mschwager:
I believe we can better achieve this will `metavariable-regex` now. Then there's no need for code execution.

ievans:
@clintgibler want to fix or close?

minusworld:
Closing due to inactivity. Reopen this PR if still relevant.

PR #737, opened 2020-07-22
Add 2 new ruby rules

minusworld:
- missing-csrf-protection
- hardcoded auth in controller


minusworld:
@clintgibler Could you take a quick look at these to see if they make sense? Particularly, I'm wondering if the class inheritance looks OK. 🤷‍♂️

PR #741, opened 2020-07-24
Fix typo in rule message

dee-see:


ajinabraham (approved):

dlukeomalley:
@ajinabraham @DrewDennison Should this fix live in semgrep-rules or an upstream repo?

ajinabraham:
The latest rules are here https://github.com/ajinabraham/njsscan/blob/master/njsscan/rules/semantic_grep/crypto/tls_node.yaml

Both the ones here and at https://github.com/semgrep/rules-nodejsscan are slightly outdated.

As I understand correctly `contrib/nodejsscan` should moved to `rules-nodejsscan`.

PR #746, opened 2020-07-29
This TODO is automatically implemented by Semgrep 0.18

nbrahms:


aryx:
@nbrahms there's a mutual dependency issue between semgrep and semgrep-rules. It's annoying but I usually
first force the fix in semgrep (even if it fails semgrep-rules), and later fix semgrep-rules.

aryx:
Note that we could also improve the test on semgrep-rules to not report an error when a new match is a todoruleid!
The main issue is that we may forget to change those todoruleid in ruleid, but maybe we can have an email or a warning
to tell people to update semgrep-rules. I think it would be better than blocking PRs in semgrep just because semgrep-rules has a todoruleid for something we just fixed ...

nbrahms:
I filed https://github.com/returntocorp/semgrep/issues/1386

PR #753, opened 2020-07-29
fix source-rule-url for httponly cookie semgrep rule

sfc-gh-spandey:


dlukeomalley:
Thank you for the contribution @sfc-gh-spandey! 🙏

PR #755, opened 2020-07-29
Ruby cookie serialization

colleend:


colleend:
@mschwager  should we also have the :hybrid check? Technically not a vuln, but more of a best-practice check.

PR #790, opened 2020-08-14
add playwright-exposed-chrome-devtools rule

inkz:
checking for exposed chrome debugger
https://github.com/nccgroup/singularity/blob/master/html/payloads/exposed-chrome-devtools.js


minusworld (approved):
🤘

PR #803, opened 2020-08-17
Add new rule for detecting vulnerable yaml parsing in Ruby

minusworld:


colleend:
looks awesome! should we also add a reference to the specific brakeman check in metadata?

colleend (approved):

PR #810, opened 2020-08-22
Add PostMessage Security Rules

snoopysecurity:
This PR contains two rules related to Postmessage security. One for postMessage API wilcard issues and another for addEventListener API not checking origin. The addEventListener rule pattern will have False Negative cases since I am only checking for two common semantic patterns.

### References
* https://labs.detectify.com/2016/12/08/the-pitfalls-of-postmessage/
* https://docs.ioin.in/writeup/www.exploit-db.com/_docs_40287_pdf/index.pdf

### DIFF
--- a/javascript/browser/security/insecure-postmessage-configuration.js
+++ b/javascript/browser/security/insecure-postmessage-configuration.js
@@ -1,0 +1,37 @@
+let data={pName : "Bob", pAge: "35"};
+var popup = window.open(/* popup details */);
+
+//ruleid:wildcard-postmessage-configuration
+popup.postMessage(data, '*');
+//ruleid:wildcard-postmessage-configuration
+popup.postMessage( JSON.stringify( data ), '*' );
+
+//postMessage Safe Usage
+popup.postMessage("hello there!", "http://domain.tld");    
+popup.postMessage( JSON.stringify( data ), 'semgrep.dev/editor');
+
+//ruleid:insufficient-postmessage-origin-validation
+window.addEventListener("message", receiveMessage, false);
+
+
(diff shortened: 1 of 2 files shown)


salecharohit:
Hi @snoopysecurity 
You'll need to add annotations like the way i've added in this file here [// ruleid:owasp.java.ssrf.possible.import.statements](
https://github.com/returntocorp/semgrep-rules/blob/14f1bb01f3f1994b27387a85671ee8ddf9ba6dcb/contrib/owasp/java/ssrf.java#L4) corresponding to the rule-ids that you are creating.
Checkout the SSRF rule in action [here](https://semgrep.dev/3eee) and see that I've added the annotations at all the places where a rule is matching.

snoopysecurity:
good spot @salecharohit, didn't realise that was the reason i was failing pre-commit tests. But i see tests are still failing regarding line numbers, but i see the same thing with all the opened PRs so not sure if this is an issue I should be trying to fix

minusworld:
> good spot @salecharohit, didn't release that was the reason i was failing pre-commit tests. But i see tests are still failing regarding line numbers, but i see the same thing with all the opened PRs so not sure if this is an issue I should be trying to fix

I'll take a look here in a bit!

minusworld:
Ah OK, I think I see what it is -- if you remove the `ruleid:` prefix from the YAML file, it should work! The YAML file just needs the rule ID, which can be thought of as a name.

The comment syntax `ruleid:` is for annotating source code files (Python, JS, etc.) and it lets the testing framework know that Semgrep should report errors there.

snoopysecurity:
awesome! thanks @minusworld

PR #811, opened 2020-08-24
Express http

colleend:
Checks for usage of http servers instead of https servers. Not sure if this is actually a good rule, because it may give too many false positives. I wrote it without thinking about the FP rate, so if its not useful we can discard.


colleend:
not sure if this check is actually good -- there might be high FP rate, and some app-writers might also want to serve http but redirect all requests to https.

minusworld:
Yeah, we may run into some issues like that occasionally. Since we'll be shipping it in packs, ostensibly they can pick-and-choose specific things for their use-case. Hopefully that mitigates the issue.

If you're interested in spot-checking how the rule does on real code, consider giving it a spin on platform! dev.massive.ret2.co

problem-based-packs/insecure-transport/js-express/using-http-server.yaml
```diff
@@ -0,0 +1,21 @@
...
+    - pattern: |
+        http.$FUNC(...);
+    - pattern: |
+        new http.$FUNC2(...);
+  message: |
+    Checks for any usage of http servers instead of https servers. Encourages the usage of https protocol instead of 
+    http, which does not have TLS and is therefore unencrypted. Using http can lead to man-in-the-middle attacks 
+    in which the attacker is able to read sensitive information.
+  metadata:
+    vulnerability: Insecure transport
```
minusworld: This is probably OWASP 2017 Top 10: A3 Sensitive Data Exposure.
Would you also be able to add in the CWE? You can just Google "CWE http" or something simlar.

colleend: yes definitely!!

colleend:
> Yeah, we may run into some issues like that occasionally. Since we'll be shipping it in packs, ostensibly they can pick-and-choose specific things for their use-case. Hopefully that mitigates the issue.
> 
> If you're interested in spot-checking how the rule does on real code, consider giving it a spin on platform! dev.massive.ret2.co

seems like there's not too many false positive rates after scanning!

PR #812, opened 2020-08-25
Add unsafe ruamel.yaml usage rule

mschwager:
Our Python YAML deserialization rule looks for PyYAML but not `ruamel.yaml`. To support internal use-cases let's check for unsafe `ruamel` too.


python/lang/security/deserialization/avoid-unsafe-ruamel.py
```diff
@@ -0,0 +1,16 @@
+from ruamel.yaml import YAML
+
+#ok:avoid-unsafe-ruamel
```
mschwager, line 3: I'm planning on adding true negative support to `semgrep --test` shortly, but it's not in yet.

minusworld (approved):

PR #815, opened 2020-08-26
Add flask xss rules

minusworld:
This adds Flask XSS Secure Defaults rules. Closes https://github.com/returntocorp/enterprise/issues/306.

These rules are dependent on [this PR](https://github.com/returntocorp/semgrep/pull/1584) to scan template files using pattern-regex. This should launch in 0.22.0, so in the meantime these rules can hang out here. :)


minusworld:
Whoops, wrong branch.

PR #816, opened 2020-08-26
Http request

colleend:
note: protocol for options is normally http by default.


problem-based-packs/insecure-transport/js-node/http-request.js
```diff
@@ -0,0 +1,78 @@
+function bad_http() {
+    // ruleid: http-request
+    http.get('http://nodejs.org/dist/index.json', (res) => {
+    const { statusCode } = res;})
+
+    // ruleid: http-request
+    https.get('http://nodejs.org/dist/index.json', (res) => {
```
minusworld, line 7: Does this actually go over HTTP? 😱

colleend: you're totally right!! I don't think it actually goes over http :o

minusworld (approved):

problem-based-packs/insecure-transport/js-node/http-request.js
minusworld: We could test using wireshark 🤔   🦈

colleend: thats a good idea!!

colleend: I'll first push this, if requests to 'http...' through the https module actually doesn't encrypt traffic, i'll update it back to the OG 💯

colleend: tested -- node actually throws "TypeError [ERR_INVALID_PROTOCOL]: Protocol "http:" not supported. Expected "https:" error if you attempt to request an http website.

PR #817, opened 2020-08-26
Add flask xss rules

minusworld:
This adds Flask XSS Secure Defaults rules. Closes returntocorp/enterprise#306.

These rules are dependent on this PR to scan template files using pattern-regex. This should launch in 0.22.0, so in the meantime these rules can hang out here. :)


colleend:
ohhh awesome!

colleend (approved):

PR #820, opened 2020-08-27
Go xss rules

minusworld:
Add Go XSS rules for the standard http server.
Closes https://github.com/returntocorp/enterprise/issues/310.
Covers:
- text/template
- html/template issues
- a few dangerous patterns in Go templates that escaping doesn't immediately solve


go/lang/security/audit/xss/template-html-does-not-escape.yaml
```diff
@@ -0,0 +1,20 @@
...
+  metadata:
+    cwe: "CWE-79: Improper Neutralization of Input During Web Page Generation ('Cross-site Scripting')"
+    owasp: 'A1: Injection'
+    references:
+    - https://golang.org/pkg/html/template/#HTML
+    - https://github.com/0c34/govwa/blob/139693e56406b5684d2a6ae22c0af90717e149b8/vulnerability/xss/xss.go#L33
+  languages: [go]
+  severity: WARNING
+  patterns:
+  - pattern-not: template.HTML("..." + "...")
```
colleend, line 18: could there be an addition of more than 2 constants?

minusworld: Yeah, but this stemmed from a platform job which revealed a handful of repositories doing this. Basically this is just here to reduce FP rate by a small amount.

PR #832, opened 2020-09-01
added xxe and ssrf folders

salecharohit:
Added xxe and ssrf folders under the ~/contrib/owasp/java folder for better identification


minusworld (approved):
I  ❤️ organization

PR #842, opened 2020-09-05
Add rule to detect eval in browser javascript

TMFRook:
Adds a semgrep rule  to detect `eval` in the browser/javascript context.


minusworld (approved):
This is awesome! Thanks so much! 🙏

PR #853, opened 2020-09-10
Django xss rules

minusworld:
Follows this policy:

# XSS Prevention in Django

Conditions:

1. User input
2. Data manually rendered to end user
3. Data unescaped in template context
4. Data rendered to dangerous location in template

### 2. Data manually rendered to end user

1. Directly writing to `HttpResponse` and friends
    1. https://r2c.dev/blog/2020/be-careful-what-you-request-for-django-method/

### 3. Data unescaped in template context

1. Disabling autoescape
    1. Global disable in DjangoTemplates settings
        1. https://docs.djangoproject.com/en/3.1/topics/templates/#django.template.backends.django.DjangoTemplates
    2. `autoescape=False` in a template Context
        1. [cf. https://github.com/django/django/blob/54ea290e5bbd19d87bd8dba807738eeeaf01a362/django/template/context.py#L135](https://github.com/django/django/blob/54ea290e5bbd19d87bd8dba807738eeeaf01a362/django/template/context.py#L135)
        2. Template.render
            1. https://docs.djangoproject.com/en/3.1/ref/templates/api/#django.template.Template.render
        3. render_to_string
            1. https://docs.djangoproject.com/en/3.1/topics/templates/#django.template.loader.render_to_string
        4. render shortcut
            1. https://docs.djangoproject.com/en/3.1/topics/http/shortcuts/#django.shortcuts.render
2. Marking “Safe” content
    1. Python
        1. `mark_safe`
        2. `SafeString` class
            1. https://docs.djangoproject.com/en/3.1/howto/custom-template-tags/
        3. `is_safe=True` with `@register.filter(...)`
        4. Conditional escape: `@register.filter(needs_autoescape=**True**)` with `def initial_letter_filter(text, autoescape=**False**):`
            1. Uncertain of this one???
        5. `__html__`???
        6. `html_safe` on a class???
    2. Templates
        1. `| safe` and `| safeseq`
        2. `{% autoescape off %}` block

### 4. Data rendered to dangerous location in template

1. Unquoted HTML attributes
    1. cf. https://docs.djangoproject.com/en/3.1/topics/security/#cross-site-scripting-xss-protection
2. Probably `href`, similar to Flask
3. JS template strings? (like in Go)
4. tags? (like in Go)

## Mitigations

1. Ban `{% autoescape off %}`
    1. Prefer `mark_safe` only if necessary
2. Ban `| safe`
    1. Prefer `mark_safe` only if necessary
3. Ban `| safeseq`
    1. Prefer `mark_safe` only if necessary
4. Ban `mark_safe`
    1. Review on a case-by-case basis and whitelist using `# nosem`
5. Ban class extensions from `SafeString`
    1. Prefer `mark_safe` only if necessary
6. Ban `is_safe=True` with `@register.filter`
    1. Prefer `mark_safe` only if necessary
7. Ban conditional escape logic in custom filter?
8. Ban directly writing to `HttpResponse` and friends (except perhaps if doing JSON or an API, or something similar)
    1. Use `render()`
9. Ban unquoted HTML attributes
10. Templates variables in `href` must use the `url` template function
    1. cf. https://docs.djangoproject.com/en/3.1/intro/tutorial03/#removing-hardcoded-urls-in-templates
11. Ban `autoescape=False` in template Contexts
12. Ban global autoescape off

|Control	|Description	|Semgrep Rule	|
|---	|---	|---	|
|1	|Ban `autoescape off`	|python.django.security.audit.xss.template-autoescape-off	|
|2	|Ban `\| safe`	|python.flask.security.xss.audit.template-unescaped-with-safe.template-unescaped-with-safe	|
|3	|Ban `\| safeseq`	|python.django.security.audit.xss.template-var-unescaped-with-safeseq	|
|4	|Ban `mark_safe`	|python.django.security.audit.avoid-mark-safe.avoid-mark-safe	|
|5	|Ban class extensions from `SafeString`	|python.django.security.audit.xss.class-extends-safestring	|
|6	|Ban `is_safe=True` with `@register.filter`	|python.django.security.audit.xss.filter-with-is-safe	|
|7	|Ban escape logic in custom filters?	|TBD	|
|8	|Ban directly writing to HttpResponse+friends	|python.django.security.audit.xss.direct-use-of-httpresponse	|
|9	|Ban unquoted HTML attributes	|python.flask.security.xss.audit.template-unquoted-attribute-var.template-unquoted-attribute-var	|
|10	|Require `url` for template vars in `href`	|python.django.security.audit.xss.template-href-var	|
|11	|Ban `autoescape=False` in template Contexts	|python.django.security.audit.xss.context-autoescape-off	|
|12	|Ban global autoescape off	|python.django.security.audit.xss.global-autoescape-off	|

### DIFF
--- a/python/django/security/audit/xss/global-autoescape-off.yaml
+++ b/python/django/security/audit/xss/global-autoescape-off.yaml
@@ -1,0 +1,21 @@
+rules:
+- id: global-autoescape-off
+  message: |
+    Autoescape is globally disbaled for this Django application. If you are
+    rendering any web pages, this exposes your application to cross-site
+    scripting (XSS) vulnerabilities. Remove 'autoescape: False' or set it
+    to 'True'.
+  metadata:
+    cwe: "CWE-79: Improper Neutralization of Input During Web Page Generation ('Cross-site Scripting')"
+    owasp: 'A7: Cross-site Scripting (XSS)'
+    references:
+    - https://docs.djangoproject.com/en/3.1/ref/settings/#templates
+    - https://docs.djangoproject.com/en/3.1/topics/templates/#django.template.backends.django.DjangoTemplates
+  languages:
+  - python
+  severity: WARNING
+  pattern: |
+    {..., 'BACKEND': ..., 'OPTIONS': {..., 'autoescape': False, ...}, ...}
+  fix-regex:
+    regex: (autoescape.*?)False
+    replacement: \1True
--- a/python/django/security/audit/xss/class-extends-safestring.py
+++ b/python/django/security/audit/xss/class-extends-safestring.py
@@ -1,12 +1,17 @@
-from django.utils.safestring import SafeString, SafeData
+from django.utils.safestring import SafeString, SafeData, SafeText
 
 # ruleid:class-extends-safestring
-class IWantToBypassEsccaping(SafeString):
+class IWantToBypassEscaping(SafeString):
+    def __init__(self):
+        super().__init__()
+        
+# ruleid:class-extends-safestring
+class IWantToBypassEscaping2(SafeText):
     def __init__(self):
         super().__init__()
 
 # ruleid:class-extends-safestring
-class IWantToBypassEsccaping2(SafeData):
+class IWantToBypassEscaping3(SafeData):
     def __init__(self):
         super().__init__()
 
--- a/python/django/security/audit/xss/class-extends-safestring.py
+++ b/python/django/security/audit/xss/class-extends-safestring.py
@@ -1,0 +1,16 @@
+from django.utils.safestring import SafeString, SafeData
+
+# ruleid:class-extends-safestring
+class IWantToBypassEsccaping(SafeString):
+    def __init__(self):
+        super().__init__()
+
+# ruleid:class-extends-safestring
+class IWantToBypassEsccaping2(SafeData):
+    def __init__(self):
+        super().__init__()
+
+# ok:class-extends-safestring        
+class SomethingElse(str):
+    def __init__(self):
+        super().__init__()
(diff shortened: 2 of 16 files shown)


python/django/security/audit/xss/global-autoescape-off.yaml
```diff
@@ -0,0 +1,21 @@
...
+    cwe: "CWE-79: Improper Neutralization of Input During Web Page Generation ('Cross-site Scripting')"
+    owasp: 'A7: Cross-site Scripting (XSS)'
+    references:
+    - https://docs.djangoproject.com/en/3.1/ref/settings/#templates
+    - https://docs.djangoproject.com/en/3.1/topics/templates/#django.template.backends.django.DjangoTemplates
+  languages:
+  - python
+  severity: WARNING
+  pattern: |
+    {..., 'BACKEND': ..., 'OPTIONS': {..., 'autoescape': False, ...}, ...}
```
colleend, line 18: can backend be something other than 'django.template.backends.django.DjangoTemplates' in this case?

colleend (approved):
lgtm!

python/django/security/audit/xss/global-autoescape-off.yaml
minusworld: Yes, it can be "Jinja2" which uses Jinja2 as the template engine. However, if my read of the docs is correct, the 'OPTIONS' can still include `'autoescape': False`.

See a few lines into https://docs.djangoproject.com/en/3.1/topics/templates/#django.template.backends.jinja2.Jinja2 for where I'm getting this info. Hopefully my read is correct.

colleend: ah yes makes sense!

PR #859, opened 2020-09-14
Java spring

colleend:


colleend:
Looked at https://snyk.io/blog/spring-boot-security-best-practices/, and did some more research on requiring https -- apparently this can be done in many ways, so its hard to generalize. I'll look into this probably after launch, as it might take some more research 😱

minusworld (approved):
Yeah, if we hit the common ones, it shoudl get us far enough along 👍  lgtm!

PR #860, opened 2020-09-14
Fix java xss rules

minusworld:
Reduce FPs in not directly writing to responsewriter rule.
Also added a new rule for detecting `<c:out>` tags in templates but this will not be used as a part of the XSS pack. It's here for completeness. (... And because I realized how to do the right regex :| )


inkz (approved):
:+1:

PR #864, opened 2020-09-16
Secure defaults XSS rules for Express.js

minusworld:
# XSS Prevention in Express.js

Conditions:

1. User input (reflected, stored)
2. Variable sent outside of template context
3. Variable is unescaped in template
4. Variable in dangerous location in template

### 2. Variable sent outside of template context

* Sending data directly to user
    * `res.send(...)`
    * `res.write(...)`
        * use `res.render(<template>, ...)` instead

### 3. Variable is unescaped in template

* Jade & Pug
    * `!=` operator
    * `!{this}` ← that syntax
        * https://stackoverflow.com/questions/6926247/nodejs-jade-escape-markup
        * https://pugjs.org/language/attributes.html#unescaped-attributes
    * & attributes
        * https://pugjs.org/language/attributes.html#attributes
* Mustache & Handlebars
    * triple mustache `{{{ ... }}}`
    * You can also use `&` to unescape a variable.
    * If you'd like to change HTML-escaping behavior globally (for example, to template non-HTML formats), you can override Mustache's escape function. For example, to disable all escaping: `Mustache.escape = function(text) {return text;};`.
* EJS
    * `<%- ... %>` 
        * cf. http://www.managerjs.com/blog/2015/05/will-ejs-escape-save-me-from-xss-sorta/

### 4. Variable in dangerous location in template

* variable in `href` attribute.
    * Jade & Pug
        * *Is there a way to do something like `url_for` in Flask?*
        * Sort of... start with a slash, like this: `a(href='/' + link)='hello'`
            * cf. https://github.com/pugjs/pug/issues/2952
    * Mustache
        * <a href=“{{ ... }}”>
            * Suggest <a href=“/{{ ... }}”> ← slash in front
    * EJS
        * <a href=“<%= ... %>”>
            * Suggest <a href=“/<%= ... %>”> ← slash in front
* In a JavaScript context
    * https://www.veracode.com/blog/secure-development/nodejs-template-engines-why-default-encoders-are-not-enough

## Mitigations

1. Ban directly writing to Response objects.
    1. Prefer rendering in templates.
2. Ban global disable.
    1. Pug
        1. None?
    2. Mustache
        1. Ban `Mustache.escape = function(text) {return text;};`
    3. EJS
        1. None?
3. Flag template unescapes
    1. Pug
        1. Flag `!=`
        2. Flag `!{...}`
        3. Flag `&attributes`
    2. Mustache
        1. Flag `{{{...}}}`
        2. Flag `&` to escape vars
    3. EJS
        1. Flag `<%- ... %>`

### DIFF
--- a/javascript/express/security/audit/xss/direct-response-write.js
+++ b/javascript/express/security/audit/xss/direct-response-write.js
@@ -1,0 +1,119 @@
+
+const express = require('express')
+const router = express.Router()
+
+router.get('/greeting', (req, res) => {
+    const { name } = req.query;
+    // ruleid: direct-response-write
+    res.send('<h1> Hello :' + name + "</h1>")
+})
+
+//template handle escaping 
+router.get('/greet-template', (req, res) => {
+    name = req.query.name
+    // ok: direct-response-write
+    res.render('index', { user_name: name });
+})
+
+module.exports = router
+
+
+app.get('/', function (req, res) {
+    var user = req.query.name;
+
+    msg = "Hi " + user
+    // ruleid: direct-response-write
+    res.send('Response</br>' + msg);
+});
+
+
+var msg = '';
+app.get('/3', function (req, res) {
+    var user = req.query.name;
+
+    msg = "Hi " + user
+    // ruleid: direct-response-write
+    res.send('Response</br>' + msg);
+});
+
+app.get('/2', function (req, res) {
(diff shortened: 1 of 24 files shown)


minusworld:
Closes https://github.com/returntocorp/enterprise/issues/312.

inkz (approved):
awesome work :+1:

PR #867, opened 2020-09-18
Java servlets

colleend:


java/servlets/security/cookie-issecure-false.yaml
```diff
@@ -0,0 +1,27 @@
+rules:
+- id: cookie-issecure-false
+  patterns:
+  - pattern-either:
```
mschwager: Nit: we don't need this `pattern-either`. It makes it a little confusing as to which patterns are being `AND`'ed.

java/servlets/security/cookie-issecure-false.yaml
```diff
@@ -0,0 +1,27 @@
...
+        ...
+  - pattern-not-inside: |
+      $COOKIE = new Cookie(...);
+      ...
+      $COOKIE.setSecure(true);
+  message: |
+    Default session middleware settings: `setSecure` not set to true. 
+    This ensures that the cookie is sent only over HTTPS to prevent cross-site scripting attacks.
+  fix-regex:
+    regex: 'false'
```
mschwager: Should we be a little tighter on the bounds where `false` can occur? E.g. `\(false\)` -> `\(true\)`? My concern is that any `false` that occurs in the `pattern` via the ellipsis operator will be fixed, not necessarily just a `setSecure(false)`.

mschwager:
Is this PR supposed to have 67 commits? I think you need to rebase on `develop`.

java/servlets/security/cookie-issecure-false.yaml
colleend: good point -- I'll change this.

colleend:
I've been having issues rebasing on develop 😱 but i'll try again after this

PR #875, opened 2020-09-22
fix FPs in explicit-unescape for Ejs and Mustache in JavaScript

inkz:
in Ejs and Mustache the common case for using raw output for including parts of the template, e.g.
```
 <%- include('user/show', {user: user}); %>
```


colleend (approved):
lgtm

PR #876, opened 2020-09-22
fix FPs for Ejs and Mustache in JavaScript

inkz:
1. Mustache rules should also apply to Handlebars templates, so I added coverage for `.hbs` files
from docs: (and experience :))
https://handlebarsjs.com/
```
Handlebars is largely compatible with Mustache templates. In most cases it is possible to swap out Mustache with Handlebars and continue using your current templates.
```

2. lots of false positives on the platform for situations like this:
```html
<script src="app.js"></script>
<%= title %>
<script src="script.js"></script>
```


minusworld (approved):
nice. I was having a hard time getting var-in-script-tag to work!!

inkz:
updated regexes :+1:

minusworld:
lgtm 👍

PR #884, opened 2020-09-22
insecureskipverify rule, strange we didn't have that before

colleend:


minusworld (approved):
lol, yeah, I guess we just had snippets of them but never official rules. 🤷‍♂️

inkz (approved):
:+1:

PR #889, opened 2020-09-24
update disallow-old-tls rules for nodejs

inkz:
crypto constats in Node.js also can be found in a `constants` module, which is deprecated, but still used
https://github.com/nodejs/node/blob/master/lib/constants.js


colleend (approved):
YASSS

PR #891, opened 2020-10-01
Ignore NullBooleanField for django's nontext field must set null=True

thornycrackers:
I know NullBooleanField is being deprecated in 3.1 but older projects can still use it so probably best to have it ignored by the rule.


minusworld (approved):
Cool, looks good! Thank you!

PR #896, opened 2020-10-05
sqlalchemy sql injection rules

colleend:
would be nice if we could use metavariable regex, but this somehow causes matching to fail.


inkz (approved):
:+1:

mschwager:
> would be nice if we could use metavariable regex, but this somehow causes matching to fail.

Can you elaborate on this? I would expect `metavariable-regex` to work here. E.g. we could simplify the rules greatly if we did a `regex: (distinct|group_by|filter|...)` instead of multiple different rules.

Also, do we care if the `format` input came from a function argument? In other words, should we instead catch any string being passed to `distinct|filter|etc`, not just `format` calls? I think any string being passed to these dangerous functions could lead to SQLi, not just arguments that are `format`ed. Thoughts?

colleend:
When I substituted order_by, group_by, etc with a metavariable, semgrep gives no matches. This is probably a matching bug, and I've filed a similar issue so I'll try to take some time this week and see if I can resolve the issue.

The same problem happens when substituting 'format' with a metavariable -- it leads to no matches. Would be nice if we could use deep matching in order to catch inputs that are passed into distinct, filter, etc, but if we do that we would match some safe calls (such as filter(cls.id == id, cls.info == info)) etc. I tried to do it that way, but having an and-not clause with the deep matching operator doesn't seem to be working. This is probably another matching bug 😓 

It might take a while to fix the metavariable/deep operator matching issues, so I'm thinking of pushing these rules now and later coming back to edit them.

mschwager:
I'm still unsure if I'm understanding correctly. Can you provide an example where Semgrep isn't catching what you'd expect? Also, what is the similar issue you created?

If we're having matching issues or deep expression deficiencies we should file bugs. Linking everything together and creating issues will help ensure we're tracking everything and that we cover all expected cases :+1:

colleend:
The issue I filed is: https://github.com/returntocorp/semgrep/issues/1776. 

So ideally, something like this would be nice: https://dev.semgrep.dev/s/z8Lz/, but the pattern-not isn't doing what's expected.

mschwager:
Ahh, interesting! Yeah, I think that's definitely a good use-case for the deep expression operator.

My only suggestion here would be: if we have any plans to combine all these rules into a single rule file in the future we may want to do so now. Otherwise combining into a single file in the future will change the full rule-id path, which breaks things like rulesets and `nosem`.

Other than that, this looks like a good first pass on sending strings to dangerous sqlalchemy functions :+1:

mschwager (approved):

colleend:
ohh I see! sounds good, I'll combine them into a single rule file. I also looked more into it, and apparently the reason why metavariable matching wasn't working for me was because I accidentally named 2 metavariables the same thing 😓 I'll refactor and then merge!

PR #900, opened 2020-10-06
md5 rules for ocaml

### LINKED ISSUE #897: Provide OCaml Rules
**Is your feature request related to a problem? Please describe.**
We do not currently have any OCaml rules in the registry, although we say that we support OCaml

**Describe the solution you'd like**
At least one rule in OCaml inside an ocaml folder (ocaml.correctness...)

**Describe alternatives you've considered**

**Additional context**
We want to be able to list OCaml here and have at least 1 result

Screen Shot 2020-10-05 at 8 25 47 PM

colleend:


ocaml/lang/security/no-md5.yaml
```diff
@@ -0,0 +1,16 @@
+rules:
+- id: no-md5
+  patterns:
+  - pattern-either:
+    - pattern: |
+        Hash.md5
+    - pattern: |
+        MAC.hmac_md5
```
mschwager, line 8: IIRC HMAC-MD5 has no known practical vulnerabilities yet:

- https://crypto.stackexchange.com/questions/9336/is-hmac-md5-considered-secure-for-authenticating-encrypted-data
- https://en.wikipedia.org/wiki/HMAC#Security

Best practices would still avoid HMAC-MD5, but it's not broken. So, depending on what we're going for here, we may want to rethink including this pattern.

colleend: good point, we should only catch md5 for now

minusworld:
Closing due to inactivity. Reopen if still relevant.

PR #903, opened 2020-10-08
Remove wrong semicolons

aryx:
With the recent changes to the Javascript pattern parser, semicolons
are now more meaningful, and a pattern like
'foo(...) {...};'  is parsed a a sequence of 2 statements: a for and
an empty statement.

test plan:
make test


aryx:
@ajinabraham I had to fix one of your rule, so you may need to modify it in your own repository of rules too.

inkz (approved):

ajinabraham:
Appriciate the heads up. What version of semgrep will have these changes? I couldn't find it from Actions logs on a quick skim.

aryx:
I guess the next one, 0.28

PR #905, opened 2020-10-09
Pointer fix

colleend:
For issue https://github.com/returntocorp/semgrep-app/issues/830


mschwager (approved):
Does this invalidate any rulesets that use this rule? Changing the file name changes the path-dependent rule-id along with the changed `id` here, which I believe our rulesets use.

colleend:
Good question, the rule isn't in any rulesets currently :)

PR #907, opened 2020-10-10
Fix javascript.lang.correctness.useless-assignment false positive.

dkasak:
A repeated assignment isn't useless if the value of the first assignment
is used as an input to a function call in the second assignment.

I didn't make it as general as possible since the other patterns also aren't.
The deep expression operator could be used to also excluded false positives
like:

    z = y;
    z = 1 + $FUNC(z);


minusworld (approved):
Nice! And especially thanks for test cases! 🤘

PR #909, opened 2020-10-13
Adding the first ocaml rules!

aryx:
This used to be in pfff/semgrep.yml but those rules are general enough
to be put in semgrep-rules/.

test plan:
make test

### DIFF
--- a/ocaml/lang/best-practice/list.ml
+++ b/ocaml/lang/best-practice/list.ml
@@ -1,0 +1,13 @@
+let test1 xs =
+  (* ruleid:list-find-outside-try *)
+  if List.find 1 xs
+  then 1
+  else 2
+
+let test2 xs =
+ (* ok *)
+ try 
+   if List.find 1 xs
+   then 1
+   else 2
+ with Not_found -> 3
--- a/ocaml/lang/correctness/physical_vs_structural.ml
+++ b/ocaml/lang/correctness/physical_vs_structural.ml
@@ -3,3 +3,9 @@ let test xs ys =
   if xs != ys
   then 1
   else 2
+
+let test2 xs ys = 
+  (* ruleid:physical-equal *)
+  if xs == ys
+  then 1
+  else 2
--- a/ocaml/lang/correctness/physical_vs_structural.ml
+++ b/ocaml/lang/correctness/physical_vs_structural.ml
@@ -1,0 +1,5 @@
+let test xs ys = 
+  (* ruleid:physical-not-equal *)
+  if xs != ys
+  then 1
+  else 2
--- a/ocaml/lang/correctness/physical_vs_structural.yaml
+++ b/ocaml/lang/correctness/physical_vs_structural.yaml
@@ -1,6 +1,11 @@
 rules:
+- id: physical-equal
+  pattern: $X == $Y
+  message: You probably want the structural inequality operator =
+  languages: [ocaml]
+  severity: WARNING
 - id: physical-not-equal
   pattern: $X != $Y
-  message: You probably want the structural inequality operator <>.
+  message: You probably want the structural inequality operator <>
   languages: [ocaml]
   severity: WARNING
--- a/ocaml/lang/correctness/physical_vs_structural.yaml
+++ b/ocaml/lang/correctness/physical_vs_structural.yaml
@@ -1,0 +1,6 @@
+rules:
+- id: physical-not-equal
+  pattern: $X != $Y
+  message: You probably want the structural inequality operator <>.
+  languages: [ocaml]
+  severity: WARNING
--- a/ocaml/lang/correctness/useless_eq.yaml
+++ b/ocaml/lang/correctness/useless_eq.yaml
@@ -1,6 +1,6 @@
 rules:
 - id: useless-equal
   pattern: $X = $X
-  message: This is always true.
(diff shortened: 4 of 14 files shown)


aryx:
I hope I've followed the hidden conventions used for the other languages @minusworld and @mschwager.
Let me know if I missed something.

nbrahms (approved):
I'm surprised @mjambon hasn't convinced you to add a
```ocaml
| _ ->
```

rule yet

aryx:
He has not even started ... give him a few minutes :)

mjambon:
@nbrahms ocaml has a warning for `| _ ->`, which can be turned on or off by the build system. Do we need a semgrep rule then? More generally, is there a short guide to help us decide what should be reported and what shouldn't?

ocaml/lang/best-practice/list.ml
```diff
@@ -0,0 +1,13 @@
...
+  then 1
+  else 2
+
+let test2 xs =
+ (* ok *)
+ try 
+   if List.find 1 xs
+   then 1
+   else 2
+ with Not_found -> 3
```
mjambon, line 13: I would also recommend using `List.find_opt`.

Similarly, there's `Hashtbl.find`, which used to be widely used and could use a semgrep rule (if you haven't added one already). There must be a few others similar functions that raise `Not_found` in `List` or in other modules of the stdlib.

Note that Janestreet Base/Core_kernel/Core clobbers the interface of the standard library. Users of that library normally do `open Base` or `open Core_kernel` or `open Core`, all of which expose a [different `List.find`](https://ocaml.janestreet.com/ocaml-core/latest/doc/base/Base/List/index.html#val-find). I wouldn't worry about it.

ocaml/lang/correctness/physical_vs_structural.ml
```diff
@@ -0,0 +1,5 @@
+let test xs ys = 
+  (* ruleid:physical-not-equal *)
+  if xs != ys
+  then 1
+  else 2
```
mjambon, line 5: Obviously you want a check for `==` as well.

ocaml/lang/correctness/physical_vs_structural.yaml
```diff
@@ -0,0 +1,6 @@
+rules:
+- id: physical-not-equal
+  pattern: $X != $Y
+  message: You probably want the structural inequality operator <>.
```
mjambon: I would explain how to use physical equality without triggering a semgrep error/warning.

ocaml/lang/correctness/useless_eq.yaml
```diff
@@ -0,0 +1,6 @@
+rules:
+- id: useless-equal
+  pattern: $X = $X
+  message: This is always true.
+  languages: [ocaml]
+  severity: ERROR
```
mjambon, line 6: Is this something people do? Normally it's done to check that a float is not a NaN. OCaml has now a `Float.is_nan` to take care of it, so there's no longer a justification for tests like `x <> x`. If compatibility with old versions of ocaml is important, we can suggest they write their own `is_nan` function e.g.
```ocaml
let is_nan x =
  classify_float x = FP_nan
```

mjambon:
More on `| _ ->`: I don't think it's bad in general. It's usually legitimate when used to check for only one case and the type supports more than 2 cases:
```ocaml
type t =
| Foo of string
| Bar of int
| Baz of t
| Zoinks

(* likely acceptable *)
match x with
| Foo y -> ...
| _ -> ...

(* less likely acceptable *)
match x with
| Foo _ | Bar _ -> ...
| _ -> ...
```
I don't know if a semgrep rule can tolerate the first example but not the second one. I think it ~~would~~ could be nice.

ocaml/lang/correctness/useless_eq.yaml
aryx: good idea. Actually we have a similar rule for python that also suggest a special function to check for NaN.

And yes people do write equal expressions ... usually because they made a  typo. This rule found a surprisingly big number of bugs in open source projects ... you would be surprised.

nbrahms: @mjambon This is a common oops in webapps:

```python
user_id == user_id
```

when someone meant

```python
user.id == user_id
```

nbrahms:
> More on `| _ ->`: I don't think it's bad in general. It's usually legitimate when used to check for only one case and the type supports more than 2 cases:
> 
> ```ocaml
> type t =
> | Foo of string
> | Bar of int
> | Baz of t
> | Zoinks
> 
> (* likely acceptable *)
> match x with
> | Foo y -> ...
> | _ -> ...
> 
> (* less likely acceptable *)
> match x with
> | Foo _ | Bar _ -> ...
> | _ -> ...
> ```
> 
> I don't know if a semgrep rule can tolerate the first example but not the second one. I think it ~would~ could be nice.

I'm not sure how we've written the ellipsis operator, but I'd expect something like the following to work:

```yaml
patterns:
- pattern: |
    match $X with ... | _ -> ...
- pattern-not: |
    match $X with | $Y -> ... | _ -> ...
```

aryx:
| More generally, is there a short guide to help us decide what should be reported and what shouldn't?
@mjambon not that I know of.

PR #910, opened 2020-10-13
Add check for Flask class GET methods with CRUD side effects

chgg-kboberg:
Title says it all: add a best practice check for Flask to warn on Flask class GET methods that have state-changing side effects.


minusworld (approved):
This is a neat solution to that problem 👍 Thanks!

PR #911, opened 2020-10-13
Port most of the ocamllint rules to semgrep

aryx:
This is porting most of the rules at:
https://github.com/cryptosense/ocamllint/blob/3477d6f7313a1394ffb1176d04b7d33d3f6670e1/lib/ocamllint_rules.ml#L140

Most of the test cases were also ported:
https://github.com/cryptosense/ocamllint/blob/3477d6f7313a1394ffb1176d04b7d33d3f6670e1/test/tests.ml#L37

A few remaining rules can't be easily encoded in semgrep
(e.g., the constant match), or
are now detected by the compiler
(e.g., the incorrect use of partial application), or
don't seem very useful
(e.g., the application of list operations on singletons).

test plan:
make test

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -4,4 +4,5 @@ test:
 	semgrep --validate --config=$$PWD/javascript $$PWD
 	semgrep --validate --config=$$PWD/java $$PWD
 	semgrep --validate --config=$$PWD/go $$PWD
+	semgrep --validate --config=$$PWD/ocaml $$PWD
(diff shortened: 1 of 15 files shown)


aryx:
next tool to port: https://github.com/upenn-cis1xx/camelot

aryx:
Note that ocamllint does not work with ocaml > 4.04 ... so I was not even able to run it on our codebase.

DrewDennison (approved):

PR #914, opened 2020-10-14
GitHub action

colleend:
should upload the tar produced by semgrep/test.py as artifact, which will speed up tests for our v2 registry.

Test plan: check if tar file is uploaded in github.


ievans (approved):
Yay!

PR #922, opened 2020-10-15
Add examples

colleend:
add examples fix!


ievans (commented):
@colleend can we add in Nathan's negative test case for the `yield` example?

nbrahms (approved):

python/lang/correctness/return-in-init.py
```diff
@@ -36,3 +36,9 @@ def __init__():
     # OK, not in class
     yield 5
 
+class Ab:
+    def func1():
+        if not hello:
+            yield from blah
```
nbrahms, line 42: I'm surprised this parses... `from` is a keyword that usually only works in imports.

nbrahms, line 42:
```suggestion
            yield "blah"
```

python/lang/correctness/return-in-init.py
```diff
@@ -36,3 +36,9 @@ def __init__():
...
+        yield from other
```
nbrahms, line 43:
```suggestion
        yield other
```

PR #924, opened 2020-10-15
Move number comparisons from pattern-where-python to metavariable-comparison where possible

mschwager:
Depends on https://github.com/returntocorp/semgrep/pull/1824.

We'll obviously need `metavariable-comparison` to be released before we can move forward here. We may even want to delay this change until a bit after to give people time to upgrade before we break any builds with backwards-incompatible rules. I'll put this in draft state until then.

But... it works! The tests pass :tada:

### DIFF
--- a/go/lang/correctness/overflow/overflow.yaml
+++ b/go/lang/correctness/overflow/overflow.yaml
@@ -8,7 +8,10 @@ rules:
       $F, $ERR := strconv.Atoi($NUM)
       ...
       int16($F)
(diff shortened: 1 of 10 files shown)


aryx:
Woohoo!

aryx:
I wonder if we could keep the old pattern-where-python: but mark it in a special way to indicate it's deprecated
for an easier migration path for users. Thrift and other IDLs have this versioning mechanism where the code generator  will accept structure with new fields. Maybe we could now modify semgrep to make it easier to accept future versions without breaking and to also accept fields marked as deprecated or superseded by another one?

colleend (approved):

PR #929, opened 2020-10-19
Remove trailing whitespace

mschwager:


minusworld (approved):

ievans:
@mschwager this makes me so happy. Can we / are we already automating this as part of the pre-commit process?

mschwager:
> @mschwager this makes me so happy. Can we / are we already automating this as part of the pre-commit process?

https://github.com/returntocorp/semgrep-rules/pull/938!

PR #930, opened 2020-10-19
Update direct-use-of-jinja2 to detect `Template.render`

minusworld:
Based on https://github.com/returntocorp/semgrep-app/pull/864/files#diff-d675585d3e90d48ca868124fa4aca8628b068b5b463ddcd0995d68c4fc6dc401R24

### DIFF
--- a/python/flask/security/xss/audit/direct-use-of-jinja2.yaml
+++ b/python/flask/security/xss/audit/direct-use-of-jinja2.yaml
@@ -20,7 +20,6 @@ rules:
   - pattern: jinja2.Template.render(...)
   - patterns:
     - pattern-inside: |
-        @$APP.route(...)
-        def $FUNC(...):
-          ...
+        $TEMPLATE = $ENV.get_template(...)
+        ...
     - pattern: $TEMPLATE.render(...)
--- a/python/flask/security/xss/audit/direct-use-of-jinja2.yaml
+++ b/python/flask/security/xss/audit/direct-use-of-jinja2.yaml
@@ -15,6 +15,7 @@ rules:
   - python
   severity: WARNING
   pattern-either:
+  - pattern: import jinja2
   - pattern: jinja2.Environment(...)
   - pattern: jinja2.Template.render(...)
   - patterns:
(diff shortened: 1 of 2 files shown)


python/flask/security/xss/audit/direct-use-of-jinja2.yaml
```diff
@@ -15,7 +15,9 @@ rules:
   - python
   severity: WARNING
   pattern-either:
+  - pattern: import jinja2
```
brendongo: Do we want to fire on import?

minusworld: I dunno, I got scared that I'd miss some other case 😐; I think for Flask, the recommendation would be to always prefer `render_template` to using `jinja2` in any capacity. What do you think?

mschwager:
We probably want to ensure this templating is being performed in a Flask context. Jinja2 can be used outside of webserver contexts, so XSS and escaping may be of little concern in certain situations.

minusworld:
> We probably want to ensure this templating is being performed in a Flask context. Jinja2 can be used outside of webserver contexts, so XSS and escaping may be of little concern in certain situations.

Originally, I had some of it scoped to being inside a Flask route. However, if you check the example listed in the original comment, you can see that we missed a finding because it wasn't in a Flask route.

My thinking here is that we can assume that if someone is using the Flask rules, they are using Flask. This rule is also highlighted in the Flask-specific XSS pack. Because of that, I want to be careful not to miss things.

However, your point is valid: this rule might be better suited for a specific organization's policy than a community-wide rule. 
OTOH (again), we're moving in a direction where individual rules are not consumed so much as rulesets or policies, which gives us the ability to specify an "intent" for consuming the rule. In this case, the "intent" is: 1) I have Flask and 2) I want to prevent XSS.

Wasn't planning on writing all this but there ya go 🤷‍♂️ 😆

mschwager (approved):
I generally try to err on the side of under-reporting vs. over-reporting. As we've learned, people dislike being inundated with FPs more than FNs. But I don't have strong feelings - we can always try this out until someone says otherwise :+1:

PR #949, opened 2020-10-22
added file finally

colleend:


mschwager (approved):
It's back!

colleend:
never thought i'd see it again 😩

gastonstrizzolo:
Hi @colleend, thanks for adding this, do you know how to run this file? I don't fully understand if the Redos detector is integrated to Semgrep because if I run Dlint in my code it has some findings, but with Semgrep I don't know if I have to import something

colleend:
Hi, thanks for the question -- currently the dlint redos detector isn't fully integrated into semgrep. Currently we're not focusing on redos, so it will be a while before we are able to make progress with this.

PR #951, opened 2020-10-23
Added support for Golang format string issues from the pgx package

bertonjulian:
The current rules do not detect format string issues when using the [pgx PostgreSQL Driver and Toolkit](https://github.com/jackc/pgx) by [jackc](https://github.com/jackc) which appears to be quite a popular package and might be worth including in the main rule set (unless this is only focused on standard lib packages), but will leave that up to you all to decide :) 

The reason it is currently not picking up format string issues is due to the packages slight differences in function names / parameters.

I have tested this rule on a few repos locally and it passed all pre-commit hooks but was having trouble running `make test` as there were lots of errors outside of the files i was changing.


minusworld (approved):
Looks good! Thank you so much!

Issue #798, opened 2020-08-16
[Check] semgrep rules for browser javascript


paolodm:
There is a folder for javascript but it looks for security issues on the server level (aka node.js).

semgrep is an amazing tool and it would be really helpful to validate the security on the browser. This request is to add a folder for browser javascript rules.

Here some ideas for checks:

* `eval` calls
* dangerouslySetInnerHTML calls
* any `.innerHTML = ` calls
* `postMessage` without filtering for origin
* trustAsHtml

- [x] This ticket has links, references, or examples.
- [x] Your check has true positive and true negative test cases.
- [x] Your check has been labeled with true positive and false positive findings on at least 10 repositories, and the link is pasted in the ticket. If there are no findings, paste the triager link anyway.
- [x] Your check PR has been reviewed and merged.
Now, close the ticket!

minusworld:
@paolodm Thanks!! These ideas are great!

inkz:
- [x] `eval` calls

- [x] `dangerouslySetInnerHTML` calls

- [x] any `.innerHTML =` calls

- [x] `postMessage` without filtering for origin

- [x] `trustAsHtml`

inkz:
@paolodm thanks for check ideas, we implemented them!
most of the rules can be found in this pack https://semgrep.dev/p/clientside-js

and React specific rules will reside in a separate pack,
for now it is available here https://github.com/returntocorp/semgrep-rules/blob/develop/typescript/react/security/audit/react-dangerouslysetinnerhtml.yaml

PR #952, opened 2020-10-23
add insecure-innerhtml rule

inkz:
Add a rule for `.innerHTML` calls https://github.com/returntocorp/semgrep-rules/issues/798


colleend (approved):
🙏

PR #955, opened 2020-10-26
Ruby requests

colleend:
- [x] http requests rules pushed
- [x] ftp requests rules pushed
- [x] telnet request rules pushed
- [x] TLS certificate verification rule [here](https://github.com/returntocorp/semgrep-rules/blob/develop/ruby/lang/security/no-send.yaml)
- [x] Make sure to run TLS 1.2 or 1.3 only. Ruby has ensured this since [2018](https://blog.rubygems.org/2018/02/24/tls-10-and-11-deprecation-notice.html)
- [x] rules for openuri 
- [x] rules for httparty and restclient


inkz (approved):
:+1:

PR #973, opened 2020-10-30
Fix python string-concat-in-list f-string compatibility and improve robustness

TeoZosa:
Just started using Semgrep after reading the r2c HN article yesterday, really good stuff! After throwing the kitchen sink at one of my Python projects, I found two small bugs in the [`string-concat-in-list`](https://github.com/returntocorp/semgrep-rules/blob/develop/python/lang/correctness/common-mistakes/string-concat-in-list.yaml) rule which caused it to:

1. incorrectly flag single f-strings as pairs of implicitly concatenated strings.
- Note: **this may actually be a bug with the Semgrep parsing itself.**
2. miss implicitly concatenated strings when they don't occur at the beginning of the list. 

This PR fixes both issues, albeit with a change to the match reporting output granularity (see the [b9d2a5b commit message](https://github.com/returntocorp/semgrep-rules/commit/b9d2a5bc6d9e5a20128e02218dc2d33eed45fc57) 
for a before/after output comparison for [`string-concat-in-list.py`](https://github.com/returntocorp/semgrep-rules/blob/develop/python/lang/correctness/common-mistakes/string-concat-in-list..py)).

### DIFF
--- a/python/lang/correctness/common-mistakes/string-concat-in-list.yaml
+++ b/python/lang/correctness/common-mistakes/string-concat-in-list.yaml
@@ -2,7 +2,7 @@ rules:
 - id: string-concat-in-list
   patterns:
   - pattern: |
-      ["..." "...", ...]
+      [..., "..." "...", ...]
   message: |
     Detected strings implicitly concatenated inside a list.
     Python will implicitly concatenate strings when not explicitly delimited.
--- a/python/lang/correctness/common-mistakes/string-concat-in-list.yaml
+++ b/python/lang/correctness/common-mistakes/string-concat-in-list.yaml
@@ -6,7 +6,7 @@ rules:
   message: |
     Detected strings implicitly concatenated inside a list.
     Python will implicitly concatenate strings when not explicitly delimited.
-    Was this supposed to be individual elements of th list?
+    Was this supposed to be individual elements of the list?
   severity: WARNING
   languages:
   - python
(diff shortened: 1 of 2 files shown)


minusworld (approved):
Nice! 👍 Thank you for these!

minusworld:
Looks like the `ruleid` comments might be a few lines off. I'll try to patch it up on Monday if you don't happen to get to them before me. Thanks again! 🤘

python/lang/correctness/common-mistakes/string-concat-in-list.yaml
```diff
@@ -1,12 +1,16 @@
 rules:
 - id: string-concat-in-list
   patterns:
+  - pattern-inside: |
+      [...]
   - pattern: |
-      ["..." "...", ...]
```
jonatanmurga, line 5: Hi

python/lang/correctness/common-mistakes/string-concat-in-list.yaml
```diff
@@ -1,12 +1,16 @@
...
+      "..." "..."
+  - pattern-not-inside: |
+      f"..."
   message: |
     Detected strings implicitly concatenated inside a list.
     Python will implicitly concatenate strings when not explicitly delimited.
-    Was this supposed to be individual elements of th list?
```
jonatanmurga, line 9: 🤒

TeoZosa:
> Looks like the `ruleid` comments might be a few lines off. I'll try to patch it up on Monday if you don't happen to get to them before me. Thanks again! 🤘

Done and done 💪 ! Hope you have a great weekend!

jonatanmurga (commented):
😃

PR #982, opened 2020-11-09
Add spacegrep rules for Hugo TOML config files.

minusworld:
I work on another project that uses Hugo, the static site generator.
I kept getting bitten by this issue when developing when I would forget to change the baseURL back to its correct URL. I came up with a few Semgrep/spacegrep rules while working on it over the weekend to help me do this. 💪  Contributing them in case they're useful elsewhere.


inkz (approved):
Awesome!

PR #983, opened 2020-11-10
add react-dangerouslysetinnerhtml rule

inkz:
first Typescript rule!
covering https://reactjs.org/docs/dom-elements.html#dangerouslysetinnerhtml

### DIFF
--- a/typescript/react/security/audit/react-dangerouslysetinnerhtml.tsx
+++ b/typescript/react/security/audit/react-dangerouslysetinnerhtml.tsx
@@ -1,0 +1,30 @@
+function createMarkup() {
+  return {__html: 'Первый &middot; Второй'};
+}
+
+function TestComponent1() {
+	// ruleid:react-dangerouslysetinnerhtml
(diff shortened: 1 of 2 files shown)


minusworld (approved):

dlukeomalley:
Are there any recommendations on what to do instead of using `dangerouslySetInnerHtml` or references to good articles explaining it? My sense is that it's a common mistake for more junior developers and some extra coaching and context might be helpful.

Thank you @inkz!

inkz:
@dlukeomalley https://snyk.io/blog/10-react-security-best-practices/
most recommendations are "do not forget to sanitize the input"

emilva:
Developers tend to use `dangerouslySetInnerHtml ` when they "need" to dangerously set inner HTML🎉  , and the best practice for that is to santize the html by using DOMPurify by default. Maybe one could argue that `dangerouslySetInnerHtml` should never be used without a call to `DOMPurify`, but I guess there are use-cases where developers need to output unsafe html tags like `<script>`, and in those cases any injected data must be encoded for its context manually.

PR #984, opened 2020-11-10
change pre-commit python version to python3

inkz:
change `pre-commit` to `python3` instead of `python3.7` so it worked on distros with different versions of Python


mschwager (approved):
This was bugging you too, eh? ;)

PR #992, opened 2020-11-12
add typescript/react rules

inkz:


typescript/react/best-practice/react-props-spreading.yaml
```diff
@@ -0,0 +1,11 @@
+rules:
+- id: react-props-spreading
+  pattern-either:
```
mschwager: Nit: we don't need the `pattern-either`.

mschwager (approved):

typescript/react/best-practice/react-props-spreading.yaml
inkz: thanks! was double checking for stuff like that and still missed this one :smile:

inkz:
merged into https://github.com/returntocorp/semgrep-rules/pull/999

PR #993, opened 2020-11-12
Remove extra semicolons in some rules, because they now matter

aryx:
Because of https://github.com/returntocorp/semgrep/pull/2028
which is needed for https://github.com/returntocorp/semgrep/issues/1948
we do not remove anymore semicolons in expression statement patterns
(e.g., 'foo(...);')
As a consequence, if you have some semicolons in a pattern, because
we do not remove them, they have an influence on the matching process
(the range of the matched code will be 1 longer, and this may have
bad effect when combining pattern not using semicolon with pattern-not
patterns using semicolons).

Test plan:
use latest semgrep via pipenv shell and
$ make test


aryx:
@ajinabraham this might have an impact in some of your rules also (even though I didn't see any regression
in our contrib/nodejsscan directory).
TL;DR: you should not use semicolons in patterns if what you want to match is an expression, e.g., use foo(...) not foo(...);

ajinabraham:
Thanks for the heads up. Will watchout for the changes in 0.32.0

javascript/jquery/security/audit/jquery-insecure-selector.yaml
```diff
@@ -49,7 +49,7 @@ rules:
             }
       - pattern-either:
         - pattern: |
-            $(<... $Y ...>);
+            $(<... $Y ...>)
         - pattern: |
             $("..." + (<... $Y ...>));
```
mschwager: Should we remove this semi-colon too then?

javascript/jquery/security/audit/jquery-insecure-selector.yaml
```diff
@@ -49,7 +49,7 @@ rules:
...
         - pattern: |
```
mschwager, line 55: Same here.

mschwager (approved):
LGTM - we may be able to remove 2 more

javascript/jquery/security/audit/jquery-insecure-selector.yaml
aryx: yep, I will remove them too

ajinabraham:
@aryx Is there a regression in semgrep 0.35 regarding semicolons?
I made the changes to rules as suggested (removed semicolons) and everything worked till 0.34
and broken in 0.35

https://semgrep.dev/s/ajinabraham:rule_semgrep?version=0.34.0
https://semgrep.dev/s/ajinabraham:rule_semgrep?version=0.35.0

PR #996, opened 2020-11-12
Update nginx insecure-redirect rule with accurate meessage.

minusworld:
Include 'permanent' case too.


minusworld:
Comment from conversation with originator:

> So if I understand correctly the issue was:
> SSL was terminated at ALB, which then forwards a plain old HTTP request to nginx, which then further redirects using plain old HTTP

inkz (approved):
interesting!

PR #1001, opened 2020-11-13
Add rules for Angular DomSanitizer methods

ekse:
This rule flags calls to bypassSecurityTrust*() methods as well as DomSanitizer.sanitize() with SecurityContext.NONE, which does not sanitize content.

https://angular.io/api/platform-browser/DomSanitizer

To be on the safer side we could check that $X is of type DomSanitizer but the method names are probably distinct enough.


minusworld (approved):
Thanks! These look really great. :D I agree, I think that the names are distinct enough. If the future proves us wrong, though, we can update it.

PR #1004, opened 2020-11-17
Add space between keyword argument in Ruby

aryx:
This was causing regression when updating to the latest tree-sitter-ruby
https://github.com/returntocorp/semgrep/pull/2058

Without the space, this is parsed as a nested call
to allow_destroy with the :false atom as an argument.

test plan:
install latest semgrep-core
make test


inkz (approved):
:+1:

PR #1009, opened 2020-11-18
Fix semgrep-rules now that semgrep-core better match imported entities

aryx:
This is related to https://github.com/returntocorp/semgrep/pull/2076

test plan:
use latest semgrep-core
make test


aryx:
@salecharohit you may need to update some of your rules. There was a problem in semgrep-core which prevented
it to match imported entities (e.g., if there was import com.foo.bar.SaxFactory).

minusworld (approved):

PR #1013, opened 2020-11-19
Update xss regex rules with 'generic'

minusworld:
Updates some XSS regex-only rules to use the 'generic' matcher instead. This expands some of these rules to account for whitespace/newlines and reduces FP rate.

### DIFF
--- a/go/lang/security/audit/xss/no-interpolation-js-template-string.yaml
+++ b/go/lang/security/audit/xss/no-interpolation-js-template-string.yaml
@@ -1,6 +1,6 @@
 rules:
 - id: no-interpolation-js-template-string
-  message: |
+  message: >-
     Detected template variable interpolation in a JavaScript
     template string. This is potentially vulnerable to
     cross-site scripting (XSS) attacks because a malicious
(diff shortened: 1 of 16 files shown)


go/lang/security/audit/xss/no-interpolation-js-template-string.yaml
```diff
@@ -15,8 +15,8 @@ rules:
     - https://github.com/golang/go/issues/9200#issuecomment-66100328
     - https://blogtitle.github.io/robn-go-security-pearls-cross-site-scripting-xss/
   languages:
-  - none
+  - generic
   severity: WARNING
   patterns:
-  - pattern-regex: |-
-      `.*{{.*}}.*`
+  - pattern-inside: <script ...> ... ... ... ... ... </script>
```
mschwager, line 21: Why so many `...` here? Isn't one enough?

mschwager (approved):

go/lang/security/audit/xss/no-interpolation-js-template-string.yaml
minusworld: spacegrep heuristically only spans 10 lines per ellipsis. So, this spans 50! 

image

minusworld: I imagine script tags to be generally more than 10 lines, so I just guessed at 50. 🤷‍♂️

Issue #1011, opened 2020-11-19
Add 5 nginx rules to registry


minusworld:
Related to https://github.com/returntocorp/enterprise/issues/481.

PR #1014, opened 2020-11-19
Add nginx 7 security rules

minusworld:
- SSRF via proxy_pass
- scheme injection
- insecure ssl version
- missing ssl version (which defaults to insecure versions)
- prefer $host to $request_host to guard against SSRF
- header redefinition
- header injection


minusworld:
Closes https://github.com/returntocorp/semgrep-rules/issues/1011.

inkz (approved):
generic/nginx/security/header-injection.yaml
and
generic/nginx/security/header-redefiniton.yaml

are empty

PR #1019, opened 2020-11-24
fix react-controlled-component-password naming

inkz:
typo in filename


inkz:
in the pipeline there is smth wrong with the rule not related to commit:
```
 /github/workspace/python/flask/security/xss/audit/template-unquoted-attribute-var.html
```

colleend:
yep, I think the CI checks are failing in all our PRs right now 😓

colleend (approved):

PR #1021, opened 2020-11-24
Upload to s3

colleend:
This pull request is a draft for https://github.com/returntocorp/semgrep-app/issues/1137.

In order to upload the `semgrep_runs_output.tar.gz` file into a S3 bucket, we would have to:
1. Create an S3 bucket for this file
2. Add in necessary environment variables in Github secrets (4)
3. Check if everything runs successfully

If this works, then we can stop uploading the `semgrep_runs_output.tar.gz` file as a github artifact and instead upload it directly to a s3 bucket and get the s3 bucket's static URL.


.github/workflows/semgrep-rules-test-develop.yml
```diff
@@ -25,3 +25,11 @@ jobs:
         name: semgrep_runs
         path: semgrep_runs_output.tar.gz
         if-no-files-found: error
+    - uses: zdurham/s3-upload-github-action@master
```
ievans, line 28: Seems a bit risky to rely on a random github action that we're passing our secrets into. Would it be easy enough to just install awscli on a base image that has pip, or maybe we could pin this to a commit hash?

colleend: Looked into this more, there's also:

https://github.com/eine/tip

where you can use a single pre-release and continuously update it to provide a consistent URL for the file 🤔 
but essentially we do run into the problem of this requiring the github token secret being passed in.

But then we wouldn't need to generate our own image with pip & awscli. What do you think?

ievans: sorry @colleend I wasn't clear about what I meant by pin -- I was thinking pin `zdurham/s3-upload-github-action` to a commit hash, rather than being `@master`, so that we can just audit the code at that commit hash and feel comfortable with it. I think my preferred solution would just be to grab a base image and install AWS CLI, but defer to you.

mschwager: +1 here. We should prefer not to pipe `AWS_SECRET_ACCESS_KEY` to random 3 star Github Action repos.

colleend: Ah yes, haven't looked at this in a bit -- discussed with Nathan a little, and we were talking about just manually uploading the new semgrep_runs json to r2c.dev until registry 2.0 updates. 

We should probably discuss this a little bit more -- will the registry 2.0 updates get rid of semgrep-runs?

minusworld:
@colleend @mschwager Is this still relevant, or shall we close it?

mschwager:
I think we can close. IIRC this is done manually on new Semgrep releases. The best course of action would probably be for Semgrep App to run this job itself, and avoid all the indirection, since this is a simple Semgrep CLI test build artifact.

PR #1028, opened 2020-12-02
update react-missing-noreferrer rule

inkz:
updating `react-missing-noreferrer` to work only for external links


minusworld (approved):
nice

PR #1029, opened 2020-12-02
* contrib/nodejsscan/xss_node.yaml: Fix use of metavar ellipsis

aryx:
With https://github.com/returntocorp/semgrep/pull/2151,
the use of <...$XXX...> will generate a parse error because
$X... is now parsed as a varargs metavariable (or metavariable ellipsis).

We need extra space between the metavar and the ... if we don't want
to use a metavar ellipsis

test plan:
CI

### DIFF
--- a/contrib/nodejsscan/xss_node.yaml
+++ b/contrib/nodejsscan/xss_node.yaml
@@ -79,25 +79,25 @@ rules:
     - pattern: |
         $LOCALVAR = <... $REQ.$QUERY ...>;
         ...
-        $ARR.push(<...$LOCALVAR...>);
+        $ARR.push(<... $LOCALVAR ...>);
         ...
         $RES.write(..., <... $ARR ...>, ...);
     - pattern: |
         $LOCALVAR = <... $REQ.$QUERY.$FOO ...>;
         ...
-        $ARR.push(<...$LOCALVAR...>);
+        $ARR.push(<... $LOCALVAR ...>);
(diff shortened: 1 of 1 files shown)


aryx:
cc @ajinabraham that you now need to take care when you use $XXX..., you need space after the metavariable
because $XXX... will soon be parsed as a special kind of metavariable.

ajinabraham:
Quick question. Isn't this a syntax case that can easily invite mistakes? Like if having and not having space has different meanings.

aryx:
We could use $...ARGS instead of $ARGS...
cc @mschwager @nbrahms @minusworld

minusworld:
Is there some possibility of something like `$...(ARGS)`?
More random ideas:
- `$...{ARGS}`
- `$...:ARGS`
- `$:ARGS...`
- `...$:ARGS...`

aryx:
Note that this PR does not commit to anything regarding the syntax of metavar ellipsis. It just adds space
in a few rules; there are actually other rules in this file where they use extra space, like <... $ARR ...>.

aryx:
and @ajinabraham space has a meaning sometimes:  i++ 1 and i+ +1 have different meaning. Space between tokens is sometimes necessary to disambiguate constructs.

aryx:
But yes maybe $...ARGS is a safer choice.

brendongo (approved):

PR #1033, opened 2020-12-03
More metavar ellipsis fixes by adding extra space

aryx:
test plan:
git grep -e '[A-Z]\.\.\.' does not show anymore matches


aryx:
Another fix, cc @ajinabraham 
Again, this does not mean we will use the $X... syntax, but it does not hurt to add extra space. In many case this
improves consistency because most of the rules use extra space around those metavariables.

inkz (approved):
:+1:

PR #1037, opened 2020-12-04
Add in hadolint 400X rules

minusworld:


generic/dockerfile/best-practice/maintainer-is-deprecated.yaml
```diff
@@ -0,0 +1,11 @@
+rules:
+- id: maintainer-is-deprecated
+  severity: INFO
+  languages: [generic]
+  pattern: MAINTAINER
```
mschwager, line 5: If possible, it might be nice to make this more pointed to avoid false positives. Here it seems like `MAINTAINER` could exist anywhere in the Dockerfile, not necessarily as a command. One option may be `MAINTAINER ...`, but I suspect that has the same issue. Maybe you have better ideas here...

mschwager (approved):
:tada:

generic/dockerfile/best-practice/maintainer-is-deprecated.yaml
minusworld: I thought that MAINTAINER (in all caps) was the surest way to grab it; once these are all merged I'm going to do a run on MAP for posterity.

The only better option I can think of is doing this as a regex rule and requiring it to be at the beginning of a line. 🤷‍♂️

PR #1039, opened 2020-12-09
Clean up ~20 rules

minusworld:
Updates messages with clear next steps.
Adds autofixes to some rules.
Makes rule match text more concise for some rules.


inkz (approved):
Amazing work!

PR #1040, opened 2020-12-09
Adding a couple PHP security rules

rezen:
- Added a rule to check for loose equality checks involving md5
- Added a rule to check for CURLOPT_SSL_VERIFYPEER set to false


ievans:
thanks for these rules @rezen! I got your CLA, @cla-bot check

php/lang/security/md5-loose-equality.yaml
```diff
@@ -0,0 +1,18 @@
...
+  patterns:
+  - pattern-either:
+    - pattern: $X == $FUNC(...)
+    - pattern: $FUNC(...) == $X
+    - pattern: $FUNC(...) == $FUNC(...)
+  - metavariable-regex:
+      metavariable: $FUNC
+      regex: md5|md5_file
+  message: |
+    Make sure comparisons involving md5 values are strict to avoid type juggling issues
```
ievans: Does "strict" mean using the `===` operator? If so, perhaps the message could be made more actionable

ievans (approved):
Looks great to me!

php/lang/security/md5-loose-equality.yaml
rezen: It does 🙂 would you like me to update the message?

minusworld (approved):

PR #1041, opened 2020-12-10
Limit Dockerfile rules to Dockerfiles only

minusworld:
And filter out some FPs for correctness rules.
FPs based on https://dev.massive.ret2.co/triager/filter/1279.


inkz (approved):
:+1:

PR #1043, opened 2020-12-14
Unsafe deserialization with RestEasy

TmmmmmR:
A quick background about this rule and the vulnerable code :

- It’s a simple REST Web service using Resteasy library for JAX-RS implementation, and it accepts as input objects of the class Pair ;
- The @Consumes annotation is used to specify which MIME media types a resource can accept from a client (the value of Content-type header). In this case, the JAX-RS do contact method is using the value  @Consumes({"*/*"}) so we can provide any type of object/data as input ;
- RESTEasy uses what we call “providers” to marshal request body into the parameter of JAX-RS method. As an example, the resteasy-yaml-provider is used for mapping/extracting Yaml object from the body of an HTTP request
- In our example, and since we didn’t specify the type of accepted object for this JAX-RS method, Resteasy will try to detect the most suitable provider to extract data from an HTTP request, and this will be based on the value of the Content-type header ;
- If we look at the list of supported providers in Resteasy, we will find an interesting one "org.jboss.resteasy.plugins.providers.SerializableProvider" which allow the unmarshaling of serialized java object. So, as you may guess, if there are “interesting” Java classes in the CLASSPATH, we can abuse this JAX-RS method to deseriliaze a java object of our choice, which may leads to (wait for it) remote code execution 😊

So this rule aims to detect the following annotations:

Explicitly accepting Java serialize object (potential RCE..)
- ```@Consumes({"application/x-java-serialized-object"})```

Or the implicit:

- No @Consumes annotation
- ```@Consumes({"*/*"})```
- ```@Consumes({MediaType.WILDCARD_TYPE})```

**Important note** : This issue has already been fixed in recent Resteasy versions by disabling dangerous providers by default (https://issues.redhat.com/browse/RESTEASY-1461), but it does not stop “creative” people from enabling them (see : http://stackoverflow.com/questions/15074431/how-do-i-make-jersey-use-java-serialization) since it’s only marked as deprecated (completely removed on last August : https://github.com/resteasy/Resteasy/pull/2489/commits/d0d6e7809e5987fcf1a10bbc87def94e82b91795), and in some case all providers are loaded at run time when using resteasy-servlet-initializer (https://docs.jboss.org/resteasy/docs/3.0.17.Final/userguide/html_single/index.html#d4e113). So, using the recent Resteasy won’t make your code safe against this issue, and this why it’s fundamental to use semgrep to detect this kind of vulnerable code patterns 😊 

More info : https://0ang3el.blogspot.ca/2016/06/note-about-security-of-resteasy-services.html
PS : this one is on the todo-list of findsecbug => https://github.com/find-sec-bugs/find-sec-bugs/issues/198


ievans:
@TmmmmmR this is a super cool PR -- sorry it's taken so long for us to get it merged. Do you have cycles to fix the test failures and sign the CLA (it's a github one-click)?  

if not perhaps @minusworld can to take a look, though we'd still need the CLA

TmmmmmR:
Thank you @ievans ! I would love to see this included in the community rules :)
Unfortunantly I can't get more details about the errors related to my PR. Maybe @minusworld can help me with this one ;)

minusworld:
I'll take a look 👍

minusworld (approved):

minusworld:
Woohoo! 🙌  Got it. Turns out the test comments were just on the wrong lines. Thanks for the contribution! Thanks also for your patience while we got this through. 👍

PR #1044, opened 2020-12-15
fixing a file name typo and improving the last-user-is-root rule base…

daghan:
Based on user feedback that the `last-user-is-root` rule wouldn't catch the following pattern:

```
FROM alpine
USER root
RUN apk install curl
USER root
CMD ["/hello"]
```


minusworld (approved):
Ahh, nice! That's really clever!

PR #1051, opened 2020-12-22
python/lang/correctness: add exception for tempfile-without-flush

### LINKED ISSUE #1000: python tempfile-without-flush: Don't require .write() for exception?
**Describe the bug**
The `python.lang.correctness.tempfile.flush.tempfile-without-flush` rule has [exceptions](https://github.com/returntocorp/semgrep-rules/blob/35e9f350a5d0190c936502fff179ba2c465c4438/python/lang/correctness/tempfile/flush.yaml#L5-L20) if `.write()` and then `.close()` or `.flush()` is used on a tempfile. However, this doesn't cover code which calls `.close()` immediately, without writing:

```
$ semgrep --config=p/r2c-ci --verbose test.py
[...]
severity:error rule:python.lang.correctness.tempfile.flush.tempfile-without-flush: possibly missing a .flush() or .close() call to temporary file handle; file may or may not exist when handle.name is used
3:handle = tempfile.NamedTemporaryFile(delete=False)
4:handle.close()
5:print(handle.name)
```

**To Reproduce**
```python3
import tempfile

handle = tempfile.NamedTemporaryFile(delete=False)
handle.close()
print(handle.name)
```

**Expected behavior**
It seems to me like the exception should trigger whenever `.close()` is invoked, no matter if `.write()` is called before that or not?

redshiftzero:
Closes #1000


ievans:
thank you for the contribution @redshiftzero! @cla-bot check pls

ievans (approved):
cc @mschwager

PR #1054, opened 2020-12-27
exit: Fix a false positive

iamleot:
`exit()` can be used as-is when imported via `from sys import exit`.


ievans (approved):
@aryx this one changed due to your relaxation of FQNs recently. Thanks @iamleot!

iamleot:
Thank you @ievans!

PR #1064, opened 2021-01-11
Rename security_electronjs.yaml to security_electron.yaml

chmccreery:
Code snippet is missing from editor because `security_electronjs` and `security_electron` are different strings


ievans:
@chmccreery should we ask @ajinabraham to do this upstream? want to make the PR against his repo?

chmccreery:
I see that he saw this- so closing

PR #1065, opened 2021-01-11
Patch eqeq check for Go

minusworld:
Patch eqeq check for Go which was failing on tests now with the new contsant propagation. This should maintain the rule but allow tests to pass.


IagoAbal (approved):
Great! Thanks for the prompt fix!

PR #1069, opened 2021-01-13
Adapt test files for the new constant propagation

aryx:
With https://github.com/returntocorp/semgrep/pull/2386, we now better
propagate constant using a dataflow-based analysis.

Before, one had to manually encode some constant propagation via extra
patterns, for instance:
 - patterns:
   - pattern-either: foo("constant")
   - pattern-either:
     $X = "constant"
     ...
     foo($X)

and the test file usually contained test code like:
  def test1():
   #ruleid: blah
   x = "constant"
   foo(x)

but now we actually propagate the constant x, so the first
pattern also match the code so I've updated the test file to:
  def test1():
   #ruleid: blah
   x = "constant"
   #ruleid: blah
   foo(x)

Note that the right fix is actually to simplfy the yaml file and removing
those manually-encoded-constant-propagation patterns that are now automatically
handled by the engine.

test plan:
compile semgrep-core with latest develop, make install
make test in semgrep-rules

### DIFF
--- a/java/lang/correctness/hardcoded-conditional.java
+++ b/java/lang/correctness/hardcoded-conditional.java
@@ -1,11 +1,13 @@
 class Bar {
-    void main() {
+    void main(boolean arg) {
         boolean myBoolean;
 
         // ruleid:hardcoded-conditional
         if (myBoolean = true) {
             continue;
         }
+        // note that with new constant propagation, myBoolean is assumed
+        // to true below
 
         // ruleid:hardcoded-conditional
         if (true) {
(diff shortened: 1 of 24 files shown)


aryx:
cc @IagoAbal

aryx:
cc @ajinabraham since I've modified a few nodejsscan rules
@ajinabraham you will be able to probably simplify many of your rules in 0.38 with the new constant propagation algorithm inside semgrep.

java/lang/correctness/hardcoded-conditional.java
```diff
@@ -1,11 +1,13 @@
 class Bar {
-    void main() {
```
underyx, line 2: did you intend to use this arg somewhere? or is it actually used and I just can't read java?

underyx: yep, I see where! never mind

underyx (approved):

PR #1070, opened 2021-01-18
raw-html-concat: Detect `.concat()` in addition to `+`

### LINKED ISSUE #2409: JS browser security concat bypass not detected
**Describe the bug**
Context: This is a throw away account to preserve my privacy. I am a dev at a Fortune 500 and my manager is an a@@ h@@@ who doesn't understand or cares about security. Recently one of our clients needed a security audit for the coding practices and we used tools like semgrep, SonarQube etc. My manager had forced us to use jquery with string concat to create html elements instead of dynamic html components creation. We all protested this but he didn't agree citing timelines. So during the security audit Semgrep found "raw-html-concat" issues and I was happy that atleast now he will be forced to follow proper standards. But instead he bypassed me and forced the juniors to use concat() instead of string concat. And this was undetected by semgrep. 

For example, in your example if he was incharge of fixing he would have converted 

`return '<div align="' + (attrs.defaultattr || 'left') + '">' + content + '</div>';`

to

`return '<div align="'.concat(attrs.defaultattr || 'left').concat('">'.concat(content)).concat('</div>');`

The changes my juniors has done to the code base is pathetic but it passes all tests and goes unnoticed. Common sense dictates that nobody in their right mind would do something like this to beautiful code, but I'm letting you know that these kind of people does exist. Thank you for your efforts. Please add additional patterns to capture this.

**To Reproduce**
Convert string concat with + to concat()

**Expected behavior**
Even if user tries to trick semgrep it should be noticed

underyx:
Closes https://github.com/returntocorp/semgrep/issues/2409

### DIFF
--- a/javascript/browser/security/raw-html-concat.yaml
+++ b/javascript/browser/security/raw-html-concat.yaml
@@ -25,6 +25,18 @@ rules:
         "..." + (<... document ...>);
     - pattern: |
         (<... document ...>) + "...";
+    - pattern: |
+        "...".concat(..., <... window ...>, ...);
+    - pattern: |
+        (<... window ...>).concat(..., "...", ...);
+    - pattern: |
+        "...".concat(..., <... location ...>, ...);
+    - pattern: |
+        (<... location ...>).concat(..., "...", ...);
+    - pattern: |
+        "...".concat(..., <... document ...>, ...);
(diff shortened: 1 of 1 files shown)


inkz (approved):
amazing!

Issue #1067, opened 2021-01-13
[Rule] Potential XSS via translate filters in Django Templates


minusworld:
**Rule Description**
Find instances of potential XSS introduced by `trans` in Django. See https://edx.readthedocs.io/projects/edx-developer-guide/en/latest/preventing_xss/preventing_xss_in_django_templates.html#html-escaping-translations-in-django-templates.

**Examples or references**

cf. https://edx.readthedocs.io/projects/edx-developer-guide/en/latest/preventing_xss/preventing_xss_in_django_templates.html#html-escaping-translations-in-django-templates

**Additional information**

I do wonder if this only applies to older versions? Be sure to check that.

**PR Checklist**
- [x] This ticket has links, references, or examples.
- [x] The rule has **true positive** and **true negative** test cases in a file that matches the rule name.

> If the rule is `my-rule`, the test file name should be `my-rule.js`.
>
> True positives are marked by comments with `ruleid: <my-rule>` and true negatives are marked by comments with `ok: <my-rule>`.

- [x] The rule has a good message. A good message includes:

> 1. A description of the pattern (e.g., missing parameter, dangerous flag, out-of-order function calls).
> 1. A description of why this pattern was detected (e.g., logic bug, introduces a security vulnerability, bad practice).
> 1. An alternative that resolves the issue (e.g., use another function, validate data first, discard the dangerous flag).

- [x] After the PR has been reviewed and approved, merge the rule and close this ticket! Thanks for contributing!

inkz:
`I do wonder if this only applies to older versions? Be sure to check that.`
Warning about no escaping in templates first appears in the docs of 2.2 version (which usually means that older versions has the same issue)
https://docs.djangoproject.com/en/2.2/topics/i18n/translation/#internationalization-in-template-code

PR #1071, opened 2021-01-19
XSS via translate filters in Django Templates

inkz:
Pull for: https://github.com/returntocorp/semgrep-rules/issues/1067


minusworld (approved):
looks great! 👍

PR #1075, opened 2021-01-22
react-css-injection: Fix FP on any user variable usage

underyx:
Wrongly alerted here https://github.com/returntocorp/semgrep-app/pull/1454#discussion_r562722720


minusworld:
👍 

@inkz Can you double-check this? I'm going to approve for now so we can keep going

minusworld (approved):

PR #1099, opened 2021-02-11
Fix todos in Golang math rand rule

mschwager:


go/lang/security/audit/crypto/math_random.yaml
```diff
@@ -5,16 +5,31 @@ rules:
...
     - https://cheatsheetseries.owasp.org/cheatsheets/Cryptographic_Storage_Cheat_Sheet.html#secure-random-number-generation
-  message: Do not use math.rand. Use `crypto/rand` instead.
+  message: Do not use `math/rand`. Use `crypto/rand` instead.
   languages: [go]
   severity: WARNING
-  patterns:
-  - pattern-either:
+  pattern-either:
+  - pattern: |
+      import mrand "math/rand"
```
dlukeomalley: Could these patterns be simplified using a `pattern-inside` clause with the below?

```
import mrand "math/rand"
...
```

mschwager: Yup! Good catch, that improves the `ruleid` location :+1: 

Updating...

colleend (approved):
lgtm!

PR #1100, opened 2021-02-11
Improve Javascript useless assign

### LINKED ISSUE #1098: "useless-assign" rule reporting false positives
**Describe the bug**
This rule: https://semgrep.dev/editor?registry=javascript.lang.correctness.useless-assign
reports findings on both of these and it should not:
```
// CASE A
// ok
q1 = result * 2;
q1 = q1 < 25 ? q1 : 0;

// CASE B
// ok
let progress = this.getProgress(file.name, percent);
progress = {...progress, loading: true};
```

**To Reproduce**
See above.

**Expected behavior**
Findings should not be reported for the two cases above.

**Additional Context**
Fixes:
(1) Unrelated, but there is a bug in the pre-existing tests in that rule.
The first few lines are not testing what was expected because of reusing a variable.
Should instead be
```
// ruleid:useless-assignment
var something_other_than_x = 1;

// ruleid:useless-assignment
x = 2;
x = 3;
console.log(x1);
```

(2) We can actually simplify the existing rule and still pass all the pre-existing tests with just this:
```
rules:
- id: useless-assignment
  patterns:

    - pattern: |
        $X = $Y;
        $X = $Z;

    - pattern-not: |
        $X = $Y;
        $X = <... $X ...>;       
```
This also covers my CASE A above but, alas, it does not cover CASE B, but I believe it should, so seems like a bug in the engine as well, perhaps.

Here it is in the playground: https://semgrep.dev/s/px90

mschwager:
Fixes https://github.com/returntocorp/semgrep-rules/issues/1098

### DIFF
--- a/javascript/lang/correctness/useless-assign.js
+++ b/javascript/lang/correctness/useless-assign.js
@@ -1,30 +1,41 @@
 // ruleid:useless-assignment
-var x = 1;
+var x1 = 1;
+x1 = 2;
 
 // ruleid:useless-assignment
-x = 2;
-x = 3;
-console.log(x);
+let x2 = 1;
+x2 = 2;
 
+// ruleid:useless-assignment
+x3 = 1;
+x3 = 2;
+
+// https://github.com/returntocorp/semgrep/issues/2552
+// todook:useless-assignment
+x4 = {value1: 42};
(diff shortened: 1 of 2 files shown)


inkz (approved):
:+1:

PR #1101, opened 2021-02-16
Remove lxml from use-defused-xml.

### LINKED ISSUE #1086: lxml should be removed from use-defused-xml
**Describe the bug**
defusedxml.lxml [was only ever an example](https://github.com/tiran/defusedxml#defusedxmllxml) and [is deprecated](https://github.com/tiran/defusedxml/issues/38). 

Ideally there should instead be a rule [checking that entity resolution is disabled](https://lxml.de/FAQ.html#how-do-i-use-lxml-safely-as-a-web-service-endpoint), and possibly [other lxml-related issues](https://github.com/tiran/defusedxml#xpath) though the latter may alresady exist.

**To Reproduce**
Run semgrep on a codebase with lxml, it tells you to remove lxml.

**Expected behavior**
It shouldn't do that.

minusworld:
Closes https://github.com/returntocorp/semgrep-rules/issues/1086.

### DIFF
--- a/python/lang/security/use-defused-xml.yaml
+++ b/python/lang/security/use-defused-xml.yaml
@@ -2,9 +2,8 @@ rules:
 - id: use-defused-xml
   pattern-either:
   - pattern: import xml
-  - pattern: import lxml
   fix-regex:
-    regex: (xml|lmxl)
+    regex: xml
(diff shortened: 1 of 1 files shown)


colleend (approved):
nice!

PR #1104, opened 2021-02-16
Consolidate template rules

minusworld:
This PR combines several HTML template rules for different engines (Flask/Jinja2, Django, and Mustache) that follow the same syntax.

These will be the official rules used in `p/xss`, to replace many variants that resulted in duplicate findings. See https://github.com/returntocorp/semgrep-rule-packs/issues/67 for more information.

Closes https://github.com/returntocorp/semgrep-rule-packs/issues/67 and https://github.com/returntocorp/enterprise/issues/548.


inkz (approved):
lgtm

PR #1109, opened 2021-02-19
Rename .test.yaml file to use .k8s extension

underyx:
Semgrep App and Semgrep CLI's pytest both got confused by having a .yaml
file here that is not a Semgrep rule.


minusworld (approved):
perfect 👌

aryx (approved):

PR #1111, opened 2021-02-19
Fix FP in django model-save rule

### LINKED ISSUE #1106: python.django.correctness.model-save.django-db-model-save-super False Positive
**Describe the bug**

The `python.django.correctness.model-save.django-db-model-save-super` rule appears to raise a false positive when the zero argument form of `super` is used to call the `.save` method.

**To Reproduce**

Create a python file containing the following:

```
from django.db import models

class MyModel(models.Model):

    def save(self, *args, **kwargs):
        print("my overriden save method")
        super().save(*args, **kwargs)


class MyOtherModel(models.Model):

    def save(self, *args, **kwargs):
        print("my overriden save method")
        super(MyOtherModel, self).save(*args, **kwargs)
```

and run semgrep:

```
semgrep --config p/r2c-best-practices mymodel.py
```

The first class is flagged by the `python.django.correctness.model-save.django-db-model-save-super` rule, and the second is not. 

**Expected behavior**

Both forms of calling the super `.save` method are equivalent, so neither should be flagged by this rule - https://docs.python.org/3/library/functions.html#super

minusworld:
Closes https://github.com/returntocorp/semgrep-rules/issues/1106.

### DIFF
--- a/python/django/correctness/model-save.py
+++ b/python/django/correctness/model-save.py
@@ -19,3 +19,18 @@ class Bad(models.Model):
     def save(self, *args, **kwargs):
         if not self.slug:
             self.slug = slugify(self.title)
+
+class ModelIssue1106(models.Model):
+
+    # ok: django-db-model-save-super
+    def save(self, *args, **kwargs):
+        print("my overriden save method")
+        super().save(*args, **kwargs)
+
+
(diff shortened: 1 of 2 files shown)


inkz (approved):
:+1:

PR #1115, opened 2021-02-23
html safe rule for erb files

colleend:
to catch this case: https://github.com/andrepcg/roasted-mail/commit/75b9c2a598e19dc66e74fba4722f0d2803e96749 for erb files

### DIFF
--- a/generic/html-templates/security/no-html-safe.erb
+++ b/generic/html-templates/security/no-html-safe.erb
@@ -1,0 +1,33 @@
+// ruleid: no-html-safe
+<%= @email.html.html_safe %>
+
+<html>
+  <head>
+    <title><%= ENV['META_TITLE'] %></title>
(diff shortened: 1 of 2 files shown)


minusworld:
Awesome!
Is there an equivalent rule for Ruby proper?
Also, might we want to restrict this to be inside ERB expressions? E.g.,
`pattern-inside: <%= ... %>`

minusworld:
Also, how does this compare to this rule here? https://github.com/returntocorp/semgrep-rules/blob/develop/ruby/rails/security/audit/xss/templates/avoid-html-safe.yaml

colleend:
Oh wow I somehow missed that rule! But yes, they do seem to be doing the same thing 😄  thanks for letting me know!! 

And we do seem to have Ruby html_safe rules so we're good on that aspect. I'll close this PR on the basis that we have this rule!

Issue #941, opened 2020-10-22
Move to one rule per YAML file


mschwager:
We currently allow one or more rules per YAML configuration file. This is nice for grouping multiple, similar rules together, but can make it difficult for other systems to analyze the configuration file <-> rules <-> tests relationship. Currently, `61 / 587` configuration files have > 1 rule:

```
$ rg -c '\- id:' | awk -F ':' '{if ($2 > 1) print $1 " " $2}'
ocaml/lang/performance/list.yaml 2
contrib/owasp/java/ssrf/ssrf.yaml 4
contrib/nodejsscan/open_redirect.yaml 2
contrib/nodejsscan/header_cors_star.yaml 2
contrib/nodejsscan/good_helmet_checks.yaml 12
contrib/nodejsscan/xss_templates.yaml 3
contrib/nodejsscan/exec_os_command.yaml 2
...
```

This ticket is still in the "debate" phase. Should we move towards enforcing one rule per configuration file?

Advantages:

- Easier for other systems to manage the configuration file <-> rule(s) <-> test relationship
- Easier mental model of how our rules work (one rule, one file)

Disadvantages:

- Lots of work (convert 61 rule files)
  - May be automateable, but then we're changing rule IDs for the sake of changing rule IDs
- Rule IDs will change, which has downstream effects like rulesets will break

Is this a good time to also consider removing path-dependent rule IDs and enforcing unique rule IDs? See https://r2c.quip.com/hIbYAewnv8Er/Extended-Internal-Dry-run#JAXACAFRZSD. Now that we've lived with the current implementation for quite a while and noted its deficiencies, perhaps we should rip the band-aid off and move towards Rule Format V2:

- Unique rule IDs
- Path-independent rule IDs
- One rule per configuration file

Food for thought...

cc @ievans

ievans:
From the registry perspective, it is a difficult to figure out which rule to highlight in the test case when we have more than 1 rule per file. And from a contributor UX thinking about eventual fork/commit back models with the live editor, I think having 1 rule=1 file (which is already the case for snippets) makes some sense. cc @DrewDennison

DrewDennison:
yep I'm good with 1 rule + file for our registry

kurt-r2c:
@minusworld is this something that we need to work on CI enforcement of? Should we be tracking that with this ticket or a separate one?

PR #1119, opened 2021-02-23
Update file_permission.yaml

chmccreery:
Do we need to update any links to the previous rule id from rulesets or anywhere else?

### DIFF
--- a/go/lang/correctness/permissions/file_permission.yaml
+++ b/go/lang/correctness/permissions/file_permission.yaml
@@ -1,5 +1,5 @@
 rules:
-- id: incorrect-default-permission
+- id: incorrect-default-permission-chmod
   message: Expect permissions to be `0600` or less for os.Chmod()
   metadata:
     cwe: 'CWE-276: Incorrect Default Permissions'
@@ -13,7 +13,7 @@ rules:
       metavariable: $PERM
       comparison: $PERM > 0o600
       base: 8
-- id: incorrect-default-permission
+- id: incorrect-default-permission-writefile
   message: Expect permissions to be `0600` or less for ioutil.WriteFile()
   metadata:
     cwe: 'CWE-276: Incorrect Default Permissions'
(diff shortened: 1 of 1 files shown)


mschwager:
I really like that you're doing this, however, changing rule IDs can be a painful task.There are 3 main areas I can see this breaking:

- Rulesets, like you mentioned. You can check for this rule ID here: https://github.com/returntocorp/semgrep-rule-packs
- `nosem` comments, this is a bit of a pain because they're in end-user's code.
- Semgrep App [findings](https://semgrep.dev/docs/managing-findings/#semgrep-ci). These are in our purview, but still a pain to update.

Soapbox: IMO we should enforce unique, immutable rule IDs _globally_, insofar as we can. I.e. Semgrep should error out if multiple rules use the same rule ID in a particular analysis/run. I know this information is mostly outside the scope of this task, but this task highlights why not doing so can be painful down the line. I've touched on this a bit before in:

- https://github.com/returntocorp/semgrep-rules/issues/941
- https://r2c.quip.com/hIbYAewnv8Er/Extended-Internal-Dry-run#JAXACAFRZSD

As for this task, I can see a few options: 1) do nothing - deal with the pain a bit longer until we can re-architect things, or 2) move fast and accept the potential breakage mentioned above. What do you think?

chmccreery:
Thanks for the input, @mschwager . For reference, this is meant to fix https://github.com/returntocorp/semgrep/issues/2537 

How about: 
3) combine these patterns under a pattern-either and deal with the error messages being less specific? Is that possible with a metavariable rule like this? 

(also fine with "do nothing until we fix registry" cc. @dlukeomalley )

mschwager:
I think that should work since they all specify `$PERM` and compare against the same value!

chmccreery:
Ok, hopefully fixed 🤞 thanks again for the help @mschwager.

See https://semgrep.dev/s/J91y/

Now, we gotta figure out where else in this code base we have similar issues. Sounds like a good time for Spacegrep on yaml!! @emjin

mschwager (approved):

PR #1122, opened 2021-02-25
Use C typed metavariables to check incorrect C-string comparison

mschwager:


colleend:
should we recommend strncmp instead of strcmp? 🤔 🤔

mschwager:
> should we recommend strncmp instead of strcmp? thinking thinking

There may be situations where `strncmp` is preferable, but not necessarily for security reasons. Since `str{n}cmp` doesn't write to memory, it only reads it, we don't have the same concerns as with things like `strcpy`.

I think `strcmp` provides the most general autofix since `strncmp` is situation dependent.

See also: https://stackoverflow.com/questions/30190460/advantages-of-strncmp-over-strcmp

colleend:
makes sense!

colleend (approved):
nice!!

PR #1123, opened 2021-02-26
fix false positives in grpc-server-insecure-connection rule

dmitris:
PR fixes the following false positive with gRPC server insecure connection rule:
```
	opts := []grpc.ServerOption{
		grpc.Creds(credentials.NewClientTLSFromCert(pool, addr)), grpc.UnaryInterceptor(auth.GRPCInterceptor)}
	grpcServer := grpc.NewServer(opts...)
```

The linked in the rule post https://blog.gopheracademy.com/advent-2019/go-grps-and-tls/#connection-without-encryption says (emphasis added):
> This, plus a Server **without any ServerOption** will result in an unencrypted connection.

When there are options with credentials, `grpc.NewServer` shouldn't be flagged.

I hope the spacing diffs in the .go file (autoformatted by the editor) are ok - if not, let me know and I'll try to remove them.

### DIFF
--- a/go/grpc/security/grpc-server-insecure-connection.go
+++ b/go/grpc/security/grpc-server-insecure-connection.go
@@ -53,6 +53,7 @@ func startServerWithOpts() {
 	options := []grpc.ServerOption{
 		grpc.Creds(credentials.NewClientTLSFromCert(pool, addr)),
 	}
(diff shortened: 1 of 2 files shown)


minusworld (approved):
🎉

Issue #731, opened 2020-07-17
[Check] Secret scanning


clintgibler:
**Check Description**

- [ ] This ticket has links, references, or examples.
- [ ] Your check has true positive and true negative test cases.
- [ ] Your check has been labeled with true positive and false positive findings on at least 10 repositories, and the link is pasted in the ticket. If there are no findings, paste the triager link anyway.
- [ ] Your check PR has been reviewed and merged.
Now, close the ticket!

Secrets detection tools we should review:
* [ ] TruffleHog v3 - https://github.com/trufflesecurity/TruffleHog
* [ ] https://github.com/dxa4481/truffleHog
  * https://github.com/dxa4481/truffleHog/blob/48ffdd3beee3b600939136fccad59f50a2610909/scripts/searchOrg.py#L10
* [ ] https://github.com/Yelp/detect-secrets
  * https://github.com/Yelp/detect-secrets/tree/master/detect_secrets/plugins
* [ ] https://github.com/thoughtworks/talisman
* [ ] https://github.com/Plazmaz/leaky-repo
* [ ] https://github.com/Skyscanner/whispers
* [ ] https://github.com/owasp/sedated
* [ ] https://github.com/michenriksen/gitrob/blob/master/core/signatures.go
* [ ] https://github.com/nielsing/yar
* [ ] [Token-Hunter](https://gitlab.com/gitlab-com/gl-security/security-operations/gl-redteam/token-hunter) - [blog post](https://about.gitlab.com/blog/2019/12/20/introducing-token-hunter/)
* [ ] [credential-digger](https://github.com/SAP/credential-digger)
* [ ] [gitlab-watchman](https://github.com/PaperMtn/gitlab-watchman)
* [ ] [gitleaks](https://github.com/zricethezav/gitleaks)
* [ ] https://semgrep.dev/p/hazanasec.possible_hardcoded_secrets
* [ ] [gitdorker](https://www.kitploit.com/2020/10/gitdorker-tool-to-scrape-secrets-from.html)
* [ ] https://github.com/internetwache/GitTools
* [ ] https://github.com/hisxo/gitGraber
* [ ] https://github.com/dwisiswant0/gf-secrets
* [ ] https://github.com/newrelic/rusty-hog
* [ ] https://github.com/datumbrain/aws-macie-pii-confidential-regexes/blob/master/regex_list.csv
* [ ] https://github.com/SnaffCon/Snaffler
* [ ] https://github.com/grab/secret-scanner
[...]

PR #1126, opened 2021-03-02
Add secret detection rules using regexes

minusworld:
Bunch of secret detection regexes. Related to https://github.com/returntocorp/semgrep-rules/issues/731.


inkz (approved):
:+1:

regex/secrets/detected-artifactory-token.yaml
```diff
@@ -0,0 +1,9 @@
+rules:
+- id: detected-artifactory-token
+  pattern-regex: |-
+    AKC[a-zA-Z0-9]{10,}
+  languages: [regex]
```
dlukeomalley, line 5: What is the `regex` language?

dlukeomalley (commented):
@minusworld I'm thrilled about this PR! Couple questions:

1. Is there any benefit to restricting regex to string literals? Curious if it helps with either accuracy and/or speed (reducing the search space, but increasing the parsing?)
2. Are there any file extensions that can be used to reduce the search space? I noticed while looking at other "generic" rules we haven't made a habit of it, and curious here if it would help  or be possible.

regex/secrets/detected-artifactory-token.yaml
minusworld: `regex` is an alias for `none`; it means that this rule only uses `pattern-regex` and will operate on text files.

minusworld:
> @minusworld I'm thrilled about this PR! Couple questions:
> 
> 1. Is there any benefit to restricting regex to string literals? Curious if it helps with either accuracy and/or speed (reducing the search space, but increasing the parsing?)
> 2. Are there any file extensions that can be used to reduce the search space? I noticed while looking at other "generic" rules we haven't made a habit of it, and curious here if it would help  or be possible.

(1) In my review of exposed secrets, I commonly found them in:
- Shell scripts of many varieties (.sh, .bash, .zsh, etc.)
- Code
- Logs

So, it's hard to say that they'll always be string literals. They could be hiding anywhere.

(2) Once merged, I'd like to run the rules at scale without any file restrictions. This will help us determine whether there are common files we can restrict to. However, I think for the largest search space, we won't want to restrict these. As mentioned above, secrets can be hiding in lots of places...

That said, this reminds me that our [rule benchmarks](https://github.com/returntocorp/semgrep-rules/actions/workflows/rule-benchmarks.yaml) don't run `generic` or `regex` rules... 🤔

PR #1129, opened 2021-03-02
Avoid running string concat rule in Python dict

mschwager:
Python has the unfortunate syntax of `{...}` referring to both sets and dicts. This means we were incorrectly detecting implicit string concatenation inside dicts as well as the intended lists and sets. This PR removes these false positives :+1:


minusworld (approved):
noice 👍

PR #1135, opened 2021-03-03
Fix avoid-pyyaml-load to handle multiple arguments

minusworld:


python/lang/security/deserialization/avoid-pyyaml-load.yaml
```diff
@@ -12,11 +12,13 @@ rules:
...
   patterns:
-  - pattern-inside: |
-      import yaml
+  - pattern-not-inside: |
+      yaml = ruamel.yaml.YAML(...)
       ...
-      yaml.load($FOO)
-  - pattern: yaml.load($FOO)
+      yaml.load(...)
+  - pattern: yaml.load(...)
```
underyx, line 28: This would match on our and probably many other projects' code, as we defined our own safe `yaml.load` to be used across the project.

underyx (changes requested):

python/lang/security/deserialization/avoid-pyyaml-load.yaml
underyx: To clarify, I'm suggesting that we bring the `import yaml` back into the pattern.

minusworld: Using the other `import yaml` pattern causes its own issues, as demonstrated in this snippet: https://semgrep.dev/s/eL2y/.

I suspect only the first `yaml.load(...)` is detected since it satisfies the `pattern-inside`.

Perhaps this is an OK trade-off, since the check "should" continue to fire after each subsequent fix, but I could also see that being very annoying... What are your thoughts?

underyx: How about this? https://semgrep.dev/s/vzPb/

minusworld: lgtm! Do you want to ferret out a test case from our code we can toss in to make sure it won't fire there?

underyx: Oh, this doesn't work :( https://semgrep.dev/s/dKzd/

But only when there's an import yaml and from internal.lib import yaml in the same file, which should be very rare 🤔

minusworld: Very rare is acceptable 😝

underyx (approved):

PR #1139, opened 2021-03-04
Add 'sscanf()' rule for string number conversions

mschwager:


minusworld (approved):
nice 👍  Good read in the references, too

PR #1141, opened 2021-03-04
Move secrets rules to generic/secrets/security

minusworld:
Since these are new, we still have time to move them before they get broader adoption. Wanted to make these match the "convention" in the repo


colleend (approved):
nice!

PR #1145, opened 2021-03-08
Add paths for noopener and noreferrer HTML rules

minusworld:
Prevents these `generic` rules from running on other files such as React code


minusworld:
Related to https://github.com/returntocorp/enterprise/issues/583

underyx (approved):

PR #1151, opened 2021-03-10
Update private key, add bcrypt, shadow, npm and sql

hazanasec:
Hi guys, 

I updated `detected-private-key.yaml` to catch a few more private keys as can see here: https://semgrep.dev/s/5QgX - I also added a few more secrets to scan: bcrypt hashes, linux shadow file. npm auth and sql dump.

Thanks,
Hazana


minusworld:
Hey @hazanasec, these look cool! Is there any way I can help you with these?

PR #1155, opened 2021-03-11
Add missing-ssl-minversion Go rule

### LINKED ISSUE #1148: TLS MinVersion too low false negative
https://semgrep.dev/s/dKOd

The following code triggers gosec error "G402 (CWE-295): TLS MinVersion too low":
`
return &tls.Config{
                 Certificates: []tls.Certificate{cert},
                 RootCAs:      rootCAs,
         }, nil
`
The reason is default-init with 0 means that omitted MinVersion parameter is effectively initialized with constant which corresponds to outdated TLSv1.0

I'm not yet familiar enough with semgrep rules to propose a meaningful fix.

minusworld:
Fixes https://github.com/returntocorp/semgrep-rules/issues/1148

### DIFF
--- a/go/lang/security/audit/crypto/missing-ssl-minversion.go
+++ b/go/lang/security/audit/crypto/missing-ssl-minversion.go
@@ -1,0 +1,70 @@
+package main
+
+import (
+	"crypto/tls"
+	"log"
+	"net/http"
+	"net/http/httptest"
+	"os"
+)
+
+// zeroSource is an io.Reader that returns an unlimited number of zero bytes.
+type zeroSource struct{}
+
+func (zeroSource) Read(b []byte) (n int, err error) {
+	for i := range b {
+		b[i] = 0
+	}
+
+	return len(b), nil
+}
+
+func main() {
+	// Dummy test HTTP server for the example with insecure random so output is
+	// reproducible.
+	server := httptest.NewUnstartedServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {}))
+	// ruleid: missing-ssl-minversion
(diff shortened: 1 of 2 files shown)


go/lang/security/audit/crypto/missing-ssl-minversion.yaml
```diff
@@ -0,0 +1,24 @@
+rules:
+- id: ssl-v3-is-insecure
```
mschwager:
```suggestion
- id: tls-v1-is-insecure
```

go/lang/security/audit/crypto/missing-ssl-minversion.go
```diff
@@ -0,0 +1,70 @@
...
+	// Typically the log would go to an open file:
+	// w, err := os.OpenFile("tls-secrets.txt", os.O_WRONLY|os.O_CREATE|os.O_TRUNC, 0600)
+	w := os.Stdout
+
+	client := &http.Client{
+		Transport: &http.Transport{
+			// ok: missing-ssl-minversion
+			TLSClientConfig: &tls.Config{
+				KeyLogWriter:       w,
+				MinVersion:         tls.VersionSSL30,
```
mschwager, line 42: Do we care that this is using `VersionSSL30`? I.e. is this only for insecure defaults, or also for explicitly insecure?

minusworld: This rule is for insecure defaults. We have another rule for explicitly insecure, i.e., checking for VersionSSL30.

mschwager (approved):

PR #1156, opened 2021-03-11
Use ellipsis in JSON pattern in fields

aryx:
In https://github.com/returntocorp/semgrep/pull/2741 we are converting
JSON records in a Dict, to be more consistent with what we do
with YAML, but a problem is that Record (and objects) allowed
the "..." to be ommitted, but Dict (and Set) require the "...".
This is yet another inconsistency. We should probably require
the "..." in object too, but in the mean time we can add it
for JSON so the semgrep PR above will pass.

test plan:
make test using semgrep-core of PR above.


aryx:
I'm sure @mschwager will rightly complain about inconsistencies.

minusworld (approved):

PR #1166, opened 2021-03-15
Better version of ESLint object injection rule

chgg-kboberg:
See https://github.com/nodesecurity/eslint-plugin-security/blob/master/docs/the-dangers-of-square-bracket-notation.md for context.

This rule is super noisy in some cases (such as using a constants.js design pattern to do bracket lookups from "variables" that are actually constants managed elsewhere).

PR adds a context-aware version that ignores some common false positives that ESLint spits out.


inkz (approved):
lgtm

PR #1179, opened 2021-03-23
Update dict-modify-iterating rule

kiwiz:
Previously, this would trigger on loops over `dictVal.keys()` and `dictVal.items()` which doesn't actually raise a `RuntimeError` in my testing. Updating to only trigger for `for v in dictVal`.

### DIFF
--- a/python/lang/correctness/dict-modify-iterating.py
+++ b/python/lang/correctness/dict-modify-iterating.py
@@ -1,19 +1,4 @@
-d = {'a': 1, 'b': 2}
+d = {"a": 1, "b": 2}
 # ruleid:dict-del-while-iterate
-for k,v in d.items():
+for k, v in d:
     del d[k]
-
-d = {'a': 1, 'b': 2}
-# ruleid:dict-del-while-iterate
-for k in d.keys():
-    del d[k]
-
-# ruleid:dict-del-while-iterate
-for k in d.keys():
-    print(d[k])
-    del d[k]
-
-# ok:dict-del-while-iterate
-for k in d.keys():
-    print(d[k])
-    x = d[k]
--- a/python/lang/correctness/dict-modify-iterating.yaml
+++ b/python/lang/correctness/dict-modify-iterating.yaml
@@ -1,15 +1,10 @@
 rules:
 - id: dict-del-while-iterate
   patterns:
-  - pattern-either:
-    - pattern: |
-        for $KEY, $VALUE in $DICT.items():
-            ...
-            del $DICT[$KEY]
-    - pattern: |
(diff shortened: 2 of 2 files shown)


aryx:
from slack conversation here: https://r2c-community.slack.com/archives/C018NJRRCJ0/p1616457439022300

ievans:
Hmm; per python documentation, I believe that the behavior when deleting items during iteration is non-deterministic. See https://docs.python.org/3/library/stdtypes.html#dict-views

> Iterating views while adding or deleting entries in the dictionary may raise a RuntimeError or fail to iterate over all entries.

kiwiz:
Ah, this is probably due to the differences between Py2 and Py3. In Py2, `.items()`, `.keys()` and `.values()` all returned an array, whereas in Py3 they return an iterator. Thus:

Py2:
```
>>> a = {"a": 1, "b": 2}
>>> for b in a:
...   del a[b]
...
  File "<stdin>", line 1, in <module>
RuntimeError: dictionary changed size during iteration
>>> a = {"a": 1, "b": 2}
>>> for b, _ in a.items():
...   del a[b]
...
>>> a = {"a": 1, "b": 2}
>>> for b, _ in a.iteritems():
...   del a[b]
...
  File "<stdin>", line 1, in <module>
RuntimeError: dictionary changed size during iteration
>>> a = {"a": 1, "b": 2}
>>> for b in a.keys():
...   del a[b]
...
>>>
```

Py3:
```
>>> a = {"a": 1, "b": 2}
>>> for b in a:
...   del a[b]
...
  File "<stdin>", line 1, in <module>
RuntimeError: dictionary changed size during iteration
>>> a = {"a": 1, "b": 2}
>>> for b, _ in a.items():
...   del a[b]
...
  File "<stdin>", line 1, in <module>
RuntimeError: dictionary changed size during iteration
>>> a = {"a": 1, "b": 2}
>>> for b in a.keys():
...   del a[b]
...
  File "<stdin>", line 1, in <module>
RuntimeError: dictionary changed size during iteration
>>> a = {"a": 1, "b": 2}
>>> for b in list(a.keys()):
...   del a[b]
...
>>>
```

If there's a way to mark rules as Py2-only, there could be a separate rule for Py3. Otherwise, we might want to update the description.

kiwiz:
See #1183

ievans:
@kiwiz unfortunately we don't have -- as far as I know -- a way to mark rules as _python3_ only. (I do think you can target python2 though).

@mschwager am I wrong on that?

I'm very sorry that your first contribution ended up not being merged! At least we have #1183; thanks for taking the time to make things better for the next person who is confused.

PR #1184, opened 2021-03-23
Fix FPs in detected-artifactory secrets rules. Added tests

### LINKED ISSUE #1178: FP in generic.secrets.security.detected-artifactory-password
**Describe the bug**
From a user on Slack:
btw, the secrets regexes is firing a bunch of false positives for us but thankfully, I think, the false positives are a good example of why semgrep is better than Yelp's original detect secrets mechanism.
https://semgrep.dev/editor?registry=generic.secrets.security.detected-artifactory-password
fires on package-lock.json files like
```
      "integrity": "sha512-12345678APAjZas32d1f32a1sd2fasdfasdf32a1sd32f1a3s2d1f...=="
```
because of the AP..
One option is to exclude package-lock.json from all of the secrets regexes; another is to modify the rule to say pattern-not-in "sha\d+-(base64string)

**Priority**
How important is this to you?

minusworld:
Fixes https://github.com/returntocorp/semgrep-rules/issues/1178


inkz (approved):
:+1:

generic/secrets/security/detected-artifactory-password.yaml
```diff
@@ -1,7 +1,10 @@
 rules:
 - id: detected-artifactory-password
-  pattern-regex: |-
-    AP[\dABCDEF][a-zA-Z0-9]{8,}
+  patterns:
+  - pattern-regex: |
+      \bAP[\dABCDEF][a-zA-Z0-9]{8,}
```
underyx, line 5: how come this is unbounded `{8,}` @minusworld? Aren't the tokens one or two pre-defined lengths?

minusworld: I copied these from another source; I set the word boundary `\b` on them in this PR. I am uncertain of the original research that went into them. I propose waiting and seeing if the issue persists. If it does, I don't think the unbounded regexes are a good approach and we should nuke these or figure out what the correct lengths are.

PR #1185, opened 2021-03-23
Contribute more regex for secrets

pierre-ernst:
Contribute regex from several sources


mschwager:
cc @minusworld for secrets expertise :+1:

.gitignore
```diff
@@ -1,3 +1,4 @@
 __pycache__/
 .benchmarks/
 .pytest_cache/
+.DS_Store
```
minusworld, line 4: 🤘

generic/secrets/security/detected-azure-service-bus-key.yaml
```diff
@@ -0,0 +1,9 @@
+rules:
+- id: detected-azure-service-bus-key
+  pattern-regex: |-
+    [A-Za-z0-9\/+-]{43}
```
minusworld: If I read this regex correctly, this will detect a lot of high entropy strings that may not be an Azure service bus key. Is there a way to modify the regex or use a `pattern-not-regex` to better detect when this is an Azure service bus key?

generic/secrets/security/detected-azure-storage-key.txt
```diff
@@ -0,0 +1,2 @@
+# ruleid: detected-azure-storage-key
+AccountKey=lJzRc1YdHaAA2KCNJJ1tkYwF/+mKK6Ygw0NGe170Xu592euJv2wYUtBlV8z+qnlcNQSnIYVTkLWntUO1F8j8rQ==
```
minusworld: It is helpful and good practice to include test cases that the rule should not detect. These can be annotated with `# ok: detected-azure-storage-key`. With both kinds of tests, we have documented examples of what the rule should and shouldn't detect, and it prevents any regressions from occurring when the rule gets updated, and it helps us make sure the rule only detects what it's supposed to detect.

Could you add some extra test cases for things this should not detect?

generic/secrets/security/detected-azure-storage-key.yaml
```diff
@@ -0,0 +1,9 @@
+rules:
+- id: detected-azure-storage-key
+  pattern-regex: |-
+    [A-Za-z0-9/+=]{88}
```
minusworld: Similar to `detected-azure-service-bus-key`, this looks like it may detect a lot of things. Can we get better accuracy to make sure the finding is an Azure Storage Key?

minusworld (changes requested):
Thanks for the rules! I have a few changes to recommend:
1. I'd like to have positive and negative test cases for all rules. This means that all `.yaml` files should have an associated test file. That test file should have `# ruleid:` and `# ok:` test cases.
2. Some of the regexes will match a lot more than the secret they are looking to find. Could we modify the regex or use something like `pattern-not-regex` to focus on detecting the actual secret?

Thanks again! Tag me on this PR if you have any questions or want help driving this forward.

I should also mention that we have a way of testing these rules by scanning lots of open source code at the same time. If you would like us to do that, let us know and we can get you the results.

Thanks!

ievans:
@pierre-ernst what's the status of this PR? Looks like the CLA is the only thing pending? Is that a blocker for you?

pierre-ernst:
Hi @ievans, I haven't had the time to work on this, but it is still on my TODO :-)
here is what I need to do:
1. improve detection and tests as per @\minusworld  suggestions
1. signing the CLA **is** a blocker, to whom should I forward our lawyers concerns?

ievans:
@pierre-ernst no worries! I can help you with the CLA, shoot an email to hello@r2c.dev, cc your legal PoC, and I'll reply. There's flexibility here; we've done corporate-wide CLAs and others.

pierre-ernst:
I've just realized that both Azure MSAL and ADAL tokens are actually JWT tokens, so no need to add specific detection rules.

minusworld:
OK, cool. Will you be removing those from this PR?

As for a plan for merging these in: there are some comments on the PR that still need to be addressed. Once they are resolved, we can skirt around the licence/cla check requirement since you've sent in a CLA independently.

@pierre-ernst Does that sound good to you? Thanks! :)

pierre-ernst:
Sounds good @minusworld

minusworld (approved):
OK, looks good! Thank you!

minusworld:
Thanks for bearing with us on the saga here! 😅

PR #1186, opened 2021-03-24
Fix last-user-is-root

mjambon:
This is motivated by a [fix in spacegrep](https://github.com/returntocorp/semgrep/pull/2803), which has to do with metavariables. The fix made the test fail, but the reasons are complicated. I'm not clear on why the rule appeared to work previously. The new rule is simpler as it doesn't rely on metavariables. It works with both the older and newer spacegrep.


inkz (approved):
:+1:

PR #1187, opened 2021-03-24
update dangerous-exec rule

inkz:
from https://github.com/returntocorp/enterprise/issues/507

the idea is to catch dynamic args if the command is "bash", "sh" etc


minusworld (approved):
nice. i really like this idea

PR #1190, opened 2021-03-25
Add rules for missing language and message fields

emjin:
Reformatted files in meta

### DIFF
--- a/meta/empty_message.test.yaml
+++ b/meta/empty_message.test.yaml
@@ -1,0 +1,8 @@
+rules:
+- id: empty-message
+  # ruleid: empty-message
+  message: |
+  languages: [yaml]
+  pattern: |
+    message: ""
+  severity: WARNING
--- a/meta/identical_id.yaml
+++ b/meta/identical_id.yaml
@@ -6,7 +6,7 @@ rules:
   #TODO: can't express it yet in yaml, some ... are not parsed correctly
   languages: [json, yaml]
   patterns:
-    - pattern-inside: |
-       [{id: $X, ...}, ..., {id: $X, ...}, ...]
-    - pattern: |
-       id: $X
+  - pattern-inside: |
+      [{id: $X, ...}, ..., {id: $X, ...}, ...]
+  - pattern: |
+      id: $X
--- a/meta/identical_id.yaml
+++ b/meta/identical_id.yaml
@@ -3,10 +3,13 @@ rules:
   message: |
     you use multiple times the same id
   severity: ERROR
-  #TODO: can't express it yet in yaml, some ... are not parsed correctly
-  languages: [json, yaml]
+  languages: [yaml]
   patterns:
   - pattern-inside: |
-      [{id: $X, ...}, ..., {id: $X, ...}, ...]
+      - id: $X
+        ...
+      ...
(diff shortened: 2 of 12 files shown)


meta/identical_id.yaml
```diff
@@ -6,7 +6,7 @@ rules:
   #TODO: can't express it yet in yaml, some ... are not parsed correctly
```
aryx: actually now you can remove the json and write directly the pattern in yaml

meta/no_language.yaml
```diff
@@ -0,0 +1,15 @@
+rules:
```
aryx, line 1: can you rename the file empty_language.test.yaml?
@mschwager are we ready to have the special logic in semgrep to not use this .test.yaml extension as input rules but just as a test files?

meta/no_message.yaml
```diff
@@ -0,0 +1,15 @@
+rules:
```
aryx, line 1: Same, rename empty_message.test.yaml

aryx (changes requested):

meta/no_message.yaml
emjin: I have a different test called empty_message.yaml, which checks if you have a message field but the text is "". Should I merge these two? Or do you have a name suggestion for that one?

aryx: I'm not sure I understand what you mean. In this PR there is a YAML rule, empty_message.yaml that detects empty messages, and a YAML test file, no_message.yaml. I just want you to rename no_message.yaml to empty_message.test.yaml (with the .test.yaml extension) to better see what is a rule and what is a test file. Otherwise with semgrep rules about semgrep rules it can be quite confusing.

aryx (changes requested):

aryx:
also by using a different .test.yaml for our test file, we can easily filter those yaml files when we look for a set of semgrep rules in a directory.

meta/no_language.yaml
mschwager: Yup, still tracking in https://github.com/returntocorp/semgrep/issues/2799. I haven't had the time to implement it, but I think we can go with the `.test.yaml` solution if someone has time to work on it. It should be reasonably quick to implement by modifying the logic here to also consider `.test.yaml` test files when the extension is `.yaml`: https://github.com/returntocorp/semgrep/blob/2508934/semgrep/semgrep/test.py#L282

mschwager:
https://github.com/returntocorp/semgrep/issues/2799 is in if you were waiting on that functionality :+1:

emjin:
Oops I was waiting for the tests.yaml decision to be finalized and then forgot about this---should be good to merge now

meta/empty_message.test.yaml
```diff
@@ -0,0 +1,8 @@
+rules:
+- id: empty-message
+  # ruleid: empty-message
+  message: |
+  languages: [yaml]
```
aryx, line 5: This is a bit confusing. The .test.yaml could be any kind of rule, like [c]. Using [yaml] here and having the test almost identical to the rule is confusing.

aryx (approved):

PR #1194, opened 2021-03-26
Add dockerfile rule to prefer copy over add

lbbooleans:
I created a generic rule for Dockerfiles that checks if the ADD command is used instead of COPY, since ADD has some side effects, as also explained at https://snyk.io/blog/10-docker-image-security-best-practices/


minusworld:
I added a specific reference about the side effects to the message. That way when it crops up, it'll save someone from hopping to the link to understand exactly what's up.

Thanks! Looks great. I'll merge it in ASAP.

minusworld (approved):

lbbooleans:
Great, thanks for merging!

minusworld:
Thanks for contributing! :)

PR #1199, opened 2021-03-29
fix: Resolve than that typo

### LINKED ISSUE #1198: Typographic error in "is an anti-pattern than"
**Describe the bug**
The rule typescript.react.security.audit.react-css-injection.react-css-injection references: "`User controlled data in a `style` attribute is an anti-pattern than can lead to XSS vulnerabilities`". This should be "`that can lead to XSS vulnerabilities`"

**To Reproduce**
`style={style}`

**Expected behavior**
I expect the rule to properly state the issue.

**Priority**
How important is this to you?

**Additional Context**
N/A. Perhaps "`then`" could also work here? Than does not. 

I've cut a PR to fix this, which I'll edit here: https://github.com/returntocorp/semgrep-rules/pull/1199

haasonsaas:
Resolves https://github.com/returntocorp/semgrep-rules/issues/1198

### DIFF
--- a/javascript/browser/security/insecure-document-method.yaml
+++ b/javascript/browser/security/insecure-document-method.yaml
@@ -13,7 +13,7 @@ rules:
       $EL.outerHTML = "...";
   - pattern-not: document.write("...")
   message: |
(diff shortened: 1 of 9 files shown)


inkz (approved):
thank you!

PR #1200, opened 2021-03-29
Fix 'return-in-init' rule

mschwager:


python/lang/correctness/return-in-init.py
```diff
@@ -1,43 +1,46 @@
 class A:
-
     def __init__(a, b, c):
         # ruleid:return-in-init
         return A(a, b, c)
 
-class A:
-    print ('hello')
+
+class B:
```
mschwager, line 7: This case wasn't caught previously because we were only detecting `return ...` not `return`. And another bug mentioned below.

python/lang/correctness/return-in-init.yaml
```diff
@@ -2,11 +2,13 @@ rules:
 - id: return-in-init
   patterns:
   - pattern-inside: |
-      class A(...):
+      class $A(...):
```
mschwager, line 5: We were only catching classes named `A` instead of using a metavariable :O

colleend (approved):
🙏  thanks for doing this! LGTM

python/lang/correctness/return-in-init.yaml
ievans: Nice catch!

PR #1201, opened 2021-03-29
Colleend/return to init fix

colleend:
* added in test case for the fp we were seeing
* tweaked rule a bit


mschwager:
I also created this fix: https://github.com/returntocorp/semgrep-rules/pull/1203. The nice thing there is the findings stay on the line near the `return`/`yield`. Either solution LGTM :+1:

minusworld:
Addressed by https://github.com/returntocorp/semgrep-rules/pull/1203

PR #1202, opened 2021-03-30
add command-injection-process-builder rule

inkz:
from https://github.com/returntocorp/enterprise/issues/586


minusworld (approved):
NICE. 🤘

PR #1204, opened 2021-03-30
Improve 'missing-pip-no-cache-dir'

mschwager:
This fixes FPs when using `PIP_NO_CACHE_DIR`, and finds additional usage when using `pip2` or `python -m pip`.


minusworld (approved):
nice 👍

PR #1205, opened 2021-03-30
Correct protocol names for Java SSLContext

lonemeow:
The weak `SSLContext` warning would trigger for correct usage, even according to the message in the rule. Also expanded the tests to check all the known weak names.

Ref: https://github.com/openjdk/jdk/blob/2499ac3db5c4a336059263a929031b4eef2d0f0a/src/java.base/share/classes/sun/security/ssl/SunJSSE.java#L97-L108


mschwager:
Good catch! Thanks for the PR. This looks good to me, we just need to get CLA approval :+1:

mschwager (approved):
Thanks again!

PR #1209, opened 2021-03-31
add seam-log-injection rule

inkz:
from https://github.com/returntocorp/enterprise/issues/586


mschwager (approved):
Wat, why on earth would you want your logging framework executing arbitrary code :exploding_head:

PR #1210, opened 2021-03-31
Add JS rule for zlib async in a loop

mschwager:
I was searching the [Node.js docs](https://nodejs.org/api/all.html) for instances of "warning" or "security" to look for new rule ideas, and this was the only thing I found that we don't already have. See here for more information: https://nodejs.org/api/zlib.html#zlib_threadpool_usage_and_performance_considerations. Note the explicit callout of synchronous calls:

> ... except those that are explicitly synchronous ...

### DIFF
--- a/javascript/lang/best-practice/zlib-async-loop.yaml
+++ b/javascript/lang/best-practice/zlib-async-loop.yaml
@@ -1,0 +1,19 @@
+rules:
+- id: zlib-async-loop
+  patterns:
+  - pattern-inside: |
+      for (...) {
+          ...
+      }
+  - pattern: zlib.$METHOD(...);
+  - metavariable-regex:
+      metavariable: $METHOD
+      regex: ^.+$(?<!Sync)
+  message: |
+    Creating and using a large number of zlib objects simultaneously
+    can cause significant memory fragmentation.
+  metadata:
+    references:
+    - https://nodejs.org/api/zlib.html#zlib_threadpool_usage_and_performance_considerations
+  severity: WARNING
+  languages: [js]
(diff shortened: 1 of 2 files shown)


javascript/lang/best-practice/zlib-async-loop.yaml
```diff
@@ -0,0 +1,19 @@
...
+      for (...) {
+          ...
+      }
+  - pattern: zlib.$METHOD(...);
+  - metavariable-regex:
+      metavariable: $METHOD
+      regex: ^.+$(?<!Sync)
+  message: |
+    Creating and using a large number of zlib objects simultaneously
+    can cause significant memory fragmentation.
```
minusworld: What should someone do instead? Stay outside of the loop, or use deflateSync, or something else?

mschwager: According to the docs:

> It is strongly recommended that the results of compression operations be cached to avoid duplication of effort.

I'm not entirely sure what a correct implementation would look like, but I suspect they're saying you should cache the `payload` -> `buffer`, and if you see the same payload again use the cache (memoize). I can add this bit to the message, what do you think? I think the fix will be application-dependent for the most part, e.g. add an `await`, use `deflateSync`, memoize, etc.

inkz (approved):
extended the rule a little bit :v:

javascript/lang/best-practice/zlib-async-loop.yaml
minusworld: Gotcha. Would you mind editing the message to how the non-Sync API causes memory fragmentation, and that solutions include `await`, `deflateSync`, or memoizing the result? I think this messaging helps educate developers who end up running across this issue.

PR #1213, opened 2021-04-01
Update python command injection rules

inkz:
from: https://github.com/returntocorp/enterprise/issues/586

### DIFF
--- a/go/lang/security/audit/unsafe-reflect-by-name.yaml
+++ b/go/lang/security/audit/unsafe-reflect-by-name.yaml
@@ -1,0 +1,27 @@
+rules:
+- id: unsafe-reflect-by-name
+  patterns:
+  - pattern-either:
+    - pattern: |
+        $SMTH.MethodByName($NAME,...)
+    - pattern: |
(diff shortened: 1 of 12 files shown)


colleend (approved):
Nice!

go/lang/security/audit/unsafe-reflect-by-name.yaml
```diff
@@ -0,0 +1,27 @@
...
+  patterns:
+  - pattern-either:
+    - pattern: |
+        $SMTH.MethodByName($NAME,...)
+    - pattern: |
+        $SMTH.FieldByName($NAME,...)
+  - pattern-not: |
+      $SMTH.MethodByName("...",...)
+  - pattern-not: |
+      $SMTH.FieldByName("...",...)
```
colleend, line 12: Would you happen to know if constant propagation will make it so we don't match cases like:
```
constvar1 = "const"
catch = valueJ.FieldByName(constvar1).String()
```

inkz: yep :+1: constant propagation works really good in this case!

PR #1215, opened 2021-04-02
add dangerous-globals-use rule

inkz:
from: https://github.com/returntocorp/enterprise/issues/586

### DIFF
--- a/python/lang/security/dangerous-globals-use.yaml
+++ b/python/lang/security/dangerous-globals-use.yaml
@@ -1,30 +1,30 @@
 rules:
 - id: dangerous-globals-use
   patterns:
-    - pattern-either: 
-      - pattern: globals().get(...)
-      - pattern: locals().get(...)
-      - pattern: globals()[...]
-      - pattern: locals()[...]
-      - patterns:
-          - pattern-either:
-            - pattern-inside: |
-                $G = globals()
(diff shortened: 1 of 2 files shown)


python/lang/security/dangerous-globals-use.yaml
```diff
@@ -0,0 +1,31 @@
+rules:
+- id: dangerous-globals-use
+  patterns:
+  - pattern-either:
+    - pattern: globals().get(...)
+    - pattern: locals().get(...)
```
minusworld, line 6: `.get(...)` is a helper method for dictionary access - we'll also need to capture `globals()[...]` which does the same thing

minusworld: Like this: 

```
>>> locals()['__builtins__'].__dict__['abs'](-12)
12
```

minusworld: Otherwise, looks good! 👍

inkz: Oh yeah, my bad! :+1:

inkz: done!

colleend (approved):
lgtm!!

PR #1218, opened 2021-04-05
Add privileged, seccomp, and selinux rules for docker-compose/k8s

mschwager:
Working toward:

- https://github.com/returntocorp/semgrep-rules/issues/1162
- https://github.com/returntocorp/enterprise/issues/595
- https://github.com/returntocorp/enterprise/issues/591

PR #1224, opened 2021-04-06
Add basic docker-compose/k8s OWASP cheat sheet rules

### LINKED ISSUE #1162: [Rule] Additional Docker rules based on OWASP Cheat Sheet
**Rule Description**
https://cheatsheetseries.owasp.org/cheatsheets/Docker_Security_Cheat_Sheet.html

Note the list of other static analysis tools under [Rule 9](https://cheatsheetseries.owasp.org/cheatsheets/Docker_Security_Cheat_Sheet.html#rule-9-use-static-analysis-tools).

*What does this rule intend to find?*

**Examples or references**

*Put an example or references here*

**Additional information**

*More information that would help someone write this rule!*

**PR Checklist**
- [ ] This ticket has links, references, or examples.
- [ ] The rule has **true positive** and **true negative** test cases in a file that matches the rule name.

> If the rule is `my-rule`, the test file name should be `my-rule.js`.
>
> True positives are marked by comments with `ruleid: <my-rule>` and true negatives are marked by comments with `ok: <my-rule>`.

- [ ] The rule has a good message. A good message includes:

> 1. A description of the pattern (e.g., missing parameter, dangerous flag, out-of-order function calls).
> 1. A description of why this pattern was detected (e.g., logic bug, introduces a security vulnerability, bad practice).
> 1. An alternative that resolves the issue (e.g., use another function, validate data first, discard the dangerous flag).

- [ ] After the PR has been reviewed and approved, merge the rule and close this ticket! Thanks for contributing!

mschwager:
Fixes #1162.

This doesn't have everything from [here](https://cheatsheetseries.owasp.org/cheatsheets/Docker_Security_Cheat_Sheet.html), but it's a good start for high signal/low noise findings that can easily be detected statically. The following rules from that cheat sheet are covered: 1, 2, 4, 6, 8.

### DIFF
--- a/yaml/kubernetes/security/exposing-docker-socket-hostpath.yaml
+++ b/yaml/kubernetes/security/exposing-docker-socket-hostpath.yaml
@@ -1,0 +1,24 @@
+rules:
+- id: exposing-docker-socket-hostpath
+  patterns:
+  - pattern-inside: |
+      volumes:
+        ...
+  - pattern: |
+      hostPath:
+        ...
+        path: /var/run/docker.sock
+  message: |
+    Container is running with a writable root filesystem. This may
+    allow malicious applications to download and run additional payloads, or
+    modify container files. If an application inside a container has to save
+    something temporarily consider using a tmpfs. Add 'readOnlyRootFilesystem: true'
+    to this container to prevent this.
+  metadata:
+    references:
+    - https://kubernetes.io/docs/concepts/storage/volumes/#hostpath
+    - https://kubernetes.io/docs/concepts/policy/pod-security-policy/#volumes-and-file-systems
+    - https://kubernetes.io/docs/tasks/configure-pod-container/security-context/
+    - https://cheatsheetseries.owasp.org/cheatsheets/Docker_Security_Cheat_Sheet.html#rule-1-do-not-expose-the-docker-daemon-socket-even-to-the-containers
+  languages: [yaml]
+  severity: WARNING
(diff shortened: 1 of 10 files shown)


yaml/kubernetes/security/exposing-docker-socket-hostpath.yaml
```diff
@@ -0,0 +1,24 @@
...
+  - pattern: |
+      hostPath:
+        ...
+        path: /var/run/docker.sock
+  message: |
+    Container is running with a writable root filesystem. This may
+    allow malicious applications to download and run additional payloads, or
+    modify container files. If an application inside a container has to save
+    something temporarily consider using a tmpfs. Add 'readOnlyRootFilesystem: true'
+    to this container to prevent this.
```
colleend: Is the message supposed to mention docker sockets instead? 🤔

colleend (commented):
yay more rules!!

I don't think I saw the rule for #4 -- perhaps that one wasn't pushed?

yaml/kubernetes/security/exposing-docker-socket-hostpath.yaml
mschwager: Oops, yeah this is totally the wrong message. I'll update :+1:

mschwager:
> yay more rules!!
> 
> I don't think I saw the rule for #4 -- perhaps that one wasn't pushed?

Good catch, the rules for 4 were actually in this separate PR: https://github.com/returntocorp/semgrep-rules/pull/1218

colleend (approved):
Nice!

PR #1231, opened 2021-04-08
Add a workflow step in benchmark job to print semgrep version

minusworld:
Adds a step to benchmark job to display the semgrep version


colleend:
ohhh nice!!

colleend (approved):

PR #1232, opened 2021-04-09
[Nodejsscan] fix after new import/require equivalences in semgrep

aryx:
Thanks to https://github.com/returntocorp/semgrep/issues/2882
some rules are now redundant (and cause CI failures because
semgrep --test does not allow a line to be matched by multiple rules).

test plan:
make test after make install latest semgrep-core


aryx:
@ajinabraham, semgrep now supports aliasing of require/import in Javascript, so a pattern like
child_process.exec(...) will match code like
```
var { exec } = require("child_process");
exec("foo")
```
or 
```
import { exec } from "child_process";
exec("foo")
```
or even if exec was aliased, like we do for Python,
so some of the nodejsscan rules are now redundant, like the one I deleted in this PR.

aryx:
We need semgrep-develop to be updated for the CI to work (I just merge the related change in semgrep).

DrewDennison (approved):

PR #1233, opened 2021-04-09
Fix FPs in route_var poritions of Flask injection rules

minusworld:
This is due to a bug where metavariables are not propagating properly with deep expression operators.

This refactors the rules to put the route definition and the sink in the same pattern clause to get around this issue.


colleend (approved):
lgtm!! thanks for doing this 😁

PR #1235, opened 2021-04-09
go database/sql sqli rule

colleend:
Go SQLi rules. Includes support for:
* [x] database/sql driver
* [x] go-pg driver (popular ORM for postgresql)
* [x] pgx driver (popular for postgresql, replaced lib/pq)

### DIFF
--- a/go/lang/security/audit/sqli/gosql-sqli.go
+++ b/go/lang/security/audit/sqli/gosql-sqli.go
@@ -1,75 +1,72 @@
-package main
+ipackage main
 
 import "database/sql"
 import "fmt"
 
-func setup() {
+func bad1() {
     db, err := sql.Open("mysql", "theUser:thePassword@/theDbName")
     if err != nil {
         panic(err)
     }
-}
-
-func bad1() {
     query = "SELECT name FROM users WHERE age=" + req.FormValue("age")
     // ruleid: gosql-sqli
     db.Query(query)
 }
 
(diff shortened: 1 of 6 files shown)


go/lang/security/audit/sqli/gosql-sqli.go
```diff
@@ -1,75 +1,72 @@
-package main
+ipackage main
```
minusworld: Is `ipackage` a typo?

minusworld: A `vim` oopsie maybe? lol

minusworld (commented):
Do you think these rules would benefit from the Go typing in Semgrep? I'm not sure how robust it is currently. cc @emjin

go/lang/security/audit/sqli/gosql-sqli.go
colleend: good catch!!! definitely a vim oopsie 😅

colleend:
Good question!! I think these rules are good even without Go typing (as we're mainly trying to catch variations of string concatenation that goes into a SQL query). My thought is that constant propagation is more helpful in this case, as we can use it to catch scenarios where a literal string is first assigned to a variable and then concatenated to a user-controlled variable 🤔 .

minusworld:
sounds good!

minusworld (approved):

Issue #1097, opened 2021-02-09
Add rules for scanning Kubernetes manifests


underyx:
For inspiration: https://polaris.docs.fairwinds.com/checks/security/

clintgibler:
Also a bunch of rules here:
* https://github.com/bridgecrewio/checkov
* https://github.com/accurics/terrascan

And other infrastructure as code scanning tools :)

mschwager:
I'm not sure how complete of a K8s solution you're looking for, but the following PRs included a good amount of security rules:

- https://github.com/returntocorp/semgrep-rules/pull/1218
- https://github.com/returntocorp/semgrep-rules/pull/1223
- https://github.com/returntocorp/semgrep-rules/pull/1224

Notably missing from the Fairwinds list: the `host*Set` rules and `securityContext.capabilities` rules. I wasn't sure about the severity and/or efficacy (FPs, FNs) of those heuristics, so I avoided them for now. Someone with more K8s experience than me could speak to those heuristics better.

mschwager:
Closing this out from https://github.com/returntocorp/semgrep-rules/pull/1236 as it and previous PRs fulfill "Add rules for scanning Kubernetes manifests". If we'd like additional K8s rules let's create specific issues for those :+1:

daghan:
@mschwager does this mean we support checkov rules?

mschwager:
> @mschwager does this mean we support checkov rules?

Here are our rules and checkov's:

- https://github.com/returntocorp/semgrep-rules/tree/develop/yaml/kubernetes/security
- https://github.com/bridgecrewio/checkov/tree/master/checkov/kubernetes/checks

Looks like we cover some, but not all, of checkov's rules. We have some overlap in Terraform and Dockerfile too, but one advantage we have is Docker Compose support.

PR #1236, opened 2021-04-12
Add host* key rules

mschwager:
This should cover fixing https://github.com/returntocorp/semgrep-rules/issues/1097. This doesn't cover everything there, but the task is very generally defined - if we're looking for something more specific we can create a new issue :+1: 

I considered adding a rule for `hostPorts`, but it seems too general, and susceptible to FPs IMO. You can easily overexpose ports with that key, but we don't have any way of knowing that without more application context.


colleend (approved):
lgtm!!

PR #1247, opened 2021-04-20
Node js sqli rules

colleend:
For driver pg

### DIFF
--- a/javascript/lang/security/audit/sqli/node-postgres-sqli.yaml
+++ b/javascript/lang/security/audit/sqli/node-postgres-sqli.yaml
@@ -1,0 +1,63 @@
+rules:
+- id: node-postgres-sqli
+  patterns:
+  - pattern-either:
+    - patterns:
+      - pattern-either:
+        - pattern: |
+            $DB.$METHOD(...,$QUERY,...)
+        - pattern: |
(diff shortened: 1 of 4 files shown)


javascript/lang/security/audit/sqli/node-postgres-sqli.yaml
```diff
@@ -0,0 +1,63 @@
...
+        const $NEWPOOL = new $POOL(...)
+        ...
+        $NEWPOOL.connect((..., $DB, ...) => {
+            ...
+        })
+  - pattern-not: $DB.$METHOD(..., "..." + "...", ...)
+  - pattern-not: $DB.$METHOD(..., $X.concat("..."), ...)
+  - metavariable-regex:
+      metavariable: $METHOD
+      regex: ^(query)$
```
mschwager: If `$METHOD` can only be `query` I think we can significantly speed up this rule by specifying the constant `query` everywhere instead of using a metavariable.

minusworld: I second this 👍

inkz: done!

inkz (approved):

PR #1250, opened 2021-04-20
Filter string concat query when it's only string literals

### LINKED ISSUE #1249: go.lang.security.audit.database.string-formatted-query can not support go sql prepared statement
**Describe the bug**
A clear and concise description of what the bug is and for which rule. Is the rule producing a false positive? Is it incorrect or missing something?

the rule:go.lang.security.audit.database.string-formatted-query can not support go sql prepared statement, such db.Exec('select * from table where key=?', req.id), the rule will produce a false positive.

**To Reproduce**
```
184:    insertSql := "insert into t_ad_experiment (exp_layer,buckets,opposite_buckets,is_transparent, " +
185:            " description,is_full,start_time,end_time,creat_time,update_time,update_user,white_list,extra,status)" +
186:            " value (?,?,?,?,?,?,?,?,?,?,?,?,?,?)"
187:    t := time.Now().Unix()
188:    InsertResult, err := DbConn.Exec(insertSql, info.Exp_layer, info.Buckets, info.Opposite_buckets,
189:            info.Is_transparent, info.Description, info.Is_full, info.Start_time, info.End_time, t, t,
190:            session.User, info.White_list, info.Extra, 0)
```
**Expected behavior**
support prepare statement identify

**Priority**
How important is this to you?
- [x] P0: blocking me from making progress

**Additional Context**

minusworld:
Closes https://github.com/returntocorp/semgrep-rules/issues/1249

### DIFF
--- a/go/lang/security/audit/database/string-formatted-query.yaml
+++ b/go/lang/security/audit/database/string-formatted-query.yaml
@@ -11,6 +11,13 @@ rules:
     cwe: "CWE-89: Improper Neutralization of Special Elements used in an SQL Command ('SQL Injection')"
     source-rule-url: https://github.com/securego/gosec
   patterns:
+  - metavariable-regex:
+      metavariable: $SINK
+      regex: (Exec|ExecContext|Query|QueryContext|QueryRow|QueryRowContext)
+  - pattern-not-inside: |
+      $VAR = "..." + "..."
+      ...
+      $OBJ.$SINK(..., $VAR, ...)
   - pattern-either:
     - pattern: $OBJ.Exec("..." + $X)
     - pattern: $OBJ.ExecContext($CTX, "..." + $X)
--- a/go/lang/security/audit/database/string-formatted-query.yaml
+++ b/go/lang/security/audit/database/string-formatted-query.yaml
@@ -11,9 +11,6 @@ rules:
     cwe: "CWE-89: Improper Neutralization of Special Elements used in an SQL Command ('SQL Injection')"
     source-rule-url: https://github.com/securego/gosec
   patterns:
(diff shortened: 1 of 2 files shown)


go/lang/security/audit/database/string-formatted-query.yaml
```diff
@@ -11,6 +11,13 @@ rules:
...
   patterns:
+  - metavariable-regex:
+      metavariable: $SINK
+      regex: (Exec|ExecContext|Query|QueryContext|QueryRow|QueryRowContext)
+  - pattern-not-inside: |
+      $VAR = "..." + "..."
+      ...
+      $OBJ.$SINK(..., $VAR, ...)
   - pattern-either:
     - pattern: $OBJ.Exec("..." + $X)
```
mschwager, line 19: Can we use `$OBJ.$SINK` for the rest of these `pattern`s too, and leverage the `metavariable-regex`? That would cut this rule down by about 10x.

mschwager (approved):

go/lang/security/audit/database/string-formatted-query.yaml
minusworld: We could, but a pattern with all metavariables involved tends to reduce performance. I was able to improve some JavaScript rules that performed this way by expanding the metavariable regex into distinct pattern clauses, which I suppose helps semgrep-core more easily identify which patterns to match.

minusworld: The reason it works a little better for this `pattern-not-inside` is because it's restricted to match only when there is some assignment with multiple string literals, which serves as sort of an "anchor"

PR #1253, opened 2021-04-21
Change language 'none' -> 'regex'

mschwager:
I think `regex` is more descriptive than `none`. Hopefully we can deprecate and remove `none` in favor of `regex` someday.


underyx:
@mschwager how about `language: any`? Or does this have to be a keyword that semgrep recognizes?

mschwager:
> @mschwager how about `language: any`? Or does this have to be a keyword that semgrep recognizes?

Yeah, unfortunately that ship has sailed unless we want to make `semgrep` changes: https://github.com/returntocorp/semgrep/blob/develop/semgrep/semgrep/semgrep_types.py#L25

minusworld (approved):

PR #1256, opened 2021-04-22
Add rule for no broken URLs

underyx:
@minusworld @mschwager I tried to do this as a proof of concept but cannot get a metavariable to match the full URL because spacegrep sees lots of separators within. It also seems like `pattern-where-python` doesn't give access to the full match, only the metavars. Do you see a workaround, or will this need a Semgrep CLI feature request?

### DIFF
--- a/generic/no-broken-urls.md
+++ b/generic/no-broken-urls.md
@@ -1,0 +1,9 @@
+# URLs
+
+This is a working URL: <https://httpbin.org/status/200>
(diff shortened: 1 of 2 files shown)


minusworld:
`pattern-where-python` has been deprecated. Therefore, I am closing this request. If you wish to continue the discussion, please re-open the thread. Thanks!

PR #1259, opened 2021-04-23
Allow further attributes after the method

mzehrer:
It is not uncommon that other attributes follow the method attribute or that the method attribute is not the last in line.


mschwager (approved):
Nice catch, thanks!

PR #1266, opened 2021-04-27
add rule to prevent raw query with sqlalchemy

0xdbe:
New rule for SQLAlchemy to prevent raw query execution.


minusworld (approved):
This is an excellent rule, thank you!

@0xdbe Would you mind if we included this in a SQLi rule pack we're working on? We hope to publish it in the next week or two.

0xdbe:
@minusworld This rule can be included in this rule pack.

PR #1273, opened 2021-05-07
fix: return-in-init should ignore None returns

ankush:
Currently this rule flags folllowing code as error:
```python
def __init__(self):
    return
```

As noted in documentation, the runtime error ONLY occurs when a non-None value is returned. So this is a false positive.

Reference:
- https://docs.python.org/3/reference/datamodel.html#object.__init__


minusworld (approved):
lgtm! Thanks!

PR #1274, opened 2021-05-07
Change meta tests to use a non-yaml rule

emjin:
For clarity, because otherwise the rule and test are easily confused


aryx (approved):
Thx!

PR #1276, opened 2021-05-10
Make meta rules more reliable

underyx:
Noticed all these `no_languages` and `no_message` false positives at https://gitlab.com/underyx/semgrep/-/jobs/1249670731.

Then figured I might as well make most these rules consistent.


underyx:
@mschwager requested review from you just cause you're online first 😇

meta/no_language.yaml
```diff
@@ -4,12 +4,10 @@ rules:
...
+  - pattern-inside: 'rules: [..., $RULE, ...]'
+  - pattern: $RULE
   - pattern-not: |
-      rules:
-      - ...
-        languages: ...
-        ...
+      ...
+      languages: ...
+      ...
```
mschwager: We should remove the trailing ellipsis in these rules. It's just going to make them slower, and return more, unnecessary data.

mschwager (approved):

meta/no_language.yaml
underyx: Good eye! Fixed, thanks!

PR #1278, opened 2021-05-11
Remove unnecessary parent operators

mschwager:
This removes unnecessary parent `patterns` and `pattern-either` operators when they contain a single `pattern`. This change may also marginally improve performance and debuggability.


colleend (approved):
Nice! Thanks for doing this 🙏

PR #1283, opened 2021-05-13
removing unnecessary ... from js rules

inkz:
removed unnecessary `...` from express js rules and slightly updated `pattern-inside`s that are used for express middleware matching


mschwager (approved):
Nice! Gotta go fast... :rocket:

PR #1287, opened 2021-05-17
identical_id: Add more wildcards to match in more places

underyx:
@kurt-r2c edit - it looks like the refactor from `meta/**` to `yaml/semgrep/**` renamed / updated the source rule.  Using @underyx test cases for the new rule.


emjin (approved):

meta/identical_id.yaml
```diff
@@ -7,10 +7,12 @@ rules:
   patterns:
   - pattern-inside: 'rules: [..., $RULE, ...]'
   - pattern-inside: |
+      ...
```
kurt-r2c: leading ellipsis is an anti-pattern we try to avoid - probably okay here?

cc @minusworld

kurt-r2c (approved):

PR #1288, opened 2021-05-19
Remove unnecessary 'patterns' and 'pattern-either' parent operators

mschwager:
Building on https://github.com/returntocorp/semgrep-rules/pull/1278. I wrote a script to do this, and the tests pass, so we should be good to go :+1:


dlukeomalley:
@mschwager I think  contrib/ rules may now not reflect what's upstream. I'm actually wondering @DrewDennison as we think through registry work (specifically `publish`) if we plan to deprecate `contrib/` rules in this repository and favor the rules being published from their source repo with appropriate licenses, descriptions, and authorship attribution.

mschwager:
> @mschwager I think contrib/ rules may now not reflect what's upstream. I'm actually wondering @DrewDennison as we think through registry work (specifically `publish`) if we plan to deprecate `contrib/` rules in this repository and favor the rules being published from their source repo with appropriate licenses, descriptions, and authorship attribution.

Roger that - yeah I think it makes sense to limit our ownership here. Having to keep things up-to-date in multiple places feels like an unnecessary burden.

colleend (approved):
cool! nice :D

PR #1291, opened 2021-05-20
Adding trivial import of Apache NoopHostnameVerifier

Gby56:
I found out that the current rules for Java hostname validations are only covering the deep implementations, but not this trivial Apache bad validator, thought it could help ! :) 
https://www.javadoc.io/doc/org.apache.httpcomponents/httpclient/4.5.3/org/apache/http/conn/ssl/NoopHostnameVerifier.html


minusworld:
@Gby56 Looks cool! Would you be able to add a test to the corresponding test file and annotate it with `// ruleid: insecure-hostname-verifier`? Let me know if you need any assistance. Thanks!

minusworld (approved):

PR #1292, opened 2021-05-20
Update detected-aws-secret-access-key.yaml

josh-bridge:
This should now match words between the keywords 'aws' and 'secret'
e.g. Previously this would not have been matched: `AWS_TEST_SECRET = "..."`

Also as the words 'access', 'key', and 'id' were all optional, they are effectively redundant and only add arbitrary strictness to the rule.

### DIFF
--- a/generic/secrets/security/detected-aws-secret-access-key.yaml
+++ b/generic/secrets/security/detected-aws-secret-access-key.yaml
@@ -1,7 +1,7 @@
 rules:
 - id: detected-aws-secret-access-key
   pattern-regex: |-
-    (("|'|`)?((?i)aws)_?((?i)secret)_?((?i)access)?_?((?i)key)?_?((?i)id)?("|'|`)?\s{0,50}(:|=>|=)\s{0,50}("|'|`)?[A-Za-z0-9/+=]{40}("|'|`)?)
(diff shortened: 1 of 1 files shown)


brendongo:
I think this rule is specifically finding usages of https://docs.aws.amazon.com/sdkref/latest/guide/setting-global-aws_secret_access_key.html not any env variable that has AWS and SECRET in the name?

josh-bridge:
> I think this rule is specifically finding usages of https://docs.aws.amazon.com/sdkref/latest/guide/setting-global-aws_secret_access_key.html not any env variable that has AWS and SECRET in the name?

Well I found it attempting to use https://semgrep.dev/p/secrets while scanning a Java project with a constant containing a secret. I would have thought it is better to catch more instances with AWS and SECRET than to ignore cases which aren't precisely 'aws_secret_access_key = '...' 

Especially given this is supposed to be an automated process and it's very possible to (like I did) not name it aws_secret_access_key

minusworld:
@josh-bridge Looks good! I'd like to run it over a batch of repos and see how the old and new versions compare. I'll report the results back here soon.

minusworld (approved):

minusworld:
I ran this over ~300 test cases and the results were identical!

PR #1304, opened 2021-05-26
Switch to lombok for Java tests

minusworld:
avro had a bunch of subdirectories for each implementation of avro in different languages. This made avro huge and the Java tests disproportionately long.

This switches to [lombok](https://github.com/projectlombok/lombok) which clocks in at about 90,000 sloc and should be a better "medium" repo for tests.


colleend (approved):
cool, lgtm!

PR #1305, opened 2021-05-26
Move C# rules from 'cs' to 'csharp'

mschwager:
From Slack:

> Random thought in retrospect: should we put the C# rules under csharp/ instead of cs/? I advised cs, but after looking at the other language directory names, csharp seems more descriptive

cc @ahmetak4n


colleend (approved):
nice!

PR #1308, opened 2021-05-31
sqlalchemy-execute-raw-query: fix rule to take params into consideration

0xdbe:


python/sqlalchemy/security/sqlalchemy-execute-raw-query.py
```diff
@@ -6,7 +6,7 @@
 engine = create_engine('postgresql://user@localhost/database')
 echo("database connexion: ok")
 # ruleid: sqlalchemy-execute-raw-query
-engine.execute("INSERT INTO person (name) VALUES ('" + name + "')")
+engine.execute("INSERT INTO person (name) VALUES ('" + name + "')", multi=False)
```
mschwager, line 11: Can you make this a completely new `# ruleid: ...` test? That way we have both tests confirming behavior :)

python/sqlalchemy/security/sqlalchemy-execute-raw-query.py
```diff
@@ -193,3 +193,34 @@
 engine = create_engine('postgresql://user@localhost/database')
 query = select(literal_column("users.fullname", String) + ', ' + literal_column("addresses.email_address").label("title")).where(and_(literal_column("users.id") == literal_column("addresses.user_id"), text("users.name BETWEEN 'm' AND 'z'"), text("(addresses.email_address LIKE :x OR addresses.email_address LIKE :y)"))).select_from(table('users')).select_from(table('addresses'))
 conn.execute(query, {"x":"%@aol.com", "y":"%@msn.com"}).fetchall()
+
+
+# SQL Composition using SQL Expression
+# ok: sqlalchemy-execute-raw-query
```
mschwager: Unless I'm missing something, I think you want these `ok` annotations right above the `execute` line :+1:

python/sqlalchemy/security/sqlalchemy-execute-raw-query.py
```diff
@@ -193,3 +193,34 @@
...
+  meta.reflect(bind=connection)
+  product_table = meta.tables['product']
+  stmt = (
+    select(product_table)
+    .where(product_table.columns[field_name] == value_name)
+  )
+  result = connection.execute(stmt)
+
+# Insert multi data record using SQL Expression
+# ok: sqlalchemy-execute-raw-query
```
mschwager: Same as above.

python/sqlalchemy/security/sqlalchemy-execute-raw-query.yaml
```diff
@@ -7,7 +7,8 @@ rules:
        safely, prepared statement should be used.
        SQLAlchemy provides TextualSQL to easily used prepared statement with
        named parameters. For complexe SQL composition, use SQL Expression
-       Langague or SQL Statements and Expressions API.
+       Langague or Schema Definition Language. In most case, SQLAlchemy ORM will
```
mschwager:
```suggestion
       Language or Schema Definition Language. In most cases, SQLAlchemy ORM will
```

mschwager (commented):
Nice catch! Thank you!

python/sqlalchemy/security/sqlalchemy-execute-raw-query.py
0xdbe: I have to check if concatenation is use before execution:

```
# Insert multi data record using SQL Expression
connection_string = 'sqlite:///db.sqlite'
engine = create_engine(connection_string, echo=True)
with engine.connect() as connection:
  meta = MetaData()
  meta.reflect(bind=connection)
  product_table = meta.tables['product']
  # ruleid: sqlalchemy-execute-raw-query
  stmt = insert(product_table) + 'test'
  values = [
      {field_name: 'hazelnut', field_price: 5},
      {field_name: 'banana', field_price: 8}
  ]
  connection.execute(stmt, values)`
```

`ok` annotations should be just before `stmt` line ?

PR #1309, opened 2021-06-01
Add rule to check for AWS EC2 IMDSv1 

x64-latacora:


x64-latacora:
@clintgibler :eyes:

x64-latacora:
@clintgibler :eyes:

terraform/lang/security/ec2-imdsv1-optional.tf
```diff
@@ -0,0 +1,8 @@
+# ruleid: ec2-imdsv1-optional
+resource "aws_instance" "test-instance" {
+  ami = "ami-0d5eff06f840b45e9"
+  metadata_options {
+    http_endpoint = "enabled"
+    http_tokens = "optional"
```
mschwager, line 6: I think you'll need to move the `# ruleid: ...` comment above this line.

Also, can you add a true negative test as well? Similarly you can use the [`# ok: ...` syntax](https://semgrep.dev/docs/writing-rules/testing-rules/).

terraform/lang/security/ec2-imdsv1-optional.yaml
```diff
@@ -0,0 +1,18 @@
...
+      AWS EC2 Instance allowing use of the IMDSv1
+    metadata:
+      references:
+        - https://aws.amazon.com/blogs/security/defense-in-depth-open-firewalls-reverse-proxies-ssrf-vulnerabilities-ec2-instance-metadata-service
+        - https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/instance#metadata-options
+    paths:
+      include:
+        - '*.tf'
+    patterns:
+      - pattern-inside: metadata_options ...
```
mschwager: I think the following may make more sense:

```
- pattern-inside: |
    metadata_options {
        ...
        ...
        ...
    }
```

x64-latacora: Improved the constraint.

mschwager:
`pre-commit` is being extra finicky! (Especially with this new Github CI approval workflow :/)

After adding in the trailing new lines (but not an extra line at the end) we should be good to go!

x64-latacora:
> After adding in the trailing new lines (but not an extra line at the end) we should be good to go!

:crossed_fingers:

mschwager (approved):
Nice! Thanks again!

Issue #1307, opened 2021-05-28
Add `category` and `framework` metadata fields to all rules


minusworld:
**Is your feature request related to a problem? Please describe.**
Instead of relying on the directory path for these, these attributes should be put in the metadata.

**Describe the solution you'd like**
Something like this:

```
metadata:
  owasp: ...
  category: security
  framework: django
```

AND probably a metalinter rule which enforces an enum on `category`, one of:
- security
- portability
- maintainability
- correctness
- best-practice
- performance

**Additional context**
This will get used in semgrep.dev's new registry search.

underyx:
Will we tag kubernetes/requests/attrs as frameworks too? I'd recommend `tech` instead, or `technology`

minusworld:
tbd; @clintgibler and @dlukeomalley and @chmccreery  for more thoughts

chmccreery:
Do we allow rules to apply to more than one category? More than one framework? I think there are some rules right now that pertain to multiple frameworks according to their descriptions... e.g. https://semgrep.dev/r?q=flask+django

Also there are rules that have no technology, right? How will this be documented - as `framework: none` or the absence of a `framework` key in the metadata? I think either is fine, it just needs to be standardized

dlukeomalley:
> Will we tag kubernetes/requests/attrs as frameworks too? I'd recommend `tech` instead, or `technology`

+1 to choosing the more generic "technology" label over framework. I think technology will be used quite broadly for the time being and better matches the intention of the label from my perspective (at least for how we intend to use it in the UI).

As you make progress on this work @minusworld I'll be curious to see if we find rules have multiple technologies (per @chmccreery question) and/or multiple categories. No strong opinions here - I see benefit to enforcing choosing a single category/framework (and splitting up rules on the basis of these things) or allow multiple values.

dlukeomalley:
For the `owasp` field, what do you think of moving towards only including the category identifier and not it's description? For example I quickly found a rule with `owasp: 'A1: Some Vulnerability'`, for which I could imagine just using `owasp: a1-2017` or similar and letting the UI, CLI, or CI runner resolving to a human readable string when printing it out. This could reduce the likelihood of typos and formatting inconsistency? Not strongly opinionated, just thinking through areas of concern and author frustration down the road.

nbrahms:
> Yayy!!!! Thank you for doing this 🙏
> 
[...]

PR #1316, opened 2021-06-03
Enrich rules with category and technology metadata

inkz:
for: https://github.com/returntocorp/semgrep-rules/issues/1307


inkz:
I updated  most of the rules - those that were easy to automate with the script, after confirming that the current format works for us I will update the rest of the rules.

discussion about the format here: https://github.com/returntocorp/semgrep-rules/issues/1307#issuecomment-853741516

generic/secrets/security/detected-amazon-mws-auth-token.yaml
```diff
@@ -6,3 +6,6 @@ rules:
   severity: ERROR
   metadata:
     source-rule-url: https://github.com/dxa4481/truffleHogRegexes/blob/master/truffleHogRegexes/regexes.json
+    category: security
+    technology:
+    - secrets
```
dlukeomalley, line 11: Interesting edge-case, is `secrets` a technology or some other class of issue within the `security` category? Is there a technology this rule checks secrets for (i.e. `aws`)?

go/jwt-go/security/jwt-none-alg.yaml
```diff
@@ -10,6 +10,9 @@ rules:
     cwe: 'CWE-327: Use of a Broken or Risky Cryptographic Algorithm'
     owasp: 'A2: Broken Authentication'
     source-rule-url: https://r2c.dev/blog/2020/hardcoded-secrets-unverified-tokens-and-other-common-jwt-mistakes/
+    category: security
+    technology:
+    - jwt-go
```
dlukeomalley, line 15: For go libraries, is this the full identifier for the library or would it be `dgrijalva/jwt-go`? I'm not sure the go idiom or what the import would specifically look like.

go/lang/best-practice/channel-guarded-with-mutex.yaml
```diff
@@ -16,3 +16,7 @@ rules:
     for more information.
   languages: [go]
   severity: WARNING
+  metadata:
+    category: best-practice
+    technology:
+    - otto
```
dlukeomalley, line 22: Same as above, for go libraries would we use their full "path" like `robertkrimen/otto`?

java/java-jwt/security/audit/jwt-decode-without-verify.yaml
```diff
@@ -9,6 +9,9 @@ rules:
     cwe: 'CWE-345: Insufficient Verification of Data Authenticity'
     owasp: 'A2: Broken Authentication'
     source-rule-url: https://r2c.dev/blog/2020/hardcoded-secrets-unverified-tokens-and-other-common-jwt-mistakes/
+    category: security
+    technology:
+    - java-jwt
```
dlukeomalley, line 14: Would this be `auth0.java-jwt` or the library identifier?

chmccreery (approved):

generic/secrets/security/detected-amazon-mws-auth-token.yaml
chmccreery: related question - is there a list of 'supported technologies' somewhere that is being used to assign these metadata labels, or are we always taking the second item in the filepath? I imagine that the latter approach will be noisy and lead to more technology categories than we actually want (ran into this with registry filters before)

java/java-jwt/security/audit/jwt-decode-without-verify.yaml
chmccreery: I actually wonder if technology here should just be `jwt` since `java` is already captured by the language information. Or is `java-jwt` strictly different from `jwt`?

generic/secrets/security/detected-amazon-mws-auth-token.yaml
inkz: oooh thats a very good call, I think we should have it as a category

dlukeomalley: I'd like to limit ourselves to the 7 categories we have so far (I think this is truly within the `security` category), but wonder if beyond secrets there is a technology this check specifically applies to.

Perhaps we do need auxiliary labels to capture stuff like `secret`, or maybe the rule-id could include `secret` in it.

java/java-jwt/security/audit/jwt-decode-without-verify.yaml
dlukeomalley: @chmccreery I think this is applicable to a specific jwt library, `auth0.java-jwt`, so only applicable in the case you're using that, and then more broadly it's also `jwt` as the technology.

PR #1317, opened 2021-06-04
add session support to the regexes as well

mlissner:
It's really common, I think, to write code like this:

```
session = requests.Session()
r = session.get(some_url)
```

Hopefully this will catch that as well without catching things it shouldn't. I also do this:

```
s = requests.Session()
s.get(some_url)
```

But I felt like that would probably catch things we didn't want, so I left that out for now.

### DIFF
--- a/python/requests/best-practice/use-timeout.yaml
+++ b/python/requests/best-practice/use-timeout.yaml
@@ -3,6 +3,8 @@ rules:
   patterns:
   - pattern-not: requests.$W(..., timeout=$N, ...)
   - pattern-not: requests.$W(..., **$KWARGS)
+  - pattern-not: session.$W(..., **$KWARGS)
+  - pattern-not: session.$W(..., **$KWARGS)  
   - pattern-either:
     - pattern: requests.request(...)
     - pattern: requests.get(...)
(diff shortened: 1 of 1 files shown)


mschwager:
Hi there @mlissner,

Thanks for catching this! As it's currently implemented, I believe this will only fire if the variable name is `session`. We can remedy this with a [`metavariable`](https://semgrep.dev/docs/writing-rules/pattern-syntax/#metavariables), e.g.

```yaml
- pattern: |
    $SESSION = requests.Session(...)
    ...
    $SESSION.get(...)
```

And the corresponding `pattern-not`:

```yaml
- pattern-not: |
    $SESSION = requests.Session(...)
    ...
    $SESSION.get(..., timeout=$N, ....)
- pattern-not: |
    $SESSION = requests.Session(...)
    ...
    $SESSION.get(..., **$KWARGS)
```

Also, can you add some tests for the corresponding change in behavior?

mlissner:
Sorry, I have a backlog of open source contributions. I would love to fix this, but I just know I won't get back around to it if I have to figure out the tests and syntax. At least not for a very long time. If somebody wants to run with it, that'd be great, but unfortunately I've taken this about as far as I can for now. I am loving semgrep though, and hope to have more time to work on it someday. (I have PRs for Django I gotta get across the line this week that I understand better.)

inkz:
resolved by https://github.com/returntocorp/semgrep-rules/pull/1338

Issue #1282, opened 2021-05-13
False positive in generic.nginx.security.missing-internal.missing-internal


sk-:
**Describe the bug**
The patterns for `generic.nginx.security.missing-internal.missing-internal` match any configuration using `proxy_pass`, even it it doesn't allow SSRF.

**To Reproduce**
```
location / {
        proxy_pass http://127.0.0.1:8000/;
}
```

**Expected behavior**
This should not raise an error.

**Priority**
How important is this to you?
- [ ] P0: blocking me from making progress
- [X] P1: this will block me in the near future
- [ ] P2: annoying but not blocking me

**Additional Context**
Based on the linked article https://github.com/yandex/gixy/blob/master/docs/en/plugins/ssrf.md it seems that there may be a vulnerability in case the proxy_pass location is not constant.

inkz:
Thanks a lot for noticing this @sk- !
link to the commit: https://github.com/returntocorp/semgrep-rules/pull/1320

PR #1320, opened 2021-06-07
update generic/nginx/security/missing-internal rule

inkz:
updates for: https://github.com/returntocorp/semgrep-rules/issues/1282

### DIFF
--- a/generic/nginx/security/missing-internal.yaml
+++ b/generic/nginx/security/missing-internal.yaml
@@ -12,7 +12,7 @@ rules:
         internal;
         ...
       }
-  - pattern: proxy_pass ...;
+  - pattern: proxy_pass ...$...;
   paths:
     include:
     - '*.conf'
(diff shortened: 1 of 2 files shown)


generic/nginx/security/missing-internal.yaml
```diff
@@ -23,7 +23,7 @@ rules:
   - generic
   severity: WARNING
   message: >-
-    This location block contains a 'proxy_pass' directive but does not contain the 'internal'
```
mschwager: I don't see `$URL` defined in any of the patterns above. Did you mean to include this?

mschwager (approved):

generic/nginx/security/missing-internal.yaml
inkz: nope, this is "debugging" information :laughing: 
thanks!

PR #1322, opened 2021-06-08
Add rule for the GitHub tokens

samm-git:
This PR adds support for the new Github token format which is described in the [github blog](https://github.blog/changelog/2021-03-04-authentication-token-format-updates/) article. Tested locally with a token generated (and revoked) in the GH UI.


inkz (approved):
Thanks a lot for your contribution @samm-git ! :+1:

Issue #1277, opened 2021-05-10
SQL injection false negative


emjin:
**Describe the bug**
See https://github.com/returntocorp/semgrep/issues/3065

**Steps to reproduce**
https://semgrep.dev/s/enochii:false-negative-of-sql-injection

The fix seems very easy: the original reporter proposed
![image](https://user-images.githubusercontent.com/5942585/117672897-45319000-b178-11eb-9356-37318c3bac72.png)

To avoid false positives, we might prefer 
`- pattern: $OBJ.Exec("..." + $X + ...)`

Perhaps this is something we need to change for the subsequent patterns as well, however.

**What is the priority of the bug to you?**

P2(regular bug that should get fixed)

enochii:
Hi! So to fix this issue, the changes should be done in original [repo](https://github.com/returntocorp/semgrep/)? Seems like I don't find the corresponding `yaml` file here.

inkz:
Hi @enochii , the rule that need to be fixed is located in this repository (https://github.com/returntocorp/semgrep-rules/)
here is the file: https://github.com/returntocorp/semgrep-rules/blob/develop/go/lang/security/audit/database/string-formatted-query.yaml
:pray: sorry for the long response.

PR #1325, opened 2021-06-08
fix false negative in rule

colleend:
For https://github.com/returntocorp/semgrep-rules/issues/1277


inkz (commented):
1. I think that not only `Exec` call should be updated
2. after further investigation, I figured out that by using `"..." + $X` we (obviously :smile:) highlight only code patterns that start with a string, which is I guess the most common scenario but I tried to make the rule broader, without introducing new FPs

please take a look :v:

inkz (approved):

PR #1329, opened 2021-06-09
Include GCM rule for hardcoded IV/nonce

securingapps:
Reusing the same IV/nonce with GCM can make encryption totally useless with the knowledge of one (cleartext,ciphertext) pair (cf https://www.youtube.com/watch?v=r1awgAl90wM)

This is a first basic rule: it only detects hardcoded IVs built from strings.
I have not found how to detect harcoded byte arrays

There is also an INFO rule for manual review each time GCM is detected.

NB: Those rules could be defined for other languages than Java.

### DIFF
--- a/java/lang/security/audit/crypto/gcm-hardcoded-iv.java
+++ b/java/lang/security/audit/crypto/gcm-hardcoded-iv.java
@@ -64,5 +64,5 @@ public class GcmHardcodedIV
 
         theIV = BAD_IV.getBytes();
     }
-    
+
 }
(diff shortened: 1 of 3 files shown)


inkz:
Thank you for your contribution @securingapps !

could you please format your rule with `pre-commit` hook https://github.com/returntocorp/semgrep-rules/blob/develop/CONTRIBUTING.md#pre-commit
and push changes :pray:

inkz (changes requested):
`pre-commit` formatting missing

securingapps:
@inkz Thank you for the feedback.
Formatting and tests should be OK now

inkz:
@securingapps yeah I see, trying to understand if it is possible to highlight code with hardcoded byte arrays also

inkz:
@securingapps I resubmitted the pull request with a few updates (split the rules into two files) https://github.com/returntocorp/semgrep-rules/pull/1334
and also attached blocking issue which will allow finding hardcoded byte arrays, will update the rule ones the issue is solved
thanks again for your contribution :v:

PR #1336, opened 2021-06-11
fix: updates javascript.express.security.audit.xss.mustache.explicit-unescape to better support non-mustache templates

sonjaq:
Solution:

Updates Regex to avoid certain angular template patterns, and adds test element to validate 

Addresses #1335 returntocorp/semgrep/issues/3344


inkz (approved):
lgtm, thank you @sonjaq! :+1:

PR #1340, opened 2021-06-14
Cleanup OWASP metadata strings

mschwager:
For https://github.com/returntocorp/enterprise/issues/668.

Makes this query more useful:

```
$ rg -g '*.yaml' -g '!*.test.yaml' -I -o 'owasp: .*' | sort | uniq -c
      1 owasp: ...'
      1 owasp: "=~/^A([0-9]|10): .+$/"'
    181 owasp: 'A1: Injection'
     53 owasp: 'A2: Broken Authentication'
    133 owasp: 'A3: Sensitive Data Exposure'
     23 owasp: 'A4: XML External Entities (XXE)'
      9 owasp: 'A5: Broken Access Control'
     46 owasp: 'A6: Security Misconfiguration'
     91 owasp: 'A7: Cross-Site Scripting (XSS)'
     31 owasp: 'A8: Insecure Deserialization'
     19 owasp: 'A9: Using Components with Known Vulnerabilities'
```

### DIFF
--- a/contrib/nodejsscan/xss_mustache_escape.yaml
+++ b/contrib/nodejsscan/xss_mustache_escape.yaml
@@ -6,7 +6,7 @@ rules:
   - javascript
   metadata:
     cwe: 'CWE-116: Improper Encoding or Escaping of Output'
-    owasp: 'A7: Cross-Site Scripting XSS'
+    owasp: 'A7: Cross-Site Scripting (XSS)'
     category: security
     technology:
     - node.js
--- a/generic/html-templates/security/unquoted-attribute-var.yaml
+++ b/generic/html-templates/security/unquoted-attribute-var.yaml
@@ -6,7 +6,7 @@ rules:
     add quotes around the template expression, like this: "{{ expr }}".
   metadata:
     cwe: "CWE-79: Improper Neutralization of Input During Web Page Generation ('Cross-site Scripting')"
-    owasp: 'A7: Cross-site Scripting (XSS)'
+    owasp: 'A7: Cross-Site Scripting (XSS)'
     references:
     - https://flask.palletsprojects.com/en/1.1.x/security/#cross-site-scripting-xss
     category: security
--- a/generic/html-templates/security/var-in-href.yaml
+++ b/generic/html-templates/security/var-in-href.yaml
@@ -12,7 +12,7 @@ rules:
     You may also consider setting the Content Security Policy (CSP) header.
   metadata:
     cwe: "CWE-79: Improper Neutralization of Input During Web Page Generation ('Cross-site Scripting')"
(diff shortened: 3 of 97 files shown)


mschwager:
For reference: https://owasp.org/www-project-top-ten/

dlukeomalley:
@mschwager Epic cleanup. Tech question for you - should we have the full OWASP label here or choose just something like `a1` and have Semgrep CLI + Registry UI + other be aware that `a1` has a nice human readable name? I worry we're prone to typos and variations of the human readable name across rule authors.

mschwager:
> ... should we have the full OWASP label here or choose just something like a1 and have Semgrep CLI + Registry UI + other be aware that a1 has a nice human readable name?

I don't have strong feelings here. I see the following trade-offs:

- Short labels
  - Pros: easier for other systems to parse this information, less typos, etc.
  - Cons: like short rule IDs, requires a lookup if you're looking at the rule. I.e. "What is A4 again...?"
- Long labels
  - Pros: full description is available in the rule text.
  - Cons: opposite of the pros in short labels.

I have a slight preference for long labels. I think we can mitigate the cons with the following:

1) To avoid typos and such in other systems we could do something like the following when an `owasp` key is present: `owasp_value.lower().startswith('a1')`, or `owasp_value.lower().startswith('a2')`, etc. This would simplify rollups on potentially "ununiform" data.
2) We could tighten the bounds on our [internal `owasp` rule](https://github.com/returntocorp/semgrep-rules/blob/develop/yaml/semgrep/metadata-owasp.yaml). E.g. something like:

```
pattern-not: 'owasp: "=~/^(A1: Injection|A2: Broken Authentication|A3: Sensitive Data Exposure|A4: XML External Entities (XXE)|A5: Broken Access Control|A6: Security Misconfiguration|A7: Cross-Sie Scripting (XSS)|A7: Cross-Site Scripting (XSS)|A8: Insecure Deserialization|A9: Using Components with Known Vulnerabilities')$/"'
```

This would help avoid typos by checking for exact matches. This could frustrate contributors in the `pre-commit` stage, but it would help enforce a consistent format :man_shrugging: 

Thoughts?

inkz (approved):
lgtm

inkz:
I agree with @mschwager !
+1 for adding owasp check to `pre-commit`

mschwager:
Let's go with this approach for now. We can easily switch to short labels in the future, and this makes that even easier :+1:

dlukeomalley:
Supportive of the bias towards action here. Proposal for future: `a1-2017`. This gives us a stable ID (the ranking changes every few years, so need to "timestamp" it) and I'd like the product pressure to pretty display OWASP categories.

PR #1342, opened 2021-06-17
Update autofix example rule

chmccreery:
this rule is our example autofix rule in playground, but did not have an autofix in the registry version. This unifies the two rules.


python/lang/correctness/exit.yaml
```diff
@@ -4,9 +4,8 @@ rules:
   - python
   message: Use `sys.exit` over the python shell `exit` built-in. `exit` is a helper for the interactive shell and may not
     be available on all Python implementations. https://stackoverflow.com/questions/6501121/difference-between-exit-and-sys-exit-in-python
-  patterns:
-  - pattern: exit(...)
-  - pattern-not: sys.exit(...)
+  pattern: exit($X)
```
emjin: Test is failing because of this. We still need to explicitly not match `sys.exit(...)` because of the `import sys; exit()` pattern.

emjin (changes requested):
This PR should be even smaller :) Just add the fix.

python/lang/correctness/exit.yaml
chmccreery: 👍

emjin (approved):

emjin:
I don't know what this PR failure is, seems unrelated

PR #1343, opened 2021-06-17
Dockerfile ignore line continuation

ben-elttam:
Fix false positives for Dockerfiles with line continuation after the `&&`.
Added more unit tests.
Fixed for `yum`, `dnf`, and `zypper`


minusworld (approved):
Thanks!!

PR #1345, opened 2021-06-21
Fix double-slash typo in test file source URL

dbolkensteyn:


minusworld (approved):
Thanks @dbolkensteyn !

PR #1346, opened 2021-06-21
add dangerous-asyncio-shell rule

inkz:
OS command injection through asyncio EventLoop

### DIFF
--- a/python/lang/security/audit/dangerous-asyncio-shell.yaml
+++ b/python/lang/security/audit/dangerous-asyncio-shell.yaml
@@ -28,4 +28,6 @@ rules:
     - https://docs.python.org/3/library/asyncio-subprocess.html
     - https://docs.python.org/3/library/shlex.html
     category: security
+  languages:
+  - python
   severity: WARNING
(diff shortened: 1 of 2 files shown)


python/lang/security/audit/dangerous-asyncio-shell.yaml
```diff
@@ -0,0 +1,33 @@
+rules:
+- id: dangerous-asyncio-shell
+  patterns:
+  - pattern-either:
+    - pattern: $LOOP.subprocess_shell($PROTOCOL, $CMD)
+    - pattern: asyncio.subprocess.create_subprocess_shell($CMD, ...)
+    - pattern: asyncio.create_subprocess_shell($CMD, ...)
+  - pattern-not-inside: |
```
minusworld, line 8: Is this still necessary with constant propagation? Either way, I'll approve

minusworld (approved):

python/lang/security/audit/dangerous-asyncio-shell.yaml
inkz: very good call:+1: 
I was going to rely on constant propagation, but my test cases did not work for some reason, I decided to add `pattern-not-inside`, but after your question I understand that constant propagation does not work for code inside/outside of:
```python
hardcoded = "blabla"
with something() as foobar:
   print(hardcoded) # not going to be tracked by constant propagation
```
worth opening an issue, thanks! :v:

PR #1348, opened 2021-06-22
dangerous asyncio subprocess calls

inkz:
2 more rules for dangerous asyncio subprocess calls


mschwager (approved):
Nice, I hadn't heard of `create_subprocess_exec` before :open_mouth:

PR #1350, opened 2021-06-23
Remove 'nodejsscan' contrib rules

### LINKED ISSUE #1347: Remove `nodejsscan` from contrib
The rules at `contrib/nodejsscan` are out-of-date and out-of-sync.

These are kept in-sync with the source repository at https://github.com/ajinabraham/njsscan/tree/master/njsscan/rules/semantic_grep via the endpoints `https://semgrep.dev/r/ajinabraham` or `https://semgrep.dev/p/njsscan`.

mschwager:
Fixes #1347.


minusworld:
Thanks @mschwager! One thing we'll need to do is make sure we patch any rulesets that are running these rules from `contrib.`. I think there are a handful.

mschwager:
> One thing we'll need to do is make sure we patch any rulesets that are running these rules from contrib.. I think there are a handful.

Ahh, good catch. Looks like there are actually quite a few:

```
$ rg 'contrib.nodejsscan' | wc -l
1039
```

We should probably do that before merging this.

colleend:
@minusworld should we merge this?

PR #1354, opened 2021-06-24
add dangerous-annotations-usage rule

inkz:
```python
class C:
    member: int = 0

C.__annotations__["member"] = "print('Hello world')"
get_type_hints(C) # 'Hello world' will be printed to console
```
annotations are evaluated when passed to `get_type_hints` (https://docs.python.org/3/library/typing.html#typing.get_type_hints)
tracking every `get_type_hints` call may end up in lots of FPs so I created a rule for highlighting arbitrary write to `__annotations__`

### DIFF
--- a/python/lang/security/audit/dangerous-annotations-usage.py
+++ b/python/lang/security/audit/dangerous-annotations-usage.py
@@ -1,0 +1,19 @@
+from typing import List, Set, Dict, Tuple, Optional, get_type_hints
+
+class C:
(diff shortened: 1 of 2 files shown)


minusworld (approved):

mschwager:
Huh, this is an interesting one I hadn't heard of before :thinking:

minusworld:
Pretty neat, huh?

PR #1355, opened 2021-06-24
Add a missing 'category' meta-linter rule

minusworld:


yaml/semgrep/missing-category.yaml
```diff
@@ -0,0 +1,20 @@
...
+  message: >-
+    This Semgrep rule is missing a valid 'category' field in the 'metadata'.
+    'category' must be one of 'security', 'correctness', 'best-practice',
+    'performance', 'maintainability', or 'portability'.
+  severity: INFO
+  patterns:
+  - pattern-inside: 'rules: [..., $RULE, ...]'
+  - pattern: $RULE
+  - pattern-not: |
+      ...
```
mschwager: Do we need this `...`? I know hanging `...` at the beginning or end tend to slow things down.

yaml/semgrep/missing-category.yaml
```diff
@@ -0,0 +1,20 @@
+rules:
+- id: missing-category
```
mschwager: Should we name this `metadata-category` to be consistent with other rules here?

mschwager (approved):
LGTM - I left a few nits for consideration

yaml/semgrep/missing-category.yaml
minusworld: Good question, I didn't test that! brb

minusworld: Not needed, good catch 👍

PR #1358, opened 2021-06-27
ReDos vulnerability patterns were added

ahmetak4n:


minusworld (approved):
Excellent! 👏  Thank you for the tests especially!

minusworld:
You should be able to scan with this rule now using: `semgrep -f r/csharp.lang.security`

PR #1360, opened 2021-06-29
Speed up ExpressJS rules

mschwager:
For https://github.com/returntocorp/semgrep-app/issues/2428.

The primary optimization here is that the following are equivalent, so unnecessary patterns can be removed:

```
- pattern-inside: function ($REQ, $RES) {...}
- pattern-inside: function $FUNC($REQ, $RES) {...}
- pattern-inside: function ($REQ, $RES, $NEXT) {...}
- pattern-inside: function $FUNC($REQ, $RES, $NEXT) {...}
- pattern-inside: $X = function $FUNC($REQ, $RES) {...}
- pattern-inside: $X = function $FUNC($REQ, $RES, $NEXT) {...}
```

```
- pattern-inside: function ... ($REQ, $RES) {...}
- pattern-inside: function ... ($REQ, $RES, $NEXT) {...}
```

Removing unnecessary patterns reduces the number of source code queries we need to make, which reduces runtime, and reduces the amount of state we need to keep in memory.

There were also a few trailing, unnecessary `...`'s I removed.

### DIFF
--- a/javascript/express/security/audit/express-path-join-resolve-traversal.js
+++ b/javascript/express/security/audit/express-path-join-resolve-traversal.js
@@ -24,6 +24,22 @@ function testCtrl3(req,res) {
     res.send('Hello World!');
 }
 
+const func4 = function testCtrl4(req,res) {
+    // ruleid:express-path-join-resolve-traversal
+    let somePath = req.body.path;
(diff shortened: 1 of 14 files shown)


minusworld (approved):
That looks really cool. 👍

PR #1369, opened 2021-07-12
Add true negative test cases for Semgrep metalints. Fix rules that failed these cases

minusworld:


yaml/semgrep/metadata-category.yaml
```diff
@@ -7,11 +7,12 @@ rules:
   severity: INFO
   patterns:
   - pattern-inside: 'rules: [..., $RULE, ...]'
-  - pattern: $RULE
-  - pattern-not: |
-      metadata:
-        ...
-        category: $CATEGORY
+  - pattern: 'id: $RULEID'
```
mschwager, line 10: In the Kubernetes rules I expressed this type of query like this:

```yaml
  patterns:
  - pattern-inside: 'rules: [..., $RULE, ...]'
  - pattern: |
      id: ...
      ...
  - pattern-not: |
      id: ...
      ...
      metadata:
        ...
        category: ...
```

I haven't empirically tested this, but I suspect a trailing `...` is slightly faster and more memory efficient than a leading `...`. A trailing one will only activate once the initial code is found whereas a leading one will activate on everything until it _may_ find the corresponding code. Not a huge deal, but may be slightly more performant.

mschwager (approved):

Issue #1365, opened 2021-07-02
Update rules to move away from features only available with `--optimizations none`


brendongo:
**Is your feature request related to a problem? Please describe.**
A lot of the recent improvements in semgrep performance are only available when
it can run with optimizations but there is some functionality unavailable and when those
functionality is detected, semgrep falls back to running without optimizations.

This includes the following:
- pattern-where-python
- equivalences
- taint mode

We should update our rules to be able to run with optimizations as much as possible.

- There are four rules running pattern-where-python
    - https://github.com/returntocorp/semgrep-rules/blob/develop/contrib/dlint/redos.yaml
    - https://github.com/returntocorp/semgrep-rules/blob/develop/go/gorilla/security/audit/handler-attribute-read-from-multiple-sources.yaml
    - https://github.com/returntocorp/semgrep-rules/blob/develop/python/flask/experimental/correctness/different-route-names.yaml
    - https://github.com/returntocorp/semgrep-rules/blob/develop/python/lang/security/audit/insecure-file-permissions.yaml
- There is the one rule running equivalences
    - https://github.com/returntocorp/semgrep-rules/blob/develop/java/lang/correctness/no-string-eqeq.yaml 
- There are no rules running taint mode

The equivalence rule seems easy enough to rewrite but not the pattern-where-python rules. Should we be implementing some scheme to support something similar with optimizations all or can we go without these rules?

# When is work done

- [ ] When no r2c official rulesets contain rules that need experimental features listed above

mschwager:
Perhaps we should remove `equivalences` instead: https://github.com/returntocorp/semgrep/issues/3049

brendongo:
Yeah definitely going to remove equivalences from semgrep this ticket is to remove it's use in our rules.

brendongo:
Relevant https://github.com/returntocorp/semgrep/pull/3493

inkz:
@minusworld @mschwager @brendongo 
Updated the rule with `equivalences`

speaking of `pattern-where-python` rules, I was not able to convert most of them,

https://github.com/returntocorp/semgrep-rules/blob/develop/python/lang/security/audit/insecure-file-permissions.yaml
this one  looks resolvable, just need more time on it

https://github.com/returntocorp/semgrep-rules/blob/develop/contrib/dlint/redos.yaml
Redos rule is just basically one big python script placed in a semgrep YAML file, so I am pretty sure we not gonna be able to somehow convert it

[...]

PR #1370, opened 2021-07-13
Update no-string-eqeq.yaml

inkz:
remove `equivalences` while keeping the rule behavior the same

for: https://github.com/returntocorp/semgrep-rules/issues/1365


colleend (approved):
nice!

PR #1381, opened 2021-07-19
update remove-package-lists rule

inkz:
for: https://github.com/returntocorp/semgrep-rules/issues/1380


inkz:
@minusworld 
smth weird with tests, updated rule works fine, but `react-props-in-state` is failing

minusworld (approved):

PR #1386, opened 2021-07-22
Split raw-html-concat into 2 rules, one for concat ops and one for .j…

minusworld:
…oin(). Uses new metavariable-pattern feature to guess whether the constructed string is HTML.


inkz (approved):
Very cool!

PR #1388, opened 2021-07-22
Remove dev env from update workflow

minusworld:
Seems to be failing a lot lately; seeing if this helps

### DIFF
--- a/.github/workflows/update-semgrep-dev.yml
+++ b/.github/workflows/update-semgrep-dev.yml
@@ -10,8 +10,6 @@ jobs:
     name: Update semgrep.dev
     runs-on: ubuntu-latest
     steps:
-    - name: update dev.semgrep.dev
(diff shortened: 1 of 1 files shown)


.github/workflows/update-semgrep-dev.yml
```diff
@@ -10,8 +10,6 @@ jobs:
     name: Update semgrep.dev
     runs-on: ubuntu-latest
     steps:
-    - name: update dev.semgrep.dev
-      run: curl --fail -X POST -L https://dev.semgrep.dev/api/admin/update-registry
```
adamberman, line 14: i know why this is failing but having trouble getting it over the line. the problem is that the database has rules with no rule_versions. so i added a migration to fix that, and then also added a foreign key to make sure that doesn't happen again. that's making a ton of tests fail. i'm not going to have time to debug this week, but if someone has time to fix the tests in https://github.com/returntocorp/semgrep-app/pull/2987, this will fix the failures

PR #1389, opened 2021-07-23
Switch to taint mode for Django open redirect mode

minusworld:


inkz (approved):
Nice!

aryx:
wow, so much red in the PR :) So much shorter!

IagoAbal:
Great! But the new taint-mode rule doesn't find anything in Django and it should... I'm investigating why, when I fix it I will let you know.

python/django/security/injection/open-redirect.yaml
```diff
@@ -15,517 +16,37 @@ rules:
...
+          if <... django.utils.http.is_safe_url(...) ...>:
+            ...
+    - pattern-not-inside: |
+        def $FUNC(...):
+          ...
+          if <... django.utils.http.url_has_allowed_host_and_scheme(...) ...>:
+            ...
+    - pattern-not: request.get_full_path(...)
+    - pattern-either:
+      - pattern: request.$W.get(...)
```
IagoAbal, line 46: You forgot:
```
- pattern: request.$W
```

IagoAbal: Actually if we add `request.$W` as in the search-mode rule, I guess we don't need `request.$W.get(...)`.

python/django/security/injection/open-redirect.yaml
```diff
@@ -15,517 +16,37 @@ rules:
...
+        def $FUNC(...):
+          ...
+          if <... django.utils.http.is_safe_url(...) ...>:
+            ...
+    - pattern-not-inside: |
+        def $FUNC(...):
+          ...
+          if <... django.utils.http.url_has_allowed_host_and_scheme(...) ...>:
+            ...
+    - pattern-not: request.get_full_path(...)
```
IagoAbal, line 44: If we add `- pattern: request.$W` which is needed to catch some bugs in Django for example, then you need to write this as:
```
    - patterns:
      - pattern-either:
        - pattern: request.$W
        - pattern: request.$W.get(...)
        - pattern: request.$W(...)
        - pattern: request.$W[...]
      - metavariable-regex:
          metavariable: $W
          regex: (?!get_full_path)
```

IagoAbal:
I fixed some issues in taint-mode that should help here. We also know why taint-mode is slower and we will fix. This is the formula that I wrote for this rule:
```
  pattern-sources:
  - patterns:
    - pattern-inside: |
        def $FUNC(...):
          ...
    - pattern-not-inside: |
        def $FUNC(...):
          ...
          <... django.utils.http.is_safe_url(...) ...>
          ...
    - pattern-not-inside: |
        def $FUNC(...):
          ...
          <... django.utils.http.url_has_allowed_host_and_scheme(...) ...>
          ...
    - pattern-not-inside: |
        def $FUNC(...):
          ...
          if <... django.utils.http.is_safe_url(...) ...>:
            ...
    - pattern-not-inside: |
        def $FUNC(...):
          ...
          if <... django.utils.http.url_has_allowed_host_and_scheme(...) ...>:
            ...
    - pattern-not: request.get_full_path
    - pattern: request.$W
  pattern-sinks:
  - pattern-either:
    - pattern: django.shortcuts.redirect(...)
    - pattern: django.http.HttpResponseRedirect(...)
```

I tested it on Django's repo, it finds the same issues as the fake-taint rule, plus a few more, and less one false positive (?).

The false positive would be this one:
```
bench/django/input/django/django/views/i18n.py
severity:warning rule:open-redirect: Data from request (next_url) is passed to redirect(). This is an open redirect and could be exploited.
Ensure you are redirecting to safe URLs by using django.utils.http.is_safe_url().
See https://cwe.mitre.org/data/definitions/601.html for more information.

31:    next_url = request.POST.get('next', request.GET.get('next'))
32:    if (
33:        (next_url or request.accepts('text/html')) and
34:        not url_has_allowed_host_and_scheme(
35:            url=next_url,
36:            allowed_hosts={request.get_host()},
37:            require_https=request.is_secure(),
38:        )
39:    ):
40:        next_url = request.META.get('HTTP_REFERER')
41:        if not url_has_allowed_host_and_scheme(
42:            url=next_url,
43:            allowed_hosts={request.get_host()},
44:            require_https=request.is_secure(),
45:        ):
46:            next_url = '/'
47:    response = HttpResponseRedirect(next_url) if next_url else HttpResponse(status=204)
```
It uses `url_has_allowed_host_and_scheme` so it shouldn't be reported, I'm not sure why the fake-taint rule reports it.

PR #1394, opened 2021-07-26
Edit PyYAML load rule + add SnakeYAML rule

bella-laybourn:
PyYAML's load_all(...) is as vulnerable as load(...) in the same way, so I added it to the rule. Also, SnakeYAML is vulnerable to deserialization attacks (for example, see here: https://securitylab.github.com/research/swagger-yaml-parser-vulnerability/#snakeyaml-deserialization-vulnerability) so I added a rule for it.


minusworld (approved):

minusworld:
Thanks @bella-laybourn! 🙏

PR #1400, opened 2021-07-29
Remove Semgrep CI; use config without token

minusworld:


mschwager:
What's the preference for this approach? I believe this means we'll miss things like inventory detection, inline PR comments, and all the other App niceties.

minusworld:
It's not my preference, actually, but there are a fair number of contributors who make PRs from forks. When that happens, a token is required for accessing Semgrep App, which forks do not have, causing their PR to fail.

We could switch over to `pull_request_target`, but this causes different problems since it now runs against the main repository's `develop` branch _instead of_ the incoming PR. We tried this once in the past and the repo immediately went into an unstable state.

So, while not my preference, this is a stopgap solution for now until we figure out how to plug into the App in a sustainable way.

mschwager (approved):
Roger that :+1:

PR #1402, opened 2021-07-31
Django/Python: Add path validation rules

malexmave:
Django uses [paths](https://docs.djangoproject.com/en/3.2/ref/urls/#path) for routing requests to different views. If paths are defined multiple times, this can be either inefficient or, in the worst case, lead to [unexpected behavior](https://docs.djangoproject.com/en/3.2/topics/http/urls/#naming-url-patterns). These rules are a first attempt at writing validators for these path issues. Feedback is welcome. In particular, I am unsure if I selected the severity (warning vs. error) and type of issue (maintainability vs. correctness) correctly. 

Let me know if you would like to see changes. If you like these rules, I can write analogous rules for the [`re_path`](https://docs.djangoproject.com/en/3.2/ref/urls/#re-path) function, but I wanted to check if this was considered useful / in scope for the rulesets before taking on the extra work for that. If you would like to se `re_path` rules: Should I extend the existing rules, or write separate rules for `re_path`?

First time contributing to this repository, so please point out any mistakes I may have made.

### DIFF
--- a/python/django/maintainability/duplicate-path-assignment.py
+++ b/python/django/maintainability/duplicate-path-assignment.py
@@ -1,0 +1,142 @@
+from django.urls import path
+
+# ruleid: duplicate-path-assignment
+urlpatterns = [
+    path('path/to/view', views.example_view, name="test"),
+    path('path/to/view', views.example_view, name="test"),
+]
+
+# ruleid: duplicate-path-assignment
+urlpatterns = [
+    path('path/to/view', views.example_view, name="test"),
+    path('path/to/other_view', view.other_view, name="hello"),
+    path('path/to/view', views.example_view, name="test"),
+]
+
+# ruleid: duplicate-path-assignment
+urlpatterns = [
+    path('path/to/view', views.example_view),
+    path('path/to/view', views.example_view),
+]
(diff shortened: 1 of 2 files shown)


minusworld:
These are super cool! And the rules look really, really good! Lots of metadata ;)

As a general rule of thumb, I use `ERROR` for rules that are A) definitely a security vulnerability, or B) definitely going to crash the program. I use `WARNING` for everything else. Since this is only my rule of thumb, it's not consistently used across the ecosystem. `WARNING` works well here, imo.

These fall pretty well into `maintainability`, too. I think that works well as a category.

Overall, these are great! Well written patterns, clear messages, and great metadata. I'll merge these in. :)

minusworld:
For the issue described in the comments with `duplicate-path-assignment-different-names`, I tried a few variations myself here (https://semgrep.dev/s/4PZ5) and was similarly unable to prevent 2 findings from coming up.

This is probably a bug, or at least an unintuitive behavior, that's worth filing on the [Semgrep repo](https://github.com/returntocorp/semgrep) itself.

minusworld (approved):

PR #1403, opened 2021-08-03
[JS] angular: security: Fix rules to use the latest Semgrep features

IagoAbal:
In Semgrep v0.38.0 we released flow-based constant propagation.

In Semgrep v0.58.0 we released a new iteration of taint-mode that should
(we hope) allow us to get rid of fake-taint rules.

Let's use them!

### DIFF
--- a/javascript/angular/security/detect-angular-element-methods.js
+++ b/javascript/angular/security/detect-angular-element-methods.js
@@ -9,6 +9,10 @@ app.controller('myCtrl', function($scope) {
     // ruleid: detect-angular-element-methods
     var now = angular.element($scope.input).wrap();
 
+    var el = angular.element($scope.input);
+    // ruleid: detect-angular-element-methods
+    var now = el.html();
+
     return now;
 
 });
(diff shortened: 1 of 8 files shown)


inkz (dismissed):
awesome! thanks for doing it :+1:

minusworld:
@inkz Do we still want to check the perf on these?

IagoAbal:
> @inkz Do we still want to check the perf on these?

I can run some benchmarks, is there any specific JS repo(s) that we consider very important to measure against? Otherwise I try with whatever is in semgrep/perf.

minusworld:
Whatever is in `perf/` is fine. If there's nothing in there, you can find a random Angular repository and use that

IagoAbal:
I'm using our benchmark for parsing stats, need to dig a bit more into this... it's ~25k files and the new rules take almost 20 times longer... but they also give 5 times more findings...

IagoAbal:
Partly it's due to minimized JS files that we should not be analyzing in the first place...

IagoAbal:
I think this is because `-filter_irrelevant_rules` (aka `-fast`) doesn't work as well for taint-mode rules. Need to fix that.

IagoAbal:
Fixed by returntocorp/semgrep#3749, we can merge this next week when 0.64.0 is released.

inkz (approved):
Amazing!

PR #1405, opened 2021-08-03
Add brakeman rules

emjin:


ruby/rails/security/audit/xss/avoid-default-routes.rb
```diff
@@ -0,0 +1 @@
+map.connect ":controller/:action/:id"
```
emjin: Creating a better test is impossible for now (need some sort of functionality to ignore paths in tests)

inkz: thats ok, but I think you need to add `ruleid: avoid-default-routes` to the test

inkz (changes requested):
https://github.com/returntocorp/semgrep-rules/pull/1405/files#r682224775

ruby/rails/security/audit/xss/avoid-default-routes.rb
minusworld: One way we might be able to fix this is to see if scanning for `*routes.rb` correctly picks up `config/routes.rb`, and then we could rename this test file to `avoid-default-routes.routes.rb`. That might get the test engine to pick it up.

inkz: oh I get it now, sorry for the first comment, my bad, we need to scan `config/routes.rb` specifically

emjin:
> https://github.com/returntocorp/semgrep-rules/pull/1405/files#r682224775

Just to clarify since there was some discussion, what change should be made?

ruby/rails/security/audit/xss/avoid-tainted-file-access.yaml
```diff
@@ -0,0 +1,61 @@
...
+  metadata:
+    owasp: 'A6: Security Misconfiguration'
+    cwe: 'CWE-276: Incorrect Default Permissions'
+    references:
+    - https://github.com/presidentbeef/brakeman/blob/main/docs/warning_types/file_access/index.markdown
+    category: security
+    technology:
+    - rails
+  message: |
+    Using user input when accessing files is potentially dangerous. A malicious actor could use this to modify or access files they have no right to. [Not sure what to say for how to fix it]
```
minusworld: @minusworld @clintgibler Augment message

ruby/rails/security/audit/xss/avoid-link-to.yaml
```diff
@@ -0,0 +1,21 @@
+rules:
+- id: avoid-link-to
+  metadata:
+    owasp: 'A6: Security Misconfiguration'
+    cwe: "CWE-279: Improper Neutralization of Input During Web Page Generation ('Cross-site Scripting')"
+    references:
+    - https://brakemanscanner.org/docs/warning_types/link_to/
+    - https://brakemanscanner.org/docs/warning_types/link_to_href/
```
emjin, line 8: link_to and link_to_href are different errors in brakeman, but I didn't know which one to follow more closely because we don't check ruby versions. In the end I biased towards laziness, but we could mimic some of their strategies to avoid false positives. Interesting discussion: https://github.com/presidentbeef/brakeman/pull/45

ruby/rails/security/audit/xss/avoid-tainted-file-access.yaml
```diff
@@ -0,0 +1,61 @@
...
+        - pattern: symlink
+        - pattern: syscopy
+        - pattern: sysopen
+        - pattern: truncate
+        - pattern: unlink
+  - metavariable-pattern:
+      metavariable: $STR
+      patterns:
+      - pattern: '...'
+      - pattern-not: |
```
clintgibler: What are you trying to do here? Not sure I understand the `pattern: '...'` and `pattern-not: "..."`.

ruby/rails/security/audit/xss/avoid-tainted-file-access.yaml
clintgibler: Currently the way this is written assumes that all of the functions in `metavariable-pattern` for `$X` could happen for each of the modules: `Dir`, `File`, etc.

But in many cases, not all modules have all of those methods - for example [Net::FTP](https://docs.ruby-lang.org/en/2.5.0/Net/FTP.html) does not have `entries`, `glob`, and a bunch of other ones.

Second, I think warning on any non string literal call to any of these function may be quite noisy in practice.

Might be a good use of taint mode to see if controller input (URL param, form param, cookie, ...) can reach one of these functions.

Third: @minusworld Thoughts on splitting this rule out into multiple rules based on sink? Something like: shell exec (code injection), file I/O (arbitrary file read/write), and network (Net::FTP/Net::HTTP -> open redirect or SSRF or something)?

ruby/rails/security/audit/xss/avoid-tainted-file-access.yaml
emjin: Basically, `Dir.link(params)` is bad, but `Dir.link("params")` is fine.

emjin: Oh I see the confusion...rewrote this to make it more clear

ruby/rails/security/audit/xss/avoid-tainted-file-access.yaml
emjin: 1. Mm I was wondering this, because (if I read correctly) brakeman also took this approach of checking every combination, including ones that aren't possible? I do think it's easier and ok to do it this way, since it shouldn't cause false positives.
2. I can fix that! Now that I understand ruby/brakeman better, I see that I misinterpreted this

ruby/rails/security/audit/xss/avoid-tainted-file-access.yaml
```diff
@@ -0,0 +1,61 @@
...
+    references:
+    - https://github.com/presidentbeef/brakeman/blob/main/docs/warning_types/file_access/index.markdown
+    category: security
+    technology:
+    - rails
+  message: |
+    Using user input when accessing files is potentially dangerous. A malicious actor could use this to modify or access files they have no right to. [Not sure what to say for how to fix it]
+  languages: [ruby]
+  severity: WARNING
+  mode: taint
```
emjin: Note: taint mode currently has a duplicate result problem

ruby/rails/security/audit/xss/avoid-default-routes.yaml
```diff
@@ -0,0 +1,24 @@
...
+- id: avoid-default-routes
+  metadata:
+    owasp: 'A6: Security Misconfiguration'
+    cwe: 'CWE-276: Incorrect Default Permissions'
+    references:
+    - https://github.com/presidentbeef/brakeman/blob/main/docs/warning_types/default_routes/index.markdown
+    category: security
+    technology:
+    - rails
+  message: |
```
emjin, line 11: Reasonably confident about this message?

ruby/rails/security/audit/xss/avoid-link-to.yaml
emjin: Updated to follow `link_to_href` more closely, still probably missing some cases. Message should possibly also be changed?

ruby/rails/security/audit/xss/avoid-link-to.yaml
```diff
@@ -0,0 +1,37 @@
...
+    category: security
+    technology:
+    - rails
+  message: |
+    This code includes user input in `link_to`. In Rails 2.x, the body of `link_to` is not escaped.
```
emjin, line 13: I'm confused about the "values starting with `javascript:` or `data:` ..." statement (copied from brakeman), did not see this reflected in the brakeman rule? Is this about the user input? What should they do here?

ruby/rails/security/audit/xss/avoid-redirect.yaml
```diff
@@ -0,0 +1,50 @@
...
+  metadata:
+    owasp: 'A6: Security Misconfiguration'
+    cwe: 'CWE-276: Incorrect Default Permissions'
+    references:
+    - https://brakemanscanner.org/docs/warning_types/redirect/
+    category: security
+    technology:
+    - rails
+  message: |
+    When a redirect uses user input, a malicious user can spoof a website under a trusted URL or access restricted parts of a site. When using user-supplied values, sanitize the value before using it for the redirect.
```
emjin, line 12: Didn't know what to say for solution

ruby/rails/security/audit/xss/avoid-render-dynamic-path.yaml
```diff
@@ -0,0 +1,30 @@
...
+  metadata:
+    owasp: 'A6: Security Misconfiguration'
+    cwe: 'CWE-276: Incorrect Default Permissions'
+    references:
+    - https://brakemanscanner.org/docs/warning_types/dynamic_render_paths/
+    category: security
+    technology:
+    - rails
+  message: |
+    Avoid rendering user input. It may be possible for a malicious user to input a path that lets them access a template they shouldn't. [Not sure about solution]
```
emjin: Not sure about solution

ruby/rails/security/audit/xss/avoid-session-manipulation.yaml
```diff
@@ -0,0 +1,21 @@
...
+  metadata:
+    owasp: 'A6: Security Misconfiguration'
+    cwe: 'CWE-276: Incorrect Default Permissions'
+    references:
+    - https://brakemanscanner.org/docs/warning_types/session_manipulation/
+    category: security
+    technology:
+    - rails
+  message: |
+    [Sorry I'm really confused about why this is actually a vulnerability. It looks like a read, but the description sounds like it's a write]
```
emjin: Very confused

ruby/rails/security/audit/xss/avoid-redirect.yaml
```diff
@@ -0,0 +1,50 @@
...
+  languages: [ruby]
+  severity: WARNING
+  mode: taint
+  pattern-sources:
+  - pattern: params
+  - pattern: cookies
+  - pattern: request.env
+  - patterns:
```
emjin, line 20: Did not copy the rule entirely, this may be missing clauses

ruby/rails/security/audit/xss/avoid-session-manipulation.yaml
inkz: it is kinda both read an write

```ruby
user_id = session[params[:name]]
current_user = User.find(user_id)
```
in this example user_id is expected to be retrieved form session, lets say it is expected that `:name` will have a value smth like `"user_id"`, but bc user can control it, there is an opportunity to retrieve smth else e.g. `"_csrf_token"` of `"username"` etc. But the data will be treated as `user_id` and it can lead to unexpected behavior

inkz: so it is a good rule! just need a message for it :+1:

ruby/rails/security/audit/xss/avoid-redirect.yaml
inkz: I think it is good enough

ruby/rails/security/audit/xss/avoid-link-to.yaml
inkz: Yes, `javascript:` and `data:` are the possible values of user input that can be malicious, so the message is correct here :+1:

ruby/rails/security/audit/xss/avoid-default-routes.yaml
minusworld: This is good 👍

ruby/rails/security/audit/xss/avoid-redirect.yaml
minusworld: There exists this `sanitize` helper function that we could recommend directly: https://api.rubyonrails.org/classes/ActionView/Helpers/SanitizeHelper.html

ruby/rails/security/audit/xss/avoid-redirect.yaml
minusworld: We could test this at scale on MAP to see how it performs. @emjin Would you be interested in doing this? Otherwise someone else can test it

ruby/rails/security/audit/xss/avoid-redirect.yaml
minusworld: ^ Which I suppose should also be added to the list of sanitizers

ruby/rails/security/audit/xss/avoid-render-dynamic-path.yaml
minusworld: The most generic solution would probably be to check templates against an allowlist.
Something like "To avoid this, check dynamic template paths against a predefined allowlist to make sure it's an allowed template."

ruby/rails/security/audit/xss/avoid-redirect.yaml
```diff
@@ -0,0 +1,50 @@
+rules:
+- id: avoid-redirect
+  metadata:
+    owasp: 'A6: Security Misconfiguration'
+    cwe: 'CWE-276: Incorrect Default Permissions'
```
minusworld: This is better categorized as `CWE-601: URL Redirection to Untrusted Site ('Open Redirect')`

ruby/rails/security/audit/xss/avoid-render-dynamic-path.yaml
```diff
@@ -0,0 +1,30 @@
+rules:
+- id: avoid-render-dynamic-path
+  metadata:
+    owasp: 'A6: Security Misconfiguration'
+    cwe: 'CWE-276: Incorrect Default Permissions'
```
minusworld: CWE-22: Improper Limitation of a Pathname to a Restricted Directory ('Path Traversal')

ruby/rails/security/audit/xss/avoid-render-dynamic-path.yaml
```diff
@@ -0,0 +1,30 @@
+rules:
+- id: avoid-render-dynamic-path
+  metadata:
+    owasp: 'A6: Security Misconfiguration'
```
minusworld: A1: Injection

ruby/rails/security/audit/xss/avoid-redirect.yaml
```diff
@@ -0,0 +1,50 @@
+rules:
+- id: avoid-redirect
+  metadata:
+    owasp: 'A6: Security Misconfiguration'
```
minusworld: This is better categorized as `A1: Injection`

ruby/rails/security/audit/xss/avoid-tainted-file-access.yaml
minusworld: This is external to the rule - once resolved in Semgrep, it'll be resolved here. Resolving convo

ruby/rails/security/audit/xss/avoid-tainted-file-access.yaml
minusworld: (3) probably good to split up, that would be consistent with the rest of the rules

ruby/rails/security/audit/xss/avoid-link-to.yaml
minusworld: No worries about `link_to_href`

ruby/rails/security/audit/xss/avoid-session-manipulation.yaml
```diff
@@ -0,0 +1,23 @@
...
+    cwe: 'CWE-276: Incorrect Default Permissions'
+    references:
+    - https://brakemanscanner.org/docs/warning_types/session_manipulation/
+    category: security
+    technology:
+    - rails
+  message: |
+    This gets data from session using user inputs. A malicious user may be able to retrieve
+    information from your session that you didn't intend them to. Do not use user input as
+    a session key.
```
minusworld: "Or write a sanitizer function with an allowlist of valid keys."

minusworld (changes requested):
(1) test `avoid-default-routes` with the filename hack
(2) update a few CWE and OWASP fields
(3) break up `avoid-tainted-file-access`

minusworld:
(4) [maybe] Add source: `request.query_parameters`
cc @clintgibler Does this look like the way Rails accesses GET query parameters?

ruby/rails/security/audit/xss/avoid-redirect.yaml
emjin: hm I have no idea how to do that, willing to learn at some point but preferably not this week or next

minusworld (approved):
Had to change one comment to a `todoruleid` until https://github.com/returntocorp/semgrep/issues/3880 is resolved, but otherwise, merge away!

minusworld:
@inkz Need your stamp here, too, if your concerns are addressed! 👍

inkz:
@minusworld looking :eyes:

inkz (approved):
lgtm!

PR #1410, opened 2021-08-06
Android: Add rules for network-security-config files

malexmave:
Android apps can specify security policies for network traffic in two places: the Manifest-file (outdated) and a network-security-policy.xml file. This PR adds rules for detecting a number of common issues related to these files, see the rule messages for more details. 

I figure these rules walk the line between security and best-practice as a categorization - I chose best-practice as there are sometimes good reasons to deviate from these best practices, and tried to write the rules in a way where they respect the [`tools:ignore` attribute](https://developer.android.com/studio/write/tool-attributes#toolsignore) (although that may not always be successful due to limitations of the `generic` parser, and my limited knowledge of which of these attributes exist) to try to reduce the amount of annoyance to developers.

As always, feedback is very welcome. I chose not to submit these rules to [MobSF](https://github.com/MobSF) as they already have [their own checks for these issues](https://github.com/MobSF/Mobile-Security-Framework-MobSF/blob/master/mobsf/StaticAnalyzer/views/android/network_security.py), which are implemented outside of their corpus of semgrep rules.

### DIFF
--- a/java/android/best-practice/manifest-security-features.yaml
+++ b/java/android/best-practice/manifest-security-features.yaml
@@ -1,0 +1,54 @@
+rules:
+- id: manifest-usesCleartextTraffic-true
+  languages:
+  - generic
+  message: >
+    The Android manifest is configured to allow non-encrypted connections.
+    Evaluate if this is necessary for your app, and disable it if appropriate.
+    This flag is ignored on Android 7 (API 24) and above if a Network Security
+    Config is present.
+  metadata:
+    category: best-practice
+    license: Commons Clause License Condition v1.0[LGPL-2.1-only]
+    technology:
+    - android
+    references:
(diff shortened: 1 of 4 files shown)


java/android/best-practice/manifest-security-features.yaml
```diff
@@ -0,0 +1,54 @@
...
+    license: Commons Clause License Condition v1.0[LGPL-2.1-only]
+    technology:
+    - android
+    references:
+    - https://developer.android.com/guide/topics/manifest/application-element#usesCleartextTraffic
+    - https://developer.android.com/training/articles/security-config
+  patterns:
+  - pattern: |
+      android:usesCleartextTraffic="true"
+  - pattern-not-inside: |
```
minusworld, line 21: Nice

minusworld (approved):
These look awesome. Thank you!

PR #1411, opened 2021-08-06
Use dashes instead of underscores in file names

mjambon:
This is for consistency with other semgrep rules. I did that only for ocaml rules.

Bash code to do that was:
```
for x in */*_*; do echo $x; git mv $x $(echo $x | tr _ - ) ; done
```


inkz (approved):
:+1:

PR #1412, opened 2021-08-06
Add rules related to handling CRLF line endings in OCaml

mjambon:
These are my first semgrep rules for actual use! Please let me know of anything that could be improved.

I tested this, including the autofix, in the live editor: https://semgrep.dev/s/mjambon:ocaml-crlf-support

The sample OCaml code compiles and might actually work, although I offer no guarantee.


inkz (approved):
lgtm :+1:

inkz:
P.S. very good verbose messages for the findings! :+1:

PR #1415, opened 2021-08-11
Expand rule to cover a broader set of origin validation scenarios

bernardoamc:
The current rule takes into account:

1. Inline function declarations within `addEventListener`
2. Specific operators for origin validation within the handler

The new rule expands it to cover:

1. Inline and external function declarations
2. Checks whether origin is being called or not within an if clause in
   the handler

To see this rule in action: https://semgrep.dev/s/rkOR

If we think checking for specific equality operators is something
desirable we can change the rule to something like: https://semgrep.dev/s/brjE


inkz (approved):
Awesome! very good addition, thank you @bernardoamc :+1:

PR #1418, opened 2021-08-16
add lazy-load-module rule

inkz:
Simple rule for bad pattern: dynamic `require` usage

### DIFF
--- a/javascript/lang/best-practice/lazy-load-module.yaml
+++ b/javascript/lang/best-practice/lazy-load-module.yaml
@@ -7,10 +7,13 @@ rules:
           ...
       }
   message: |
-    Lazy loading can complicate code bundling if care is not taken.
+    Lazy loading can complicate code bundling if care is not taken, also `require`s are run synchronously by Node.js.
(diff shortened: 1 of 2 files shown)


javascript/lang/best-practice/lazy-load-module.yaml
```diff
@@ -0,0 +1,16 @@
+rules:
+- id: lazy-load-module
+  patterns:
+  - pattern: require(...)
+  - pattern-inside: |
+      function $NAME(...) {
+          ...
+      }
+  message: |
+    Lazy loading can complicate code bundling if care is not taken.
```
minusworld: What's the recommended fix for this? Just to do global requiring?

inkz: Yes! and I definitely need to add this information to the message

inkz: done!

minusworld (approved):
excellent recommendation :D

PR #1423, opened 2021-08-18
add code-string-concat rule

inkz:
Command injection prevention rule, looking for string concatenation of stuff like `"function(){" + userInput + "}"`


minusworld (approved):
lgtm

PR #1425, opened 2021-08-18
Add rule for detecting FLASK_ENV set to 'development'

minusworld:


mschwager (approved):
I would leave `accidentally` out of the filename since we don't know the intent. I.e. we don't know if it's an accident or not, just that it's enabled.

Other than that, LGTM!

Issue #1431, opened 2021-08-22
Many rules have hard-coded line breaks in messages, resulting in poor terminal display


huonw:
**Describe the bug**

Many rules have messages that include literal line-breaks, which results in poor display in a terminal. This is usually because the message field is configured with `|`. For instance, the following two code blocks are equivalent:

```yaml
message: |
   Some message here
...
message: |
   Some wrapped 
   message
```

```yaml
message: "Some message here\n"
...
message: "Some wrapped\nmessage\n"
```

(Note the trailing line break too.)

https://yaml-multiline.info has more info/description.

Arguably this is a bug in semgrep itself, which could strip out the newlines from the messages.

**To Reproduce**

Run any rule that contains a message configured with `|`, like https://github.com/returntocorp/semgrep-rules/blob/e6c956f4fc5832594e6e8ff89123726c8e54ec6c/yaml/semgrep/missing-message-field.yaml#L3-L7

```shell
semgrep -c yaml/semgrep/missing-message-field.yaml yaml/semgrep/missing-message-field.test.yaml                 
```
```
running 1 rules...
yaml/semgrep/missing-message-field.test.yaml
severity:warning rule:yaml.semgrep.missing-message-field: This rule does not have a message. Semgrep requires that
rules have a message. Include a message to explain what the rule does.
Consider writing a message that explains why this is an issue and
how to fix it.

3:- id: unchecked-subprocess-call
ran 1 rules on 1 files: 1 findings
```

Note the rule description is:

1. broken across multiple lines, exactly matching the file configuration
2. has an extra blank line after the definition

This is even worse when using flags like `--vim` or `--emacs`, that are explicitly documented as outputting in a "single-line" format, as the message still breaks across multiple lines.

```shell
semgrep --vim -c yaml/semgrep/missing-message-field.yaml yaml/semgrep/missing-message-field.test.yaml  
```

```
running 1 rules...
yaml/semgrep/missing-message-field.test.yaml:3:3:W:yaml.semgrep.missing-message-field:This rule does not have a message. Semgrep requires that
rules have a message. Include a message to explain what the rule does.
Consider writing a message that explains why this is an issue and
how to fix it.

ran 1 rules on 1 files: 1 findings
```

**Expected behavior**

Messages should almost always be displayed on a single line, especially when using the `--vim`/`--emacs` format.

**Priority**
How important is this to you?
- [ ] P0: blocking me from making progress
- [ ] P1: this will block me in the near future
[...]

PR #1432, opened 2021-08-22
Warn on multi-line 'message' in semgrep rules

huonw:
This adds a rule `yaml/semgrep/multi-line-message` that flags multi-line messages in semgrep rules, identified as suboptimal in #1431. As discussed in #1431, this may actually be better addressed in semgrep itself, rather than a rule/convention. This is done in the first commit e08ebd8.

It also rewrites all existing rules that use `message: |` to use `message: >-` instead (mechanically: `sed -i '' 's/message: |/message: >-/g' **/*.yaml`), in the second commit 776d488.


yaml/semgrep/multi-line-message.yaml
```diff
@@ -0,0 +1,18 @@
...
+- id: multi-line-message
+  message: >-
+    This rule has a multi-line message field, which may display poorly
+    in a terminal. Consider ensuring it is on one line. For example,
+    use `message: >-`, not `message: |`.
+  languages: [yaml]
+  patterns:
+  - pattern-inside: 'rules: [..., $RULE, ...]'
+  - pattern: |
+      message: "=~/[\\n\\r]/"
```
huonw, line 11: I couldn't work out the right escaping to express this with metavariable-regex, so I've gone with the deprecated string matching syntax instead.

minusworld:
Cool, thanks! I'm gonna skim over all the changes just to make sure there aren't any adverse effects, then we can get this rule in 👍

minusworld (changes requested):
Looks like there's a few rules that broke: https://github.com/returntocorp/semgrep-rules/pull/1432/checks?check_run_id=3447556374

huonw:
Ah, thanks for starting CI. I'll look at fixing those sometime over the next day or two.

huonw:
I think I've addressed most of them now.

minusworld:
Huh, weird that `test-develop` is failing. I'll have to investigate further. It's not required but I wanna double-check to make sure it doesn't break things in the next release. Aside from that, this PR is pretty much done. Thanks! 😄

huonw:
Great! 

I'm sure you'll find this quickly enough, but it looks like other builds are failing with a similar `cp: can't stat '/src/semgrep_runs_output.tar.gz': No such file or directory` error in `test-develop` too, e.g. #1440.

minusworld:
@huonw OK, we've made some progress here - the ❌ 's are now a result of this branch being out-of-sync with the `develop` branch upstream. If you merge the develop branch in here, I think that will fix the problem. Thanks for your patience 😅

huonw:
I've merged and updated the messages of a few extra rules that have landed in develop recently.

minusworld (approved):

minusworld:
Thanks @huonw !

PR #1427, opened 2021-08-19
Fix rules following returntocorp/semgrep#3714

IagoAbal:
returntocorp/semgrep#3714 fixes constant propagation for expressions where type arguments are involved as in Java's `new` operator. For the tests to pass we need to update semgrep-rules too.

PR #1433, opened 2021-08-23
Add node-mssql-sqli rule

inkz:
new rule for SQL injection via `mssql` module

### DIFF
--- a/javascript/lang/security/audit/sqli/node-mssql-sqli.yaml
+++ b/javascript/lang/security/audit/sqli/node-mssql-sqli.yaml
@@ -1,0 +1,59 @@
+rules:
+- id: node-mssql-sqli
+  languages:
+  - js
+  - typescript
+  message: >
+    Detected string concatenation with a non-literal variable in a
+    node-postgres JS SQL statement. This could lead to SQL injection if the variable is user-controlled
+    and not properly sanitized. In order to prevent SQL injection,
+    used parameterized queries or prepared statements instead.
+    You can use parameterized statements like so:
+    `$REQ.query('SELECT $1 from table', [userinput])`
+  metadata:
+    category: security
+    license: Commons Clause License Condition v1.0[LGPL-2.1-only]
+    references:
+    - https://node-postgres.com/features/queries
+  patterns:
+  - pattern-inside: |
+      require('mssql')
+      ...
+  - pattern-inside: |
+      $REQ = $POOL.request(...)
+      ...
+  - pattern-either:
+    - patterns:
+      - pattern-either:
+        - pattern: |
+            $REQ.query($QUERY,...)
+        - pattern: |
+            $REQ.query($QUERY, ...).$INTFUNC(...)
+      - pattern-either:
+        - pattern-inside: |
+            $QUERY = $X + $Y
+            ...
+        - pattern-inside: |
+            $QUERY += $X
+            ...
+        - pattern-inside: |
+            $QUERY = $X.concat($Y)
+            ...
+        - pattern-inside: |
+            $QUERY = `...${...}...`
+            ...
+      - pattern-not-inside: |
+          $QUERY += "..."
+          ...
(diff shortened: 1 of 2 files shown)


inkz:
failure related to a different rule

javascript/lang/security/audit/sqli/node-mssql-sqli.yaml
```diff
@@ -0,0 +1,59 @@
...
+            $QUERY += $X
+            ...
+        - pattern-inside: |
+            $QUERY = $X.concat($Y)
+            ...
+        - pattern-inside: |
+            $QUERY = `...${...}...`
+            ...
+      - pattern-not-inside: |
+          $QUERY += "..."
```
clintgibler, line 46: With the `+=`, couldn't user input still be in `$QUERY`? Something like:

```javascript
var foo = "select * from [foobar] WHERE user_id =" + userInput);
foo += " sort desc";
pool.query(foo);
```

Maybe it's good to keep and there are cases I'm not thinking about properly, just wanted to ask. Sould it be `$QUERY = "..."`?

Would `metavariable-pattern` potentially help simplify?

clintgibler (approved):
I had a small question and I wonder if any cleaning up/improving could be done, but overall lgtm 👍

javascript/lang/security/audit/sqli/node-mssql-sqli.yaml
```diff
@@ -0,0 +1,59 @@
...
+          ...
+      - pattern-not-inside: |
+          $QUERY = "..." + "..."
```
emjin, line 49: @IagoAbal This seems like a feasible constant prop improvement.

emjin: (If not there already)

javascript/lang/security/audit/sqli/node-mssql-sqli.yaml
emjin: Not sure if metavariable-pattern would be great here. I think this seems like something taint mode could do?

minusworld:
The failing `test-develop` check is expected until https://github.com/returntocorp/semgrep-rules/pull/1427 is merged, which requires the next release of Semgrep.

inkz:
@clintgibler @emjin very good questions, all SQLi rules look like a perfect candidates for taint mode, but some further investigation needed for this. I decided to push this rule first to unblock our user (here: https://github.com/returntocorp/semgrep-rules/issues/1430)

I am going to vote for Sql Injection be our next problem-centric / cheat sheet series :+1:

javascript/lang/security/audit/sqli/node-mssql-sqli.yaml
IagoAbal: You shouldn't need `"..." + "..."` with const-prop enabled, just `"..."` should work.

JasonCubic:
@inkz 

I just tested this out, and it works so great!  :-)

One small note though.

The parameter code example on the description for this rule isn't correct.  using parameters in mssql is a little more complex than the example in the rule description indicates.  https://www.npmjs.com/package/mssql#input-name-type-value

I would suggest either leaving the example out or linking to the mssql docs.  because this:
```
const dbResult = await request.query('SELECT * FROM [ramshackle].[dbo].[users] WHERE user_id = $1', [id]);
```
only returns the user_id of 1 no matter what the id variable is set to (it ignores the $).

I think mssql makes you do something like this:
```
request.input('USER_ID', mssql.Int, id);
const dbResult = await request.query('SELECT * FROM [ramshackle].[dbo].[users] WHERE user_id = @USER_ID');
```

inkz:
@JasonCubic thanks a lot for help! :+1:
I used `postgresql` rule as a template and forgot to change description.
Created a fix: https://github.com/returntocorp/semgrep-rules/pull/1449

PR #1435, opened 2021-08-24
Prototype Pollution rules (part 1)

inkz:
Rules for identifying Prototype Pollution (https://github.com/HoLyVieR/prototype-pollution-nsec18/blob/master/paper/JavaScript_prototype_pollution_attack_in_NodeJS.pdf)


javascript/lang/security/audit/prototype-pollution/prototype-pollution-assignment.yaml
```diff
@@ -0,0 +1,44 @@
...
+      }
+  - metavariable-pattern:
+      metavariable: $A
+      patterns:
+      - pattern-not: '"..."'
+  - metavariable-pattern:
+      metavariable: $B
+      patterns:
+      - pattern-not: '"..."'
+  - metavariable-regex:
```
mschwager: This is checking for integer index access right? We could do something similar to the [GitLab rule's](https://gitlab.com/gitlab-org/security-products/analyzers/semgrep/-/blob/main/rules/eslint.yml) here:

```yaml
  - pattern-not: |
      $X[($B : float)] = ...
  - pattern-not-inside: |
      $X = $SMTH[($A : float)]
      ...
```

I believe this will have the added advantage of filtering out things like:

```javascript
index = 5;
obj[index] = req.input;
```

mschwager (approved):
LGTM, don't feel the need to block on the `metavariable-regex`, just a suggestion :+1:

minusworld:
lgtm 👍  Can we check these on MAP?

javascript/lang/security/audit/prototype-pollution/prototype-pollution-assignment.yaml
inkz: ooh nice! thanks :ok_hand:

inkz: updated the rules :+1:

inkz:
@minusworld 
* MAP tests here
https://dev.massive.ret2.co/jobs/1409
https://dev.massive.ret2.co/jobs/1410

* I triaged some findings and (very) approximate stats are

Rule id | TP  / FP
------------ | -------------
prototype-pollution-function  | 6 / 4
prototype-pollution-loop | 9 / 1
prototype-pollution-assignment | 9 / 5

* I made notes on FPs and going to work on them now

* also good news is that `prototype-pollution-function` rule, which uses taint mode and seems to be a little slow, actually working fine and find stuff even in large and/or minified files :+1:

minusworld:
Awesome. That's a brilliant-looking table 😄 

I'll approve - do you think we might be able to get the TP rate for `prototype-pollution-function` up by just a hair in the future?

minusworld (approved):

colleend (approved):
the TP/FP table is awesome :D

inkz:
so I updated the rules to mitigate some FPs

inkz:
I guess tests are failing because of https://github.com/returntocorp/semgrep-rules/pull/1446#issuecomment-909769327

PR #1437, opened 2021-08-25
ssrf pattern for csharp

gdraperi:
Hello,

I am pushing a semgrep rule to catch SSRF (Server Side Request Forgery) vulnerabilities in CSharp.

First time I work for Semgrep so I might have missed some steps.

Regards,

Gregory DRAPERI


minusworld:
@gdraperi This looks great! And all the checks passed, so you didn't miss any steps!

I have a quick question: do you know what a "good" example looks like? It's helpful to have a good example in the test code marked with `// ok: ssrf` which gives users an opportunity to see how they might fix things. (It also is good to tell users how to fix the issue in the rule message.)

Thanks! 😄

gdraperi:
@minusworld Thanks for the review! I have added some additional code and some pattern. Not sure about the OWASP ID?

minusworld:
I apologize - I neglected to mention how the comment syntax worked. The documentation is here: https://semgrep.dev/docs/writing-rules/testing-rules/#example

For cases that we want to match, the annotation is `// ruleid: <rule-id>`
For cases that we do not want to match, the annotation is `// ok: <rule-id>`

I'm also not sure about the OWASP - maybe it makes more sense as A3: Sensitive Data Exposure?

minusworld (approved):
Awesome!! Looks great 👍  😄

PR #1440, opened 2021-08-26
Checkov iam rules

colleend:
All the Checkov AWS IAM terraform rules, taken from https://github.com/bridgecrewio/checkov/tree/ca830e14745c2c8e1b941985f305abe985d7f1f9/checkov/terraform/checks/data/aws.

As a note, we don't have full terraform support yet, so some of the rules may look a little hacky because of that fact. Generic mode worked pretty well, but results will definitely be better once we have the language support.


inkz (approved):
great work!

minusworld:
https://linear.app/r2c/issue/RULES-18

colleend:
I checked this with Checkov's vulnerable web app [terragoat](https://github.com/bridgecrewio/terragoat#aws-setup). 

With Checkov IAM rules, there were 2 rules that flagged the same piece of code:
* [CKV_AWS_111](https://github.com/bridgecrewio/checkov/blob/ca830e14745c2c8e1b941985f305abe985d7f1f9/checkov/terraform/checks/data/aws/IAMWriteAccess.py)
* [CKV_AWS_109](https://github.com/bridgecrewio/checkov/blob/ca830e14745c2c8e1b941985f305abe985d7f1f9/checkov/terraform/checks/data/aws/IAMPermissionsManagement.py)

on the code:
```
		29 | data aws_iam_policy_document "policy" {
		30 |   statement {
		31 |     actions = ["es:*"]
		32 |     principals {
		33 |       type        = "AWS"
		34 |       identifiers = ["*"]
		35 |     }
		36 |     resources = ["*"]
		37 |   }
		38 | }
```

As you can see, this is because `es:*` with the star includes a banned function in the two checks. (I'm pretty sure CKV_AWS_111 is a subset of CKV_AWS_109). However, in the semgrep rules, I only ban the specific function. Therefore, after running semgrep on the terraform code, we get no results. 

I'm going to edit these rules so that we can catch the case where groups of functions including the banned function should be flagged.

colleend:
Update: I added in a data object that an IAM policy can be created with.

Now, we catch the issue that checkov flagged along with a separate issue: 

```
terraform/aws/eks.tf
severity:warning rule:semgrep-rules.terraform.lang.security.iam.no-iam-creds-exposure: Ensure IAM policies don't allow credentials exposure. Credentials exposure actions return credentials
as part of the API response, and can possibly lead to leaking important credentials. Instead,
use another action that doesn't return sensitive data as part of the API response.

10:    actions = ["sts:AssumeRole"]
```
I investigated this and this seems to be inline with what the credentials exposure rule is trying to catch.

minusworld (approved):

PR #1441, opened 2021-08-26
Improve AngularJS rules

emjin:
Delete unnecessary pattern-either clause in detect-angular-open-redirect.yaml


emjin:
Closing this, will revisit after Lewis's stream

PR #1446, opened 2021-08-31
javascript: Fix test expectations after returntocorp/semgrep#3781

IagoAbal:
Taint mode sometimes reported duplicate matches and
returntocorp/semgrep#3781 fixes that.


brendongo:
Another circular dependency with semgrep. The semgrep-rules-latest tests will fail until next release

PR #1447, opened 2021-08-31
add rule for public s3 bucket

ap00rv:
- S3 bucket CloudFormation objects have a property called as `PublicAccessBlockConfiguration ` , which in turn, has 4 sub-properties. 
     - "BlockPublicAcls" 
     - "BlockPublicPolicy" 
     - "IgnorePublicAcls" 
     - "RestrictPublicBuckets"
   
    
   If any of these properties is set to `false`, there is a chance that an attacker may be able to perform some actions on some or all objects in the bucket.  This rule checks if any of these properties is set to `false` and if the detected pattern is part of an S3 bucket definition. 

More information here : https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-s3-bucket-publicaccessblockconfiguration.html

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,3 +3,4 @@ __pycache__/
 .pytest_cache/
 .DS_Store
 .vscode/
+.venv
--- a/json/aws/security/public-s3-bucket-test.json
+++ b/json/aws/security/public-s3-bucket-test.json
@@ -1,0 +1,27 @@
+
+  {
+    "Resources": {
+      "MyBucketF68F3FF0": {
+        "Type": "AWS::S3::Bucket",
+        "Properties": {
+          "BucketEncryption": {
+            "ServerSideEncryptionConfiguration": [
+              {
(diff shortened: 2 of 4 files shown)


ap00rv:
I am not sure how to address failure message in `semgrep-rules-test / test-develop ` check. Can someone please help me to understand the problem ?

minusworld:
@ap00rv  I think this might be a problem unrelated to this PR because other PRs are also failing with the same issue. Someone will look into it and resolve it soon. In any case, it's not a required check so this can probably be approved and merged in. I or someone else will review the rule in a little while. 👍

minusworld (approved):
Finally figured out the CI issues. Looks great! Thank you!

minusworld:
By the way, we're trying out a new thing where new rules get announced by a bot on Twitter. Yours should be live: https://twitter.com/SemgrepRegistry

PR #1449, opened 2021-09-01
Fix node-mssql-sqli description

inkz:
As it was mentioned here: https://github.com/returntocorp/semgrep-rules/pull/1433#issuecomment-909620481
I forgot to change description while copy-pasting from `postgresql` version of the rule :smile:


colleend (approved):
lgtm! seems like the CI is failing from something unrelated to this PR though 😓

Issue #1450, opened 2021-09-01
Fix ❌  on `test-develop`


minusworld:
**Describe the bug**
The rule examples tar file is not being generated. Investigate why.
This does not appear to be an issue locally, so it may be a change in GHA.

**To Reproduce**
Run any `test-develop` action in a PR.

**Expected behavior**
--> ✅  instead

**Priority**
How important is this to you?
- [x] P0: blocking me from making progress
- [ ] P1: this will block me in the near future
- [ ] P2: annoying but not blocking me

PR #1451, opened 2021-09-01
Fix test-develop action

minusworld:


minusworld:
Closes https://github.com/returntocorp/semgrep-rules/issues/1450

colleend (approved):
we don't have any rules that use pattern-where-python anymore right?

Edit: Confirmed this with a quick search!

colleend:
thanks for doing the investigating!

minusworld:
Right

PR #1455, opened 2021-09-02
Detect @Html.Raw on JSON

Sjord:
A common pattern in ASP.NET is to convert a backend C# object to JavaScript,
using the following pattern in the Razor template:

```
<script>
var a = @Html.Raw(Json.Encode(Model));
</script>
```

This encodes `Model` to JSON and assigns it to `a`. The `Html.Raw` is used to
avoid HTML encoding. Unfortunately, this also makes it vulnerable to XSS. By
including `</script>` in the JSON, it is possible to break out of the script
context.

This rule detects this pattern, the combination of `@Html.Raw` with JSON
encoding. I chose not to match the surrounding `<script>`. This makes it more
specific where the error occurs, and this pattern would also be vulnerable
outside of JavaScript.

### DIFF
--- a/csharp/razor/security/html-raw-json.cshtml
+++ b/csharp/razor/security/html-raw-json.cshtml
@@ -1,0 +1,20 @@
+<script>
+    // ruleid: html-raw-json
+    var a = @Html.Raw(SomeFunc(Model).ToJson());
+</script>
+<script type="text/javascript">
+    SomeOtherCall();
+    // ruleid: html-raw-json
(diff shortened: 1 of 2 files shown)


Sjord:
@inkz, could you take a look at this? Is this file in the right place at `csharp/razor/security`?

inkz (approved):
lgtm!
@Sjord yes, it makes sense to put it into `csharp/razor/security` :+1:

Issue #1390, opened 2021-07-23
False positive for return-in-init when return in internal function


Sjord:
**Describe the bug**

[`return-in-init`](https://github.com/returntocorp/semgrep-rules/blob/master/python/lang/correctness/return-in-init.yaml) warns about a return statement in `__init__`. However, this may be valid if another function is defined within `__init__` and return is used there.

**To Reproduce**

```
class Odd:
    def __init__(self, numbers):
        def is_odd(n):
            return n % 2 == 1
        self.numbers = filter(is_odd, numbers)
```

```
$ semgrep --config=p/ci
test1.py
severity:error rule:python.lang.correctness.return-in-init.return-in-init: `return` should never appear inside a class __init__ function. This will cause a runtime error.
4:            return n % 2 == 1
```

**Expected behavior**

I expect no error from `return-in-init` in this case.

**Priority**
How important is this to you?
- P2: annoying but not blocking me

Sjord:
[Here is another false positive](https://github.com/html5lib/html5lib-python/blob/master/html5lib/treebuilders/etree_lxml.py#L193-L228), where a whole class is defined in the `__init__`.

Sjord:
I have created some test cases:

```
class Odd:
    def __init__(self, numbers):
        def is_odd(n):
            # ok:return-in-init
            return n % 2 == 1
        self.numbers = filter(is_odd, numbers)
        
        # ruleid:return-in-init
        return self.numbers

class Even:
    def __init__(self):
        class EvenNumber:
            def __init__(self, n):
                self.n = n
                # ruleid:return-in-init
                return n
            
            def is_even(self):
                # ok:return-in-init
                return self.n % 2 == 0
        
        self.number = EvenNumber()
    
    def not_init(self):
        class EvenNumber:
            def __init__(self, n):
                self.n = n
                # ruleid:return-in-init
                return n
            
            def is_even(self):
                # ok:return-in-init
                return self.n % 2 == 0
        
        # ok:return-in-init
        return EvenNumber()
```

I have attempted to solve this, but I can't think of a pattern that works perfectly. I tried variations on the following, but this no longer matches anything.

```
  - pattern-not-inside: |
      def $METHOD(...):
          ...
  - metavariable-pattern:
      metavariable: $METHOD
      patterns:
        - pattern-not: __init__
```

I tried this, but this is not valid:

```
  - patterns-either:
    - pattern-inside: |
        def __init__(self):
          ...
          def __init__(...):
[...]

PR #1456, opened 2021-09-06
python-lang-correctness: allow return in inner functions

staaldraad:
Python allows inner functions, meaning you could have a return within an inner function, inside a `__init__`. 
The current rule falsely flags this as an Error.

With this change, the following will be allowed:

```python
class I:
    def __init__(self, x):
        def internal_func():
          # ok:return-in-init
          return 5
```

Inner functions info: https://realpython.com/inner-functions-what-are-they-good-for/


macoMarv86 (approved):
Duplicate of  #1390

staaldraad:
🙌🏼 yay for synchronicity.

Closing

PR #1457, opened 2021-09-07
Don't report return-in-init if there's another function in the __init__

### LINKED ISSUE #1390: False positive for return-in-init when return in internal function

Sjord:
To prevent false positives, ignore returns in inner functions.

Closes #1390.


python/lang/correctness/return-in-init.py
```diff
@@ -75,3 +75,41 @@ class H:
...
         return None
+
+class Odd:
+    def __init__(self, numbers):
+        def is_odd(n):
+            # ok:return-in-init
+            return n % 2 == 1
+        self.numbers = filter(is_odd, numbers)
+
+        # ruleid:return-in-init
```
IagoAbal: I guess this would be a `todoruleid` because the new rules won't find any matches if you define a nested function inside `__init__`.

inkz (approved):
lgtm

python/lang/correctness/return-in-init.py
inkz: :+1: I changed to `todoruleid`

PR #1474, opened 2021-09-15
Modify semgrep rule lints

minusworld:
- Modify `owasp` metadata rule lint to handle lists for multiple OWASP years
- Modify semgrep-rule-lints workflow to use local files


inkz (approved):
:+1:

PR #1478, opened 2021-09-16
Update GHA run-shell-injection rule after triaging real-world results

minusworld:
Uses a list of user input for run-shell-injection rule defined in https://securitylab.github.com/research/github-actions-untrusted-input/. This is to reduce false positives


inkz (approved):
lgtm

PR #1482, opened 2021-09-17
Missing SRI attribute rule

jsherm-fwdsec:
- Finds script and link tags in HTML files that point to external files and are missing the integrity attribute as per ASVS item number 14.2.3 (https://github.com/OWASP/ASVS/blob/master/4.0/en/0x22-V14-Config.md#v142-dependency)
- Test cases added


minusworld (approved):
Very nice! Thank you @jsherm-fwdsec!

jsherm-fwdsec:
Awesome thanks!

PR #1487, opened 2021-09-24
 add PHP rules for Symfony (part 2) 

inkz:


php/lang/security/non-literal-header.yaml
```diff
@@ -0,0 +1,18 @@
+rules:
+- id: non-literal-header
+  patterns:
+  - pattern: header(...)
+  - pattern-not: header("...",...)
+  message: >-
+    Using user input when deleting files with `unlink()` is potentially dangerous. A malicious actor could use this to modify
```
minusworld: I think this message is incorrect?

minusworld (changes requested):
Rule message

php/lang/security/non-literal-header.yaml
inkz: Ooops :sweat_smile: thanks a lot for finding it! updated :+1:

php/lang/security/ldap-bind-without-password.php
```diff
@@ -0,0 +1,36 @@
+<?php
+
+$ldapconn = ldap_connect("foo.com");
+
+// ruleid: ldap-bind-without-password
+$ldapbind = ldap_bind($ldapconn);
+
+// todoruleid: ldap-bind-without-password
+LDAP_BIND($ldapconn, "username");
```
minusworld-gha-demo, line 9: @inkz What's would it take for the PHP parser to match this test case?

minusworld-gha-demo (approved):

minusworld (approved):

PR #1492, opened 2021-09-28
[Ruby] fix for https://github.com/returntocorp/semgrep/pull/3951

aryx:
With https://github.com/returntocorp/semgrep/pull/3951
we consider identifier on their own on a line as an hidden
function call. This introduced a regression in semgrep-rules
which can be fixed by being more precise.

test plan:
make test


colleend (approved):
nice, lgtm!

PR #1494, opened 2021-09-28
converting terraform generic iam rules to hcl

colleend:
Converting so that these work in HCL!

### DIFF
--- a/terraform/lang/security/iam/no-iam-data-exfiltration.yaml
+++ b/terraform/lang/security/iam/no-iam-data-exfiltration.yaml
@@ -64,7 +64,7 @@ rules:
           "rds:*"
       - pattern: |
           "rn: secretsmanager:*"
-  message: |
+  message: >-
     Ensure that IAM policies don't allow data exfiltration actions that are not resource-constrained.
     This can allow the user to read sensitive data they don't need to read.
     Instead, make sure that the user granted these privileges are given these permissions on specific resources.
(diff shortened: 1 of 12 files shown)


terraform/lang/security/iam/no-iam-data-exfiltration.yaml
```diff
@@ -61,8 +74,7 @@ rules:
     - https://cloudsplaining.readthedocs.io/en/latest/glossary/data-exfiltration/
     category: security
     license: Commons Clause License Condition v1.0[LGPL-2.1-only]
-  paths:
-    include:
-    - '*.tf'
-  languages: [generic]
+    technology:
```
inkz, line 78: is it possible to add CWE number here?
(I would propose smth like https://cwe.mitre.org/data/definitions/200.html, but you know better)

inkz (changes requested):
add CWE numbers to some rules (e.g. `no-iam-data-exfiltration`, `no-iam-priv-esc-funcs`)

if it actually make sense to do it :v:, because rules are IaC related maybe it is not applicable

aryx (commented):
Woohoo!

colleend:
sounds good, I can look into it and add the appropriate CWE #s to some of the rules!

minusworld (approved):

inkz (approved):
lgtm!

PR #1502, opened 2021-09-30
Improve detection for host header injection

vin01:
Hi, I noticed that current recommendation for host header injection prevention is not correct.

> Use the '$host' variable instead, which will use server names listed in the 'server_name' directive

From [docs](http://nginx.org/en/docs/http/ngx_http_core_module.html#var_host):

> $host
> in this order of precedence: host name from the request line, or host name from the “Host” request header field, or the server name matching a request

Request header is preferred *even* if server_name is present. Current message from the test is misleading and may result in insecure configurations while giving a false impression that value of `$host` can be trusted in backend applications.

Open issue in gixy: https://github.com/yandex/gixy/issues/93

### DIFF
--- a/generic/nginx/security/request-host-used.conf
+++ b/generic/nginx/security/request-host-used.conf
@@ -7,3 +7,13 @@ server {
     proxy_pass http://backend;
   }
 }
+
+server {
+  listen 80;
+
+  location @app {
+    # ruleid: request-host-used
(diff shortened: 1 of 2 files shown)


minusworld:
Good find! Thanks @vin01 !

minusworld (approved):

PR #1510, opened 2021-10-01
Remove top-ellipses from various rules for performance

minusworld:
@inkz Could you take a look at these to see if I unintentionally broke any behavior? Thanks 🙏  :D


inkz (approved):
:+1:

PR #1518, opened 2021-10-05
update jwt-decode-without-verify rule

inkz:
So after some experiments I want to return top ellipsis to `jwt-decode-without-verify` rule.
Main issue is that in a lot of scenarios token is first decoded to get for example expiration time and then if everything is ok token is verfied, so both `verify then decode` and `decode then verify` patterns exist and we want to find when only `decode` is used.

@minusworld  Should I add smth like `nosem` comment to the rule now?


minusworld:
@inkz sgtm

inkz:
@minusworld green light :green_circle:

minusworld (approved):

PR #1521, opened 2021-10-07
Filter multiple CMDs when they appear in different images

### LINKED ISSUE #1520: Dockerfile rules - false positive on multi-stage build with CMDs
`generic.dockerfile.correctness.multiple-cmd-instructions.multiple-cmd-instructions`

**Describe the bug**
A clear and concise description of what the bug is and for which rule. Is the rule producing a false positive? Is it incorrect or missing something?

https://semgrep.dev/r?q=generic.dockerfile.correctness.multiple-cmd-instructions.multiple-cmd-instructions is flagging multiple CMD instructions, even when those CMDs are each in a different stage of the build.

**To Reproduce**

```Dockerfile
FROM busybox AS stage1
CMD /bin/true
FROM stage1 AS stage2
CMD /bin/false
```

**Expected behavior**

This rule should only fire if there's two `CMD`s within a stage (between `FROM` instructions).

**Priority**
How important is this to you?

minusworld:
Fixes https://github.com/returntocorp/semgrep-rules/issues/1520


colleend (approved):
lgtm!

PR #1522, opened 2021-10-08
Filter hardcoded strings from Ruby deserialization rule

minusworld:


minusworld:
Faulty understanding - needs to filter out static strings to File.open(...) instead

PR #1525, opened 2021-10-11
Updated java/lang/security/audit/object-deserialization.yaml with actionable reference on how to fix

thedevappsecguy:
Updated metadata with actionable reference which is having a broad set of guidelines ,so that this can be remediated according to the context generally faced by java developers.


minusworld (approved):

minusworld:
@thedevappsecguy Thank you! :D

PR #1528, opened 2021-10-12
Updated java/lang/security/audit/crypto/gcm-detection.yaml with a clear rule message,CWE

thedevappsecguy:
Updated java/lang/security/audit/crypto/gcm-detection.yaml with a clear rule message and CWE which maps to Nonce reuse.


thedevappsecguy:
Some problem with Validation will create a pull request again post validation.

Issue #1523, opened 2021-10-08
Consider filtering `.to_json` from Ruby HTML unescape rules


minusworld:
**Describe the bug**
`.to_json` inside unescaped sections in ERB templates may be safe. Investigate in depth if this can be filtered out.

Example:

```
<script>
  var value = <%== @some_var.to_json %>
</script>
```

**Priority**
How important is this to you?
- [ ] P0: blocking me from making progress
- [x] P1: this will block me in the near future
- [ ] P2: annoying but not blocking me

PR #1529, opened 2021-10-12
Consider html_safe alias plus .to_json a safe pattern

bernardoamc:
This happens because by default Rails escape unicode characters.


minusworld (approved):

minusworld:
Closes https://github.com/returntocorp/semgrep-rules/issues/1523

PR #1530, opened 2021-10-12
Updated java/lang/security/audit/crypto/gcm-detection.yaml with a clear rule message

thedevappsecguy:
Updated java/lang/security/audit/crypto/gcm-detection.yaml with a clear rule message and CWE which maps to Nonce reuse.


thedevappsecguy:
Some issue with my editors and AV ,I will close this PR and open clean PR.

Issue #1524, opened 2021-10-08
[Ruby] Hardcoded file paths into deserialization functions


minusworld:
**Describe the bug**
Filter out cases of static strings from `File.open(...)` to deserialization functions like `YAML.load(...)`.

**Priority**
How important is this to you?
- [ ] P0: blocking me from making progress
- [ ] P1: this will block me in the near future
- [x] P2: annoying but not blocking me

colleend:
👋  do you mean `File.read(...)` in this case?

PR #1547, opened 2021-10-15
Eliminate FPs from static file reads into YAML.loads

colleend:
For issue https://github.com/returntocorp/semgrep-rules/issues/1524. 

This eliminates some possible false positives from static files read into the `YAML.load` function.


minusworld (approved):
\o/

PR #1548, opened 2021-10-18
update useless-innerfunction rule

### LINKED ISSUE #563: python.lang.maintainability.useless-innerfunction.useless-inner-function false positive
**Describe the bug**
With pytest-benchmark, you can define benchmarks like this:

```python
def some_test(benchmark):
  @benchmark
  def func():
     time.sleep(5)
```

This produces: 
```
semgrep python.lang.maintainability.useless-innerfunction.useless-inner-function
⚠ *** finding(s) due to staged changes in 9.82 s
     > semgrep/tests/performance/test_ci_perf.py:48
```

**Expected Result**
I'm not sure the perfect behavior, but probably in general, we shouldn't trigger `useless-inner-function` if the function has a decorator on it.

inkz:
Closes https://github.com/returntocorp/semgrep-rules/issues/563
I believe it is safe to assume that if a function has a decorator we can skip it because there is a high chance of FP

### DIFF
--- a/python/lang/maintainability/useless-innerfunction.py
+++ b/python/lang/maintainability/useless-innerfunction.py
@@ -21,6 +21,11 @@ def A():
     def C():
         print_error('another')
 
+    # ok:useless-inner-function
+    @something
+    def D():
+        print_error('with decorator')
+
     return B(), C()
 
 def foo():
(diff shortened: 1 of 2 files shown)


colleend (approved):
makes sense to me 👍

PR #1551, opened 2021-10-21
Add first bash rules! Bad things with curl

minusworld:


mjambon (approved):
cool! I'm not sure what the errors are. Let me know if it looks like something I should fix.

PR #1553, opened 2021-10-22
Add rules to check for unquoted variable expansions

mjambon:
These two rules as is should already be useful but
* I'll need to fix a bug in semgrep that will allow catching an unquoted `foo$bar` (https://github.com/returntocorp/semgrep/issues/4121).
* We need to write a similar rule to check for unquoted command substitution (`$(...)` and `` `...` ``).

The rule I called "IFS tampering" seemed natural and is a prerequisite to tolerating special expansions like `$$` (process ID). An unquoted `$$` is normally safe because it's a number, but if IFS is set to split on a digit, weird things will happen:
```
$ echo $$
318248
$ IFS=8
$ echo $$
31 24
```

### DIFF
--- a/bash/lang/correctness/unquoted-expansion.bash
+++ b/bash/lang/correctness/unquoted-expansion.bash
@@ -1,0 +1,58 @@
+# ruleid: unquoted-variable-expansion-in-command
+exec $foo
+
+# ruleid: unquoted-variable-expansion-in-command
+exec $FOO
+
+# ruleid: unquoted-variable-expansion-in-command
+exec ${foo}
+
+# ruleid: unquoted-variable-expansion-in-command
+exec $1
+
+# ruleid: unquoted-variable-expansion-in-command
+exec ${foo%.bar}
+
+# ruleidXXX: unquoted-variable-expansion-in-command
+#exec $foo.bar
+
+# ruleidXXX: unquoted-variable-expansion-in-command
+#exec ${foo}.bar
+
+# ok: unquoted-variable-expansion-in-command
+exec "$foo"
+
+# ok: unquoted-variable-expansion-in-command
+exec "$FOO"
+
+# ok: unquoted-variable-expansion-in-command
+exec "${foo}"
+
+# ok: unquoted-variable-expansion-in-command
+exec "$1"
+
+# ok: unquoted-variable-expansion-in-command
+exec "${foo%.bar}"
+
+# ok: unquoted-variable-expansion-in-command
+exec "${foo}.bar"
+
+# ok: unquoted-variable-expansion-in-command
+exec "${foo}".bar
+
+# ok: unquoted-variable-expansion-in-command
+exec "$foo".bar
+
+# ok: unquoted-variable-expansion-in-command
+x=$foo
+
+# ok: unquoted-variable-expansion-in-command
+PATH=$foo:$PATH bar
+
+# Expands without splitting (if IFS wasn't tempered with)
+# ok: unquoted-variable-expansion-in-command
+echo $$
+
+# Special exception for semgrep users: $_foo is tolerated.
+# ok: unquoted-variable-expansion-in-command
+echo $_foo
--- a/bash/lang/correctness/unquoted-expansion.bash
+++ b/bash/lang/correctness/unquoted-expansion.bash
@@ -13,11 +13,11 @@ exec $1
 # ruleid: unquoted-variable-expansion-in-command
 exec ${foo%.bar}
 
-# ruleidXXX: unquoted-variable-expansion-in-command
-#exec $foo.bar
+# ruleid: unquoted-variable-expansion-in-command
+exec $foo.bar
 
-# ruleidXXX: unquoted-variable-expansion-in-command
-#exec ${foo}.bar
+# ruleid: unquoted-variable-expansion-in-command
+exec ${foo}.bar
 
 # ok: unquoted-variable-expansion-in-command
 exec "$foo"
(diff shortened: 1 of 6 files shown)


mjambon:
Not sure what's going on. The `ifs-tampering` rule works fine but not the other one:
```
(TODO: 0) /github/workspace/bash/lang/correctness/unquoted-expansion.yaml
	✖ unquoted-variable-expansion-in-command                       TP: 0 TN: 12 FP: 0 FN: 7
	test: /github/workspace/bash/lang/correctness/unquoted-expansion.bash, expected lines: [2, 5, 8, 11, 14, 17, 20], reported lines: []
```

Super weird. It also stopped working locally for me.

mjambon:
Apparently I lost a `$`: `${$VAR}` somehow became `${VAR}`.

mjambon:
CI is failing and I don't know why:
```
Run returntocorp/semgrep-rules-test-action@master
/usr/bin/docker run --name f88420a2e2d7392be2450c8969a4c60dd25477_dad6b9 --label f88420 --workdir /github/workspace --rm -e HOME -e GITHUB_JOB -e GITHUB_REF -e GITHUB_SHA -e GITHUB_REPOSITORY -e GITHUB_REPOSITORY_OWNER -e GITHUB_RUN_ID -e GITHUB_RUN_NUMBER -e GITHUB_RETENTION_DAYS -e GITHUB_RUN_ATTEMPT -e GITHUB_ACTOR -e GITHUB_WORKFLOW -e GITHUB_HEAD_REF -e GITHUB_BASE_REF -e GITHUB_EVENT_NAME -e GITHUB_SERVER_URL -e GITHUB_API_URL -e GITHUB_GRAPHQL_URL -e GITHUB_WORKSPACE -e GITHUB_ACTION -e GITHUB_EVENT_PATH -e GITHUB_ACTION_REPOSITORY -e GITHUB_ACTION_REF -e GITHUB_PATH -e GITHUB_ENV -e RUNNER_OS -e RUNNER_NAME -e RUNNER_TOOL_CACHE -e RUNNER_TEMP -e RUNNER_WORKSPACE -e ACTIONS_RUNTIME_URL -e ACTIONS_RUNTIME_TOKEN -e ACTIONS_CACHE_URL -e GITHUB_ACTIONS=true -e CI=true -v "/var/run/docker.sock":"/var/run/docker.sock" -v "/home/runner/work/_temp/_github_home":"/github/home" -v "/home/runner/work/_temp/_github_workflow":"/github/workflow" -v "/home/runner/work/_temp/_runner_file_commands":"/github/file_commands" -v "/home/runner/work/semgrep-rules/semgrep-rules":"/github/workspace" f88420:a2e2d7392be2450c8969a4c60dd25477

====== BEGIN STDOUT ======
semgrep --validate --config=$PWD/python $PWD
semgrep --validate --config=$PWD/c $PWD
semgrep --validate --config=$PWD/javascript $PWD
semgrep --validate --config=$PWD/java $PWD
semgrep --validate --config=$PWD/go $PWD
semgrep --validate --config=$PWD/ocaml $PWD
semgrep --test --strict --test-ignore-todo --quiet $PWD
====== END STDOUT ======
====== BEGIN STDERR ======
METRICS: Using configs from the Registry (like --config=p/ci) reports pseudonymous rule metrics to semgrep.dev.
To disable Registry rule metrics, use "--metrics=off".
Using configs only from local files (like --config=xyz.yml) does not enable metrics.

More information: https://semgrep.dev/docs/metrics

Configuration is valid - found 240 valid configuration(s), 0 configuration error(s), and 294 rule(s).
Configuration is valid - found 14 valid configuration(s), 0 configuration error(s), and 14 rule(s).
Configuration is valid - found 144 valid configuration(s), 0 configuration error(s), and 162 rule(s).
Configuration is valid - found 87 valid configuration(s), 0 configuration error(s), and 94 rule(s).
Configuration is valid - found 61 valid configuration(s), 0 configuration error(s), and 67 rule(s).
Configuration is valid - found 14 valid configuration(s), 0 configuration error(s), and 24 rule(s).
make: *** [Makefile:8: test] Error 1
====== END STDERR ======
```
https://github.com/returntocorp/semgrep-rules/runs/3972507155?check_suite_focus=true

mjambon:
ok, so I added a rule for catching unquoted `$(...)` and `` `...` ``. It works on my machine, which uses the latest `develop` branch of semgrep. The output of CI now looks normal, i.e. it shows the status of each test. Some tests are failing. For now, I'll assume it's because it uses an older version of semgrep.

mjambon:
I just re-ran the jobs. Now "test-develop" works :tada: but "test-latest" still fails. 
image

mjambon:
In some cases, an unquoted expansion is kinda safe (e.g. `eval`) but also requires advanced understanding by the reader of the code, which probably shouldn't be assumed. I was about to suggest making an exception for `eval`, since all its arguments are concatenated with spaces anyway, but actually it doesn't work. The following will print `a b` instead of `a`[6 spaces]`b`:

```bash
$ prog='echo "a      b"'
$ eval $prog
a b
```

So no, even the argument of `eval` should be quoted:

```bash
$ prog='echo "a      b"'
$ eval "$prog"
a      b
```

mjambon:
I see that `test-latest` uses the docker image `returntocorp/semgrep:latest`, which is tied to the latest official release.

minusworld:
@mjambon I assume there's a fix in `develop` that correctly finds all the tests. Once `0.71` is released this week, we can check this again.

minusworld (approved):
lgtm

mjambon:
Now that 0.71 is released, the "latest" tests pass. Merging.

PR #1556, opened 2021-10-22
PD-698 Add yaml rule: semgrep-github-action-push-without-branches

TinMarkovic:
Details at: https://linear.app/r2c/issue/PD-698/write-semgrep-rules-to-flag-semgrep-ci-misconfigurations


TinMarkovic:
Hello @emjin and @chmccreery - help out please :) 

Clara, I'd welcome a second review. I've implemented both suggestions.

Emma, I'd like to ask you for help to "optimize" this rule. Not sure how else to detect the `returntocorp/semgrep-action@v1` string, which Clara was nice enough to point out as a strong signal we're doing a semgrep scan via the action.

Finally, if either knows if I can:
- Ignore the YAML lint for duplicate keys in test file
- Use 2 (or more!) test files at once

The test is unwieldy as top-level YAML is an object itself, so for a test of a "full yaml object", I need separate files... or just to tolerate the hack.

TinMarkovic:
> Not sure how else to detect the `returntocorp/semgrep-action@v1` string, which Clara was nice enough to point out as a strong signal we're doing a semgrep scan via the action.

I've added a commit that does it a different way, by utilizing `path-include`. Might be a better option, might not... listening for opinions :)

minusworld:
This is a cool use case :D

I think the path include is perfect for this.

If I understand your intent correctly, you may also be able to simplify the rule like this:

```yaml
rules:
  - id: semgrep-github-action-push-without-branches
    message: The 'branches' field (in the push event configuration) contains no
      branches. This causes all branches to be scanned and may result in
      unneccessary duplicate findings across the entire codebase.
    severity: WARNING
    languages:
      - yaml
    paths:
      include:
        - ".github/workflows/semgrep.yml"
    patterns:
    - pattern-inside: |
        on:
          ...
          push:
            ...
    - pattern: |
        branches:
    - pattern-not-inside: |
        branches: [...]
    metadata:
      category: correctness
      technology:
        - github-action
```

This will scope everything to detections only within `on: push:`, and I believe will flag all `branches:` that **do not** have any items in them. lmk if this helps!

https://semgrep.dev/s/Oepb

TinMarkovic:
Thanks Grayson, but it then doesn't go through the other ways to define `push` without scoping any branches to it, causing a lot of (unnecessary) findings.

The other approaches I've added specifically to cover all cases.

minusworld:
Cool, that works - I can fix pre-commit.

minusworld (approved):

chmccreery:
We should probably add this to the "ci" pack, and maybe also just default-add it to people's nolicy-audit policies. What do you all think?

minusworld:
Totally agree! We should definitely apply Semgrep rules we know will be useful out-of-the-box to all new users. Same with GitHub Actions.

Would you prefer to add it in-app, or add it to the CI pack?

chmccreery:
I think both. Just chatted with @underyx and we think the best way forward is to create a ruleset called "semgrep-misconfigurations" with just this rule for now (so we can add more rules in the future as they come up). Do you think you could do that for us @minusworld? Then I'd add that ruleset to the default nolicy (or even blocker) column.

minusworld:
Sure thing - I can get someone to do it. Would you mind making tickets on https://linear.app/r2c/team/RULES/triage?

chmccreery:
@minusworld done! https://linear.app/r2c/issue/RULES-222/create-semgrep-misconfigurations-ruleset

PR #1557, opened 2021-10-22
Add actual taint example for bash

minusworld:


mjambon:
Regarding the rule: it should also handle quoted cases.
The patterns [right now are](https://github.com/returntocorp/semgrep-rules/blob/ea0e831e6fbf9806d272d7b531bcecea9a0f0f66/bash/curl/security/curl-eval.yaml#L23-L24):
```
    - pattern: $DATA=$(curl ...)
    - pattern: eval $(curl ...)
```
Their quoted versions will be common and are not treated as equivalent by semgrep, so I suggest adding them:
```
    - pattern: $DATA=$(curl ...)
    - pattern: $DATA="$(curl ...)"
    - pattern: eval $(curl ...)
    - pattern: eval "$(curl ...)"
```

Independently, an unquoted `$(curl ...)` in a command (right-hand-side of the assignment is fine) isn't recommended even though it's usually fine in the case of `eval` because all its arguments are concatenated back (as long as nobody messed with `IFS`, which also has its own rule brought by #1553 ). I'll add to the discussion on #1553.

bash/curl/security/curl-eval.yaml
```diff
@@ -21,6 +21,8 @@ rules:
   pattern-sources:
   - pattern-either:
     - pattern: $DATA=$(curl ...)
```
IagoAbal, line 23: Ideally the pattern should only be `curl ...` or `$(curl ...)` if that is not possible. Then you shouldn't need both unquoted and quoted versions.

IagoAbal: That is:
```
pattern-sources:
- pattern: curl ...
```
or if that is not possible then
```
pattern-sources:
- pattern: $(curl ...)
```

colleend (approved):
lgtm!

bash/curl/security/curl-eval.yaml
colleend: I think grayson used `$DATA=$(curl ...)` and `eval curl ...` to avoid false positives -- @minusworld probably has more context on this than me though 😅

minusworld: Since this is just updating the test case, I think it's fine for now - I can check to see if Iago's suggestions work as expected at a later time

mjambon: Note that new features have been added in the last two weeks:
* simple expression matching: the pattern `foo` now matches in `cmd foo`. No need for the pattern to be `... foo ...`.
* deep ellipsis: pattern `$DATA=<... $(curl ...) ...>` matches `x="foo $(curl url)"`.

PR #1560, opened 2021-10-25
Fix bug in Go rule tainted-url-host which assumed concatenation inclu…

minusworld:
…ded format string parameters

### DIFF
--- a/go/lang/security/injection/tainted-url-host.yaml
+++ b/go/lang/security/injection/tainted-url-host.yaml
@@ -2,13 +2,10 @@ rules:
 - id: tainted-url-host
   languages:
   - go
-  message: >-
-    User data flows into the host portion of this manually-constructed URL.
-    This could allow an attacker to send data to their own server, potentially exposing
-    sensitive data such as cookies or authorization information sent with this request.
-    They could also probe internal servers or other resources that the server runnig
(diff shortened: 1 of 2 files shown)


go/lang/security/injection/tainted-url-host.yaml
```diff
@@ -2,13 +2,10 @@ rules:
...
   languages:
   - go
-  message: >-
-    User data flows into the host portion of this manually-constructed URL.
-    This could allow an attacker to send data to their own server, potentially exposing
-    sensitive data such as cookies or authorization information sent with this request.
-    They could also probe internal servers or other resources that the server runnig
-    this code can access. (This is called server-side request forgery, or SSRF.) Do
-    not allow arbitrary hosts. Instead, create an allowlist for approved hosts hardcode
+  message: User data flows into the host portion of this manually-constructed URL. This could allow an attacker to send data
```
inkz, line 5: why linter passes it? :thinking: 
https://github.com/returntocorp/semgrep-rules/blob/develop/yaml/semgrep/multi-line-message.yaml

minusworld: Oh, I changed the message too.

I don't think the metalinter detects `message:` that are on the same line.

colleend (approved):
lgtm!

PR #1561, opened 2021-10-26
Rule for md5 used as a password hashing algorithm in Java

minusworld:
This is an idea I had while scanning a bunch of open source code.
This rule checks if MD5 reaches a method with `password` in the name, giving it a higher indication that it is used as a password hash.

This is a little more precise than the usual "flag every md5 occurrence".

Preliminary scans of OSS show no findings, so I theorize it will not produce a ton of noise.


inkz (approved):
:+1:

PR #1562, opened 2021-10-26
Kotlin rules

colleend:
Translating some java findsecbug rules to kotlin. With this, we'll have 10 kotlin rules (which is the standard for beta).


minusworld (approved):
lgtm

java/lang/security/audit/crypto/rsa-no-padding.yaml
```diff
@@ -14,8 +14,11 @@ rules:
...
     technology:
     - java
+    - kotlin
   message: >-
     Using RSA without OAEP mode weakens the encryption.
   severity: WARNING
-  languages: [java]
+  languages:
+  - java
+  - kt
```
aryx, line 23: nice :) excellent that we can just reuse the same rule.
cc @nbrahms

kotlin/lang/security/desede-is-deprecated.yaml
```diff
@@ -0,0 +1,23 @@
...
+      control_id: 6.2.5 Insecure Algorithm
+      control_url: https://github.com/OWASP/ASVS/blob/master/4.0/en/0x14-V6-Cryptography.md#v62-algorithms
+    references:
+    - https://csrc.nist.gov/News/2017/Update-to-Current-Use-and-Deprecation-of-TDEA
+    category: security
+    technology:
+    - kotlin
+    license: Commons Clause License Condition v1.0[LGPL-2.1-only]
+  severity: WARNING
+  pattern: $CIPHER.getInstance("=~/DESede.*/")
```
aryx, line 20: why we could not factorize this rule with the similar one for Java?

colleend: the java ones have semicolons at the end, which doesn't match some kotlin files (as semicolons are optional for kotlin)

aryx: It should not matter. First, you should remove the semicolon in the java pattern, it's more general because
it then looks for any expressions, not just expressions-statements. Then even if the pattern use the Java syntax
which has semicolons, it can still be applied on other languages. In theory it could apply to any language because they all convert to the generic AST. What matters in the language list is the first language, which is the one used to parse the actual pattern in the file, but after that it's converted to the generic AST and so any other language can be used.

colleend: hmm, I see -- makes sense that we can remove the semicolon in the java pattern -- we might want to write a script in order to do that for all patterns that have the semicolon at the end where we're searching for expressions (I can create an issue for that in the backlog). 

Interestingly enough, I tried running the kotlin rule with semicolons and it error'ed: https://semgrep.dev/s/y1wO. This probably isn't expected behavior -- should I file an issue?

aryx: Yes the spec says you can have semicolons, but in practice there seems to be no kotlin code out there using semicolons, so tree-sitter-kotlin has this _semis rules which actually accept newlines but not semicolons.

colleend: I see, very interesting 👍 . Looked through some more Kotlin examples, very true that in practice there's no use of semicolons. We don't need to worry about the semicolon behavior then!

PR #1569, opened 2021-10-27
Filter constant strings from select clauses in Django queryset extra

### LINKED ISSUE #575: Ignore cases where constant strings are provided to django queryset `extra`
Example failing test cases: https://semgrep.live/gxJK

Observed while scanning https://github.com/DefectDojo/django-DefectDojo/blob/de98f281165dedffb27a0a394fb0bbb66ed0e72e/dojo/metrics/views.py#L57-L69

Having dataflow here would help a lot.

minusworld:
Fixes https://github.com/returntocorp/semgrep-rules/issues/575

### DIFF
--- a/python/django/security/audit/query-set-extra.yaml
+++ b/python/django/security/audit/query-set-extra.yaml
@@ -1,7 +1,9 @@
 rules:
 - id: avoid-query-set-extra
-  message: QuerySet.extra does not provide safeguards against SQL injection and requires very careful use. Therefore, it is
-    advisable not to use .extra and to instead use the Django ORM. See https://docs.djangoproject.com/en/3.0/ref/models/querysets/#django.db.models.query.QuerySet.extra
+  message: >-
+    QuerySet.extra does not provide safeguards against SQL injection and requires very careful use. Therefore, it is
+    advisable not to use .extra and to instead use the Django ORM. See 
+    https://docs.djangoproject.com/en/3.0/ref/models/querysets/#django.db.models.query.QuerySet.extra
     for more details.
   metadata:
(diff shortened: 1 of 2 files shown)


python/django/security/audit/query-set-extra.yaml
```diff
@@ -1,18 +1,24 @@
...
-  severity: ERROR
-  pattern-either:
-  - pattern: $X.objects.$FUNC(...).extra(...)
-  - pattern: $X.objects.$FUNC(...).$FILTER(...).extra(...)
-  - pattern: $X.objects.$FUNC(...).$FILTER(...).$UPDATE(...).extra(...)
+  languages:
+  - python
+  severity: WARNING
+  patterns:
+  - pattern: $MODEL.extra(...)
```
inkz, line 23: maybe add smth like:
```
  - metavariable-regex:
      metavariable: $MODEL
      regex: .*objects.*
```
to filter non `QuerySet`s like (e.g. `foobar.extra('yo')`)

minusworld: I think that you can use `.extra` on querysets without the `objects` property. See https://docs.djangoproject.com/en/3.2/ref/models/querysets/#extra

I guess I should check to see how common `.extra` is and if it's prone to FPs.

colleend (approved):
lgtm, @inkz is this ok with you?

python/django/security/audit/query-set-extra.yaml
inkz: oh ok, I get it now :+1:

inkz:
> lgtm, @inkz is this ok with you?

yes!

PR #1571, opened 2021-10-27
add node.js tainted-sql-string rule

inkz:
tainted sql string for Express


inkz:
fix `todo` when https://github.com/returntocorp/semgrep/issues/4160 is resolved

minusworld (approved):

PR #1574, opened 2021-10-27
Add rule to detect Kolide API keys

blaedj:
This commit:
 - adds a rule to detect [Kolide](https://k2.kolide.com) api keys, as
 documented at https://kolidek2.readme.io/docs/kolide-api-token-format


minusworld (approved):
Sweet, thanks! :D

PR #1580, opened 2021-10-28
Agglomerating Kotlin and Java Rules

colleend:
Removing some semicolons from java rules so they match expressions, not just expression statements. This also lets us agglomerate kotlin and java rules more.


inkz (approved):
wow, that is awesome!

aryx (approved):

PR #1581, opened 2021-10-29
Fix cwe, owasp formats and remove duplicate patterns

emjin:
Have these rules pass the `p/semgrep-rule-lints` checks


inkz (approved):
lgtm!

PR #1582, opened 2021-10-29
[Kotlin] remove semicolon to fix parse error

aryx:
This prevents https://github.com/returntocorp/semgrep/pull/4174 to
pass

test plan:
make test


IagoAbal (approved):

colleend:
good catch!

PR #1583, opened 2021-10-30
Add useless-cat rule for bash, similar to shellcheck's SC2002

mjambon:
See https://github.com/koalaman/shellcheck/wiki/SC2002


mjambon:
> semgrep-rules-test / test-latest (pull_request) Cancelled after 360m — test-latest

mmm, that's quite a long time. Re-running.

colleend:
hmm, seems like its still taking longer than 360 min -- does `make test` work locally for you?

mjambon:
> hmm, seems like its still taking longer than 360 min -- does make test work locally for you?

I hadn't run `make test`, only manual tests of the form `semgrep -c foo.yaml foo.bash`. Running `make test` now, let's see what happens.

mjambon:
well, `make test` worked for me locally:
```
final confusion matrix: TP: 3375 TN: 1849 FP: 0 FN: 0
```

mjambon:
It has to be a version problem, since `test-develop` succeeds but `test-latest` fails. It's weird that it hangs, though. Taking a closer look.

mjambon:
ok, I can reproduce the problem with semgrep 0.70.0. Looks like infinite looping in the python code:
```
    PID USER      PR  NI    VIRT    RES    SHR S  %CPU  %MEM     TIME+ COMMAND                                          
 365915 martin    20   0  276104  32672   9808 S 100.0   0.2   2:45.30 semgrep                                          
```
When running this:
```
~/semgrep-rules $ semgrep --test --strict --test-ignore-todo bash
Running 1 rules...
ran 1 rules on 1 files: 0 findings
Exception in thread Thread-3:
  File "/usr/lib/python3.8/threading.py", line 932, in _bootstrap_inner
    self.run()
  File "/usr/lib/python3.8/threading.py", line 870, in run
    self._target(*self._args, **self._kwargs)
  File "/usr/lib/python3.8/multiprocessing/pool.py", line 576, in _handle_results
    task = get()
  File "/usr/lib/python3.8/multiprocessing/connection.py", line 251, in recv
    return _ForkingPickler.loads(buf.getbuffer())
  File "/home/martin/.local/share/virtualenvs/semgrep-JT-h3F_l/lib/python3.8/site-packages/attr/_make.py", line 626, in _frozen_setattrs
    raise FrozenInstanceError()
attr.exceptions.FrozenInstanceError
Running 1 rules...
Running 1 rules...
ran 1 rules on 1 files: 2 findings
ran 1 rules on 1 files: 2 findings

```

colleend:
hmm yikes, it seems like one of the threads of the threadpool is getting frozen and therefore the program can't terminate all the threads at the end. I'm not sure what's causing this freeze --> Matt introduced the multiprocessing a few months back and we haven't gotten any issues. 

I'll look into this if I have any free time. Thanks for digging into this! Worst case, we can revert the `test.py` file to what it was before multiprocessing was introduced.

mjambon:
@colleend I'm looking into this. One thing is we've lost some logging info on the semgrep-core side since we did some refactoring two weeks ago, so I'm restoring this first.

colleend:
Awesome, thanks for looking into this. Also good catch on the logging info of semgrep-core!

colleend (approved):
looked into this again, it seems like the `useless-cat` rule passes, but some taint rules don't pass (perhaps because of the semgrep version running in the checks). These taint rules pass when we create a new PR and run the CI checks on the new PR, so they indeed do pass now. 

The rule itself looks good, so I am approving it!

underyx:
@colleend I'd exclude this rule from our default shell ruleset btw, it's not a very practical one. The source wiki page says so as well:

> Pointing out UUOC is a long standing shell programming tradition, and removing them from a short-lived pipeline in a loop can speed it up by 2x. However, it's not necessarily a good use of time in practice, and rarely affects correctness. Ignore as you see fit.

In fact, I usually use `cat foo |` over the alternatives as I think it provides benefits for readability.

colleend:
Good to know, thanks for the context! I haven't added this rule to any rulepacks, and I'll leave it at that for now.

PR #1584, opened 2021-10-30
Add bash rule iteration-over-ls-output similar to shellcheck SC2045

mjambon:
https://github.com/koalaman/shellcheck/wiki/SC2045 is the original shellcheck rule


colleend (approved):
nice! lgtm!

PR #1587, opened 2021-11-03
Update test expectations after adding await to IL

mmcqd:
One `todoid` had to be updated to a `ruleid` after https://github.com/returntocorp/semgrep/pull/4177


IagoAbal (approved):

IagoAbal:
Merged it as `test-develop` passes, `test-latest` will pass when we release Semgrep 0.72.0 next week.

PR #1590, opened 2021-11-04
New Rule - contains-bidirectional-characters

april:
This semgrep rule effectively tests for the [trojansource.codes](https://trojansource.codes/) vulnerability, by essentially doing the same thing that Github is doing: looking for bidirectional characters in code.

This is an unusual rule - it wants to scan across all code but NOT any binary files (like movies and images), so it can't use "generic". There's also no option to have "all" as a language instead. So it currently lists every language, and that can be revisited over time.

One significant downside of this rule is that the time it takes semgrep to process a large number of files seems to scale linearly with the number of languages listed, not the number of files scanned. Two languages takes about as long as one. Four languages takes twice as long as two, and so on.

On my local system (i9 MBP):

• Only `yaml`: 14279 files, 0 findings, 12.40s
• All the languages listed, same directory: 17425 files, 125.34s

About 20% more files had rules run against them, but it took about 10x as long to do so.

### DIFF
--- a/generic/unicode/security/bidi.py
+++ b/generic/unicode/security/bidi.py
@@ -4,10 +4,10 @@ from types import SimpleNamespace
 # A useful lookup table for the bidirectional (bidi) characters
 bidi = SimpleNamespace(
     **{
-        "LRE": "\u202A",  # left-to-right embedding (LRE)
-        "RLE": "\u202B",  # right-to-left embedding (RLE)
-        "LRO": "\u202D",  # left-to-right override (LRO)
-        "RLO": "\u202E",  # right-to-left override (RLO)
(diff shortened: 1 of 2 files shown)


emjin:
The test-latest failure should be unrelated. cc @IagoAbal to confirm

minusworld:
Yeah it looks like something related to taint rules. I should have some time to re-think the CI jobs here next week.

minusworld (approved):

april:
It wasn't clear to me if I should add the `metadata` section or not. I'd be happy to do so now, since that is triggering some of your test rules.

minusworld:
No worries! I added in some metadata. :D

minusworld (approved):

PR #1617, opened 2021-11-18
add new rules to detect bad asp.net practices in web.config

SPoint42:
New rules to detect bad .NEt practices in configuration


minusworld:
@SPoint42 I'm going to split the given tests into separate files so that it works with the testing system.

SPoint42:
Ok

minusworld (approved):

PR #1625, opened 2021-11-22
adding rules for scala

colleend:
3 security rules and 1 correctness rule! 

* the nullcipher kotlin rule should #justwork for scala (i tested this on a scala file I wrote).


inkz (approved):
lgtm

PR #1630, opened 2021-11-29
Add new rule for disabled AWS API Gateway caching.

Ajurr:


minusworld:
Sadly we already have this rule in the registry: https://github.com/returntocorp/semgrep-rules/blob/develop/terraform/aws/best-practice/missing-api-gateway-cache-cluster.yaml

:'(

PR #1645, opened 2021-12-06
Fix ci

colleend:
This PR removes the `yaml/` files and the `contrib/` files from `.semgrepignore`, which actually works now!!!! 

This results in semgrep NOT ignoring the `yaml/` and `contrib/` files, and therefore actually running and producing results we can then compare with expected results.

### DIFF
--- a/.semgrepignore
+++ b/.semgrepignore
@@ -1,5 +1,2 @@
 yaml/semgrep/*.test.yaml
-contrib/nodejsscan/
-contrib/owasp/
-contrib/react/
 template.yaml
--- a/.semgrepignore
+++ b/.semgrepignore
@@ -1,2 +1,1 @@
-yaml/semgrep/*.test.yaml
 template.yaml


minusworld:
Ahh, interesting - we probably want to make these exempted from `semgrep-rule-lints` then in the `pre-commit` file. I think that's why they were added in the first place

colleend:
ah I see -- I think we might need to change `test.py` in order for them to be properly excluded. I can look into this for a second and see how long this will take 👍 

Edit: though we probably want to still lint the files in `yaml/`, just not the files in `contrib/` right?

minusworld:
I think we could exclude the files from `yaml/semgrep/` ... iirc that was the original target anyway

minusworld:
Ah well, let's merge this and then we'll track down and address any real issues as they crop up

minusworld (approved):

PR #1648, opened 2021-12-09
added support for logging error without handling

yoav-orca:
Following the thread in [slack](https://r2c-community.slack.com/archives/C018NJRRCJ0/p1638992547208200?thread_ts=1638944288.203100&cid=C018NJRRCJ0), I opened a PR for this rule.
It matches instances where an error is logged as an error/exception but not handled.


yoav-orca:
I hope I fixed failures

minusworld (approved):

PR #1650, opened 2021-12-10
Java log4j security: Added Lookup injection rule.

SvenTo:
Added a rule for possible Lookup injection into Log4j messages. Lookups provide a way to add values to the Log4j messages at arbitrary places. If the message parameter contains an attacker controlled string, the attacker could inject arbitrary lookups, for instance ``${java:runtime}``. This cloud lead to information disclosure or even remove code execution if ``log4j2.formatMsgNoLookups`` is enabled. This was enabled by default until version 2.15.0.

Reference: https://www.lunasec.io/docs/blog/log4j-zero-day/

### DIFF
--- a/java/log4j/security/log4j-message-lookup-injection.java
+++ b/java/log4j/security/log4j-message-lookup-injection.java
@@ -9,7 +9,7 @@ public class VulnerableLog4jExampleHandler implements HttpHandler {
 
   public void handle(HttpExchange he) throws IOException {
     string userAgent = he.getRequestHeader("user-agent");
-
+    // ruleid: log4j-message-lookup-injection
     log.info("Request User Agent:" + userAgent);
 
   }
--- a/java/log4j/security/log4j-message-lookup-injection.java
+++ b/java/log4j/security/log4j-message-lookup-injection.java
@@ -1,0 +1,16 @@
+import org.apache.log4j.Logger;
+
+import java.io.*;
+import java.util.*;
+
+public class VulnerableLog4jExampleHandler implements HttpHandler {
+
+  static Logger log = Logger.getLogger(log4jExample.class.getName());
+
+  public void handle(HttpExchange he) throws IOException {
+    string userAgent = he.getRequestHeader("user-agent");
+
+    log.info("Request User Agent:" + userAgent);
+
+  }
+}
--- a/java/log4j/security/log4j-message-lookup-injection.yaml
+++ b/java/log4j/security/log4j-message-lookup-injection.yaml
@@ -1,0 +1,187 @@
+rules:
+- id: log4j-message-lookup-injection
+  metadata:
+    cwe: "CWE-74: Improper Neutralization of Special Elements in Output Used by a Downstream Component ('Injection')"
+    owasp: 'A1: Injection'
+    source-rule-url: https://www.lunasec.io/docs/blog/log4j-zero-day/
+    references:
+    - https://issues.apache.org/jira/browse/LOG4J2-3198
+    - https://www.lunasec.io/docs/blog/log4j-zero-day/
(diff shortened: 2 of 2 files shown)


ievans:
See also: https://twitter.com/lapt0r/status/1469160133737340932?s=20

IagoAbal:
Maybe there is a better way to write the rule, but this particular proposal would be better re-formulated as a taint-mode rule, IMO.

minusworld:
I'm going to go ahead and merge this in so people can get started looking for this.

However, I agree that there are improvements that could be made.

A few things come to mind:
- There may be other logging frameworks with `info`, `error`, and similar methods, and I don't think this rule scopes to log4j
- Could reduce the number of patterns by using this paradigm instead to filter out the constant string case:
```yaml
- pattern: $LOGGER.$METHOD(...)
- pattern-not: $LOGGER.$METHOD("...")
- metavariable-regex:
    metavariable: $METHOD
    regex: (info|error|warn|etc.)
```
- Could use typed metavariables to make sure $LOGGER is a Logger instance: `(Logger $LOGGER).$METHOD(...)`

minusworld (approved):

minusworld:
cc @SvenTo for the above comments

sschakraborty:
@SvenTo I believe what you're trying to do with this rule is you're looking for message being built using String concatenations or String.format(...) patterns before being sent to the logger.

However a statement like ```logger.info("Received request with host: {}", request.host());``` is still vulnerable. Are we looking at these cases?

@minusworld @colleend Thoughts?

minusworld:
@sschakraborty Some people in our Slack channel also came up with this iteration of this rule: https://semgrep.dev/s/chegg:log4j2_tainted_argument

I think this might be closer to what you're looking for.

In general, I think https://semgrep.dev/s/chegg:log4j2_tainted_argument might be a more precise rule since it will detect any non-constant string to any Logger method, and is scoped to be within files that import something from org.apache.log4j.

@SvenTo What do you think about using the linked rule instead? Am I missing any corner cases here? 

Thanks everyone! :D

SvenTo:
@sschakraborty
Yes, I initially looked for log statements being built using String concatenations in the message parameter, but the mentioned statement is also vulnerable. I wanted to improve the rule, but I've seen a few more challenges:

1. If you're logging with the [SLF4J API in combination with log4j <2.15.0](http://slf4j.org/log4shell.html), you're still vulnerable but ``import org.apache.log4j.$PKG;`` will not show the occurrences [1].

2. In the end, the rule will probably list a lot of false positives because the rule just finds all log statements with dynamic input. This is fine if you just want to find a suitable log statement for an attack. However, I don't think this is a good rule for semgrep in the long-term. It probably makes sense to implement this rule as [a ``mode: taint`` rule](https://r2c.dev/blog/2021/taint-mode-is-now-in-beta/) as suggested by @IagoAbal because this will eliminate a lot of false positives. Even if you have a patched log4j version, there are more attack scenarios [when logging attacker-controlled input](https://owasp.org/www-community/attacks/Log_Injection).

What do you think? I didn't have the time to improve the rule yet. So feel free to make a pull request for this rule if you have an improved version.

BTW exception messages are not vulnerable, for instance this code is safe:
```java
string attackerControlledInput = "JNDI: ${jndi:rmi://localhost:1389/a} ${java:version}";
try {
    throw new Exception("Example: " + attackerControlledInput);
} catch(Exception e) {
    logger.error("Exception", e);
}
```

[1] Not to forget code using the [Project Lombok](https://projectlombok.org/api/lombok/extern/slf4j/Slf4j.html).

sschakraborty:
@SvenTo I'm working on a rule on the similar lines. I'll be testing with many different scenarios. When it's done, I'll bring in the changes to your rule. Thanks!

PR #1660, opened 2021-12-14
remove formatting checks from pre-commit

minusworld:


.pre-commit-config.yaml
```diff
@@ -9,19 +9,3 @@ repos:
   - id: check-executables-have-shebangs
   - id: check-merge-conflict
   - id: check-symlinks
-  - id: trailing-whitespace
-  - id: end-of-file-fixer
```
colleend, line 13: we don't want trailing-whitespace or end-of-file-fixer?

minusworld: No, these make modifications to the file is and one of the primary reasons that PRs fail. I want to streamline the process and auto-apply formatting later rather than hoping contributors remember to run pre-commit.

minusworld: @colleend ping

brendongo: We can make it so that they don't modify the file if that is preferable?

minusworld: I still don't want the tests to fail - and then I want to automatically apply the formatting in CI we want rather than hoping the humans remember to do it

enncoded (approved):

.pre-commit-config.yaml
colleend: sorry, my bad -- just saw this! makes sense to apply formatting automatically later, it'll be a better experience for contributors 👍

PR #1662, opened 2021-12-15
Action to push dev to release

colleend:
This PR:
* creates a new action daily (between 17:30-19:30 UTC, aka 9:30AM-11:30AM PST) that merges the develop branch into release.

Here's an example of the PR it creates:
https://github.com/returntocorp/semgrep-rules/pull/1687


minusworld:
OK, this looks good to me. Could you scan the GHA you're using with security-audit just to see if there's anything egregious in there?

.github/workflows/push-dev-to-release.yml
```diff
@@ -0,0 +1,30 @@
+name: Create Pull Request to Push Dev to Release
+
+on:
+  workflow_dispatch:
+  schedule:
+  - cron: 30 17,19 * * *
```
minusworld: I think this cron schedule will run at 1730 AND at 1930. You can just use one or the other

colleend: cool, I'll just use 1730 UTC then 👍

colleend:
yes, can do! 

Looked into this, the semgrep run led to:
```
dist/index.js
rule:javascript.lang.security.detect-child-process.detect-child-process: Detected non-literal calls to $EXEC(). This could lead to a command injection vulnerability. Details: https://sg.run/l2lo
1973:        const exitCode = yield exec(commandLine, args, Object.assign(Object.assign({}, options), { listeners }));
3606:                    yield exec(`${cmdPath} /s /c "rd /s /q "%inputPath%""`, {
3607:                        env: { inputPath }
3608:                    });
3611:                    yield exec(`${cmdPath} /s /c "del /f /a "%inputPath%""`, {
3612:                        env: { inputPath }
3613:                    });
ran 245 rules on 38 files: 3 findings

A new version of Semgrep is available. See https://semgrep.dev
```

I looked into the last two findings, and `cmdPath` is technically the path of `cmd.exe` in Windows and the first exec is contained in a class that isn't really used. It does also seem like there's a bunch of `execs` in the code, and I'd be down to try to do this all through `git` command line instead. @minusworld what do you think?

minusworld:
I -think- since that code is just in `dist/` it's a "compiled" version of the code and therefore probably OK.

However, if you wanted to try the GitHub API for making PRs, here's a ref: https://docs.github.com/en/rest/reference/pulls#create-a-pull-request

The latter option would certainly be simpler, and less prone to issues as a result 🤷‍♂️  My 2c. I think either is fine.

colleend:
hmm I see! I'm going to try doing this with 'github/hub' first, and if I can't get it figured out by tomorrow then I'll use the 'create-pull-request' action. Thanks for the 2c 👍

colleend:
Changed to using the Github API! Now an example of a file created by this action is https://github.com/returntocorp/semgrep-rules/pull/1687.

minusworld (approved):
Sweet, this looks pretty good

colleend:
yay!

PR #1666, opened 2021-12-16
Fix pg-orm-sqli rule

jamiees2:
https://github.com/returntocorp/semgrep-rules/pull/1235 missed adding a pattern-either outer filter for `$DB`, so `$DB` ends up being any variable. This causes issues with other functions that expose a `Join`, or really any other method filtered by `$METHOD`, since it matches it when it shouldn't have.

This just fixes that issue and adds a test that `path.Join` still works.

Signed-off-by: James Elias Sigurdarson <jamiees2@gmail.com>

### DIFF
--- a/go/lang/security/audit/sqli/pg-orm-sqli.go
+++ b/go/lang/security/audit/sqli/pg-orm-sqli.go
@@ -2,6 +2,7 @@ package main
 
 import (
     "fmt"
+    "path"
 
     "github.com/go-pg/pg/v10"
     "github.com/go-pg/pg/v10/orm"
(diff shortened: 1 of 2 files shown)


jamiees2:
oh I understand, the problem is that the rule's behavior is basically to filter method calls to the methods filtered by $METHOD, and almost just ignore $DB. This way it can match calls like .Model(...).Where(...).WhereOr(

jamiees2:
I don't know if I can fix this issue, this change basically breaks the rule, since you never can call $DB.WhereOr, but the rule as is doesn't do what it's meant to. For now, we've disabled it for us

minusworld:
Hmm, let me see if I can mess with the rule enough to figure out a way to limit it to DB-like objects. Semgrep can do limited typing for Go (see https://r2c.dev/blog/2020/type-awareness-in-semantic-grep/) so maybe that would work

minusworld:
@jamiees2 Would it be sufficient to filter out irrelevant calls like `path.Join(...)` as we come across them?

I edited the rule to just filter out `path.Join(...)` in this case. As we find more conflicts, we could add them instead of looking for a general solution for now.

colleend (approved):
nice! Thanks for working on this 🙏

PR #1670, opened 2021-12-17
Add rule for specifying encoding on file open (python best practice)

munyanjacob:
While working on freelawproject/reporters-db#74 and freelawproject/reporters-db#75, found that it is python [best practice](https://www.python.org/dev/peps/pep-0597/#motivation) to specify encoders when opening files because python uses the locale encoder for `open()`, and Windows machines may default to a locale encoding that corrupts special characters. Python [`open()` documentation](https://docs.python.org/3/library/functions.html#open) specifies that this applies when opening files in text mode (not binary mode).

This rule throws an error for all open() calls that are not in binary mode and do not specify an encoding (and could cause unintended behavior for Windows users).

There are probably more cases where Python defaults to the locale encoding, but I thought this would be a good one for now.

Pretty new to open source - please give any feedback! Will look out for the CLA.

### DIFF
--- a/python/lang/best-practice/unspecified-open-encoding.py
+++ b/python/lang/best-practice/unspecified-open-encoding.py
@@ -1,0 +1,77 @@
+def func1():
+    # ruleid:unspecified-open-encoding
+    fd = open('foo')
+    fd.close()
+
+def func2():
+    # ruleid:unspecified-open-encoding
+    fd = open('foo', mode="w")
+    fd.close()
+
+def func3():
+    import os
+    db_root="test"
+    # ruleid:unspecified-open-encoding
+    with open(os.path.join(db_root, "data.json")) as f:
+        i = 2
+
+def func4():
+    import os
+    # ruleid:unspecified-open-encoding
+    with open(os.path.join("test", "b", mode="b")) as f:
+        i = 2
+
+def func15():
+    # ruleid:unspecified-open-encoding
+    fd = open('foo', buffering=1)
+    fd.close()
+
+def func5():
+    # ok:unspecified-open-encoding
+    fd = open('foo', 'b', closefd=True)
+    fd.close()
+
+def func6():
+    # ok:unspecified-open-encoding
+    fd = open('foo', mode="b")
+    fd.close()
+
+def func7():
+    # ok:unspecified-open-encoding
+    fd = open('foo', encoding='utf-8')
+    fd.close()
+
+def func8():
+    # ok:unspecified-open-encoding
+    fd = open('foo', encoding="utf-8", mode="w")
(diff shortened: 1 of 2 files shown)


python/lang/best-practice/unspecified-open-encoding.py
```diff
@@ -0,0 +1,77 @@
...
+    import os
+    db_root="test"
+    # ruleid:unspecified-open-encoding
+    with open(os.path.join(db_root, "data.json")) as f:
+        i = 2
+
+def func4():
+    import os
+    # ruleid:unspecified-open-encoding
+    with open(os.path.join("test", "b", mode="b")) as f:
```
minusworld, line 21: Are the parentheses correct here?

minusworld: Or is this intended to test `mode` but NOT inside `open`?

python/lang/best-practice/unspecified-open-encoding.yaml
```diff
@@ -0,0 +1,36 @@
...
+      - metavariable-regex:
+          metavariable: $MODE
+          regex: (?!.*b.*)
+
+  message: >-
+    Missing 'encoding' parameter.
+    'open()' uses device locale encodings by default, corrupting files with special characters.
+    Specify the encoding to ensure cross-platform support when opening files in text mode (e.g. encoding="utf-8").
+  languages: [python]
+  severity: ERROR
```
minusworld: nit: Mark this as a WARNING since it won't immediately crash the program and since the locale encoding might be right some of the time

python/lang/best-practice/unspecified-open-encoding.py
munyanjacob: Yep, this was intended as a test to make sure only args to `open` are checked. A previous version of the rule looked for `mode` with a regex instead of the `pattern:` format so I added the test. Not necessary anymore - should I remove?

python/lang/best-practice/unspecified-open-encoding.yaml
munyanjacob: You're right - for some reason I thought `encoding="locale"` from [here](https://www.python.org/dev/peps/pep-0597/#encoding-locale) was implemented, but it was just a proposal. Corrected this in the recent commit.

python/lang/best-practice/unspecified-open-encoding.py
minusworld: Any test is a good test :D I was just asking to make sure I understood. 👍

minusworld (approved):

mlissner:
This is great!

PR #1675, opened 2021-12-17
Identify SecretsManager secret policies that do not have the block_public_policy flag explicitly set

christophetd:
A big argument for this rule is that https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/secretsmanager_secret_policy#block_public_policy has an insecure example

### DIFF
--- a/terraform/lang/security/secretsmanager-block-public-access.tf
+++ b/terraform/lang/security/secretsmanager-block-public-access.tf
@@ -1,0 +1,24 @@
+resource "aws_secretsmanager_secret" "example" {
+  name = "example"
+}
+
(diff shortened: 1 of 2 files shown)


minusworld:
@christophetd Is the default for BlockPublicPolicy `false`? Looking at the documentation, it reads:

> BlockPublicPolicy
> Specifies whether to block resource-based policies that allow broad access to the secret. By default, Secrets Manager blocks policies that allow broad access, for example those that use a wildcard for the principal.

Is the behavior different when this is set to `true`, `false`, and not set at all? 🤔

PR #1676, opened 2021-12-17
Use regular semgrep for semgrep-rule-lints so that we can exclude certain files

minusworld:


colleend (approved):
nice!! super exciting :D

PR #1679, opened 2021-12-17
Consider 0 or more whitespace characters before regex in taitned-sql-…

minusworld:
…string rules. Discovered FN in DjangoGoat; added this FN as a test in Django variant of the rule


colleend (approved):
Grayson: "Last one, I promise 😄 "
Also Grayson: 

LGTM! thanks for fixing :D

minusworld:
> Grayson: "Last one, I promise 😄 " Also Grayson:
> 
> LGTM! thanks for fixing :D

It was the last one I -asked- for :D
Not the last one I'll make! Can't keep me away

PR #1682, opened 2021-12-17
Validate rules before running --test

minusworld:
Turns out that running `--test` swallows any invalid configs and doesn't produce an error:

```
➜  security git:(debugging) ✗ semgrep --validate --test
The following config files produced errors:
	s3-public-read-bucket.yaml: invalid configuration file found (1 configs were invalid)
18 yaml files tested
check id scoring:
(TODO: 0) s3-public-rw-bucket.yaml
	✔ s3-public-rw-bucket                                          TP: 1 TN: 0 FP: 0 FN: 0
(TODO: 0) s3-cors-all-origins.yaml
	✔ all-origins-allowed                                          TP: 1 TN: 1 FP: 0 FN: 0
...
➜  security git:(debugging) ✗ echo $?
0
```

and this caused an issue where we released an invalid rule and broke the app for a while today.

This PR adds an explicit validation step.

### DIFF
--- a/.github/workflows/semgrep-rules-test-develop.yml
+++ b/.github/workflows/semgrep-rules-test-develop.yml
@@ -16,7 +16,7 @@ jobs:
     - uses: actions/checkout@v2
     - name: validate rules
       run: |
-        docker run --rm -v $(PWD):/src returntocorp/semgrep:develop --validate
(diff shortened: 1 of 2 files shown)


underyx (approved):

colleend (commented):
good catch!! was actually thinking about this yesterday, but forgot to mention it 😅

Issue #1631, opened 2021-11-29
[Rule] Checkov Terraform AWS rules


minusworld:
**Rule Description**
Checkov Terraform AWS rules

*What does this rule intend to find?*
- [x] ALBDropHttpHeaders
- [x] APIGatewayCacheEnable
- [x] AppLoadBalancerTLS12
- [x] AthenaWorkgroupEncryption
- [x] AutoScalingTagging
- [x] BackupVaultEncrypted
- [ ] CloudWatchLogGroupKMSKey
- [ ] CloudWatchLogGroupRetention
- [ ] CloudfrontTLS12
- [ ] CloudtrailEncryptedWithCMK
- [ ] CodeBuildProjectEncrypted
- [ ] CodeBuildProjectEncryption
- [ ] DBInstanceBackupRetentionPeriod
- [ ] DBInstanceLogging
- [ ] DocDBEncryptedWithCMK
- [ ] DynamoDBTablesEncrypted
- [ ] EBSDefaultEncryption
- [ ] EBSSnapshotCopyEncryptedWithCMK
- [ ] EBSVolumeEncryptedWithCMK
- [ ] EFSFileSystemEncryptedWithCMK
- [ ] ELBv2AccessLogs
- [ ] EMRClusterIsEncryptedKMS
- [ ] ElasticCacheAutomaticBackup
- [ ] ElasticacheReplicationGroupEncryptedWithCMK
- [ ] ElasticsearchNodeToNodeEncryption
- [ ] FSXOntapFSEncryptedWithCMK
- [ ] FSXWindowsFSEncryptedWithCMK
- [ ] GlacierVaultAnyPrincipal
- [ ] IAMAdminPolicyDocument
- [ ] ImagebuilderComponentEncryptedWithCMK
- [ ] KMSKeyWildcardPrincipal
- [ ] KMSRotation
- [ ] KinesisStreamEncryptedWithCMK
- [ ] KinesisVideoEncryptedWithCMK
- [ ] LBCrossZone
- [ ] LBDeletionProtection
- [ ] LambdaEnvironmentEncryptionSettings
- [ ] LustreFSEncryptedWithCMK
- [ ] QLDBLedgerDeletionProtection
- [ ] QLDBLedgerPermissionsMode
- [ ] RDSClusterIAMAuthentication
- [ ] RDSIAMAuthentication
- [ ] RDSMultiAZEnabled
- [ ] RedShiftSSL
[...]

PR #1689, opened 2021-12-21
Adding migrated checkov rules to AWS terraform checks

mertcancoskuner:
Migrated the Checkov rules listed at https://github.com/returntocorp/semgrep-rules/issues/1631 except some of them such as WAFHasLogs or WafHasAnyRules.


minusworld:
@mertcancoskuner Wow, thanks!! I've approved for CI and will look over the files sometime in the next few weeks. Many of the approvers (including myself) are on holiday, so we'll get to it when we can. Thanks again!

clintgibler:
Yay, this is awesome, thanks @mertcancoskuner!!  🙌 🎉  As @minusworld said, many of us are on holiday, but we'll review this as soon as we can. Thanks again and happy holidays :D

minusworld:
@mertcancoskuner We're back! Looking over the PR, it looks like the test files are using different test annotations from what the CI here expects.

The CI here uses a special `> semgrep --test` flag which looks for `# ruleid: <id_here>` directly above the expected match lines in files with the same name.

I added this annotation to aws-elasticache-automatic-backup-not-enabled as an example. Would you mind updating the test files to match this format? I'd also be happy to take half of them myself -- let me know.

Thanks!

mertcancoskuner:
> @mertcancoskuner We're back! Looking over the PR, it looks like the test files are using different test annotations from what the CI here expects.
> 
> The CI here uses a special `> semgrep --test` flag which looks for `# ruleid: <id_here>` directly above the expected match lines in files with the same name.
> 
> I added this annotation to aws-elasticache-automatic-backup-not-enabled as an example. Would you mind updating the test files to match this format? I'd also be happy to take half of them myself -- let me know.
> 
> Thanks!

@minusworld gotcha. I'll update the others accordingly.

terraform/aws/security/aws-cloudtrail-encrypted-with-cmk.yaml
```diff
@@ -0,0 +1,26 @@
...
+      resource "aws_cloudtrail" $ANYTHING {
+        ...
+      }
+  - pattern-not-inside: |
+      resource "aws_cloudtrail" $ANYTHING {
+        ...
+        kms_key_id = ...
+        ...
+      }
+  - pattern-regex: (^aws_kms_key\.(.*))
```
minusworld: It looks like the intent of this regex is to filter out cases where `kms_key_id` is set to something with `aws_kms_key` as a prefix. Is that right?

In that case you'll have to group the `pattern-not-inside` and regex under a parent `patterns` clause so that they go together. In addition, this might be easier to do by capturing the right hand side of `kms_key_id` in a metavariable and using `metavariable-pattern` to test it.

terraform/aws/security/aws-ssm-document-logging-issues.yaml
```diff
@@ -0,0 +1,31 @@
+rules:
+- id: aws-ssm-document-logging-issues
+  patterns:
+  - pattern-inside: |
+      resource "aws_ssm_document" $ANYTHING {
+        ...
+      }
+  - pattern: content = "$STATEMENT"
+  - metavariable-pattern:
+      metavariable: $STATEMENT
```
minusworld: It looks like this might be missing the language field so that it scans JSON or YAML.

To do two languages with `metavariable-pattern`, you may have to break this up into two `patterns` clauses, one with `pattern: content="$STATEMENT"` and `metavariable-pattern` for JSON, and a second one with `pattern: content="$STATEMENT"` and `metavariable-pattern` for YAML.

terraform/aws/security/aws-dynamodb-table-unencrypted.tf
```diff
@@ -0,0 +1,72 @@
...
+  hash_key       = "UserId"
+  range_key      = "UserId"
+
+  attribute {
+    name = "UserId"
+    type = "S"
+  }
+
+  server_side_encryption {
+      enabled = true
```
minusworld, line 71: This one will still get flagged because it does not contain the `kms_key_arn = ...` inside. Do you want to flag this one, or does the pattern need amending?

terraform/aws/security/aws-elb-access-logs-not-enabled.yaml
```diff
@@ -0,0 +1,40 @@
+rules:
+- id: aws-elb-access-logs-not-enabled
+  patterns:
+  - pattern-either:
```
minusworld: I think this rule is missing clauses for the `aws_alb` resource in addition to `aws_lb`

terraform/aws/security/aws-codebuild-project-unencrypted.tf
```diff
@@ -0,0 +1,58 @@
...
+}
+
+# unknown
+
+resource "aws_codebuild_project" "no_artifacts" {
+  name         = "example"
+  service_role = "aws_iam_role.example.arn"
+
+  artifacts {
+    type = "NO_ARTIFACTS"
```
minusworld, line 49: Do you have a preference for whether this case with NO_ARTIFACTS should be flagged? Currently it is flagged, and in that case it'll need a test annotation

terraform/aws/security/aws-elasticsearch-nodetonode-encryption.tf
```diff
@@ -0,0 +1,73 @@
...
+
+resource "aws_elasticsearch_domain" "instance_count_not_bigger_than_1" {
+  domain_name           = "instance_count_not_bigger_than_1"
+
+  cluster_config {
+    instance_count = 1 // a value <= 1
+  }
+}
+
+resource "aws_elasticsearch_domain" "node_to_node_encryption_enabled" {
```
minusworld, line 20: Should this one be flagged? It's currently highlighted: https://semgrep.dev/s/5yqX

terraform/aws/security/aws-transfer-server-is-public.yaml
```diff
@@ -0,0 +1,27 @@
+rules:
+- id: aws-transfer-server-is-public
+  patterns:
+  - pattern-either: 
+    - pattern: |
+        resource "aws_transfer_server" $ANYTHING {
+          ...
+        }
+    - pattern: |
```
minusworld: Based on the test case, is this supposed to be `pattern-not-inside`?

terraform/aws/best-practice/aws-s3-bucket-versioning-not-enabled.tf
```diff
@@ -0,0 +1,108 @@
...
+            ]
+        }
+    }
+  ]
+}
+POLICY
+}
+
+
+resource "aws_s3_bucket" "this" {
```
minusworld, line 98: This one also gets flagged. Could try filtering it out with `enabled = var.$ENABLED`.

Sadly we don't have TF variable resolution yet... but maybe someday!

terraform/aws/best-practice/missing-aws-lb-deletion-protection.yaml
```diff
@@ -0,0 +1,22 @@
+rules:
+- id: missing-aws-lb-deletion-protection
+  patterns:
+    - pattern: |
+        resource "aws_lb" $ANYTHING {
+          ...
+        }
+    - pattern-not-inside: |
+        resource "aws_lb" $ANYTHING {
```
minusworld: This rule looks like it's missing the `aws_alb` variants

terraform/aws/best-practice/missing-aws-cross-zone-lb.yaml
```diff
@@ -0,0 +1,22 @@
+rules:
+- id: missing-aws-cross-zone-lb
+  patterns:
```
minusworld, line 3: Missing `aws_alb` cases

terraform/aws/best-practice/aws-s3-object-lock-not-enabled.tf
```diff
@@ -0,0 +1,46 @@
...
+  acl    = "private"
+
+  object_lock_configuration {
+    object_lock_enabled = "Disabled"
+  }
+}
+
+# unknown
+
+resource "aws_s3_bucket" "default" {
```
minusworld, line 43: Do you have a preference for whether this one gets flagged? Currently, the rule does flag here; however, this would alert on many S3 configurations.

If you want to alert here, it'll need a test annotation

terraform/aws/best-practice/aws-elasticache-automatic-backup-not-enabled.tf
```diff
@@ -0,0 +1,41 @@
...
+}
+# ruleid: aws-elasticache-automatic-backup-not-enabled
+resource "aws_elasticache_cluster" "disabled" {
+  cluster_id           = "cluster"
+  engine               = "redis"
+  node_type            = "cache.m5.large"
+  num_cache_nodes      = 1
+  parameter_group_name = "default.redis6.x"
+
+  snapshot_retention_limit = 0
```
minusworld, line 30: The associated rule will not flag this one since the rule only checks for the presence of `snapshot_retention_limit` set to anything. You can filter this out with another `pattern-not-inside` clause checking for `0`.

terraform/aws/best-practice/aws-rds-iam-authentication-not-enabled.tf
```diff
@@ -0,0 +1,69 @@
...
+  instance_class    = "db.t3.small"
+  password          = "password"
+  username          = "username"
+
+  iam_database_authentication_enabled = false
+}
+
+# unknown
+
+resource "aws_db_instance" "mariadb" {
```
minusworld, line 64: Would you prefer if this one gets flagged? Currently it will get flagged.

minusworld (changes requested):
I left a bunch of comments.
They can be summarized as:
- All the `-cmk` rules need an alternate implementation. I've asked @lejliu to help figure out how to write these.
- There are a handful of rules that match on `aws_lb` but not `aws_alb`.
- Several of the test files have an `# unknown` section at the end - we'll have to decide if we want to prefer flagging these resources or not. I think I left comments on each one with my point of view, so hopefully they're easy to find and discuss.

I fixed a few of the minor issues as I went, like test annotations on the wrong line. Hopefully that means the results of the test run only have issues we actually need to address.

Thanks again! Let us know how we can be of help. :D

terraform/aws/security/aws-cloudtrail-encrypted-with-cmk.yaml
mertcancoskuner: That is right. Although, in a second thought, that might not be needed at all and can be scrapped altogether. What do you think?

terraform/aws/security/aws-dynamodb-table-unencrypted.tf
mertcancoskuner: Yes, since it has no cmk, we want to flag this one

terraform/aws/security/aws-elb-access-logs-not-enabled.yaml
mertcancoskuner: Yes, `aws_alb` is missing and needs fixing

terraform/aws/security/aws-ssm-document-logging-issues.yaml
mertcancoskuner: Will change it accordingly

terraform/aws/security/aws-codebuild-project-unencrypted.tf
mertcancoskuner: It may cause false positives (which is valid for all unknown cases). Do you think it's worth increasing coverage for the sake of having more false positives?

terraform/aws/security/aws-elasticsearch-nodetonode-encryption.tf
mertcancoskuner: This shouldn't be flagged. Need to work on this to identify what is the issue

terraform/aws/security/aws-transfer-server-is-public.yaml
mertcancoskuner: That's true, it should be `pattern-not-inside` to look for `endpoint_type = "VPC"`. Will fix

terraform/aws/best-practice/aws-s3-object-lock-not-enabled.tf
mertcancoskuner: It may cause false positives (which is valid for all unknown cases). Do you think it's worth increasing coverage for the sake of having more false positives?

terraform/aws/best-practice/aws-rds-iam-authentication-not-enabled.tf
mertcancoskuner: It may cause false positives (which is valid for all unknown cases). Do you think it's worth increasing coverage for the sake of having more false positives?

mertcancoskuner:
> I left a bunch of comments. They can be summarized as:
> 
> * All the `-cmk` rules need an alternate implementation. I've asked @lejliu to help figure out how to write these.
> * There are a handful of rules that match on `aws_lb` but not `aws_alb`.
> * Several of the test files have an `# unknown` section at the end - we'll have to decide if we want to prefer flagging these resources or not. I think I left comments on each one with my point of view, so hopefully they're easy to find and discuss.
> 
> I fixed a few of the minor issues as I went, like test annotations on the wrong line. Hopefully that means the results of the test run only have issues we actually need to address.
> 
> Thanks again! Let us know how we can be of help. :D

- We can remove the `aws_kms_key` check as terraform might be linting these already in case of any non-compliant input which would solve `-cmk` implementation.
- I'll add `aws_alb` cases wherever applicable.
- I think having fewer false positives is better in terms of overall user experience which I also prefer but having those may increase the rule coverage. What do you think?

terraform/aws/security/aws-codebuild-project-unencrypted.tf
minusworld: I think it depends on whether this is an issue worth reporting or not. However, if we're uncertain, I'd say have more false positives -- if it annoys someone more knowledgeable than us, they will make an issue about it ;)

terraform/aws/best-practice/aws-s3-object-lock-not-enabled.tf
minusworld: Same as above

minusworld:
> * We can remove the `aws_kms_key` check as terraform might be linting these already in case of any non-compliant input which would solve `-cmk` implementation.
> * I'll add `aws_alb` cases wherever applicable.
> * I think having fewer false positives is better in terms of overall user experience which I also prefer but having those may increase the rule coverage. What do you think?

I'm sure there's a way to get the `-cmk` rules to work... but we can remove them for now and put them up in another PR.

Re: `aws_alb`: great! Thank you 🙏 

Fewer false positives generally makes end users happier, in our experience. I think that makes sense, contrary to my inline comments above. 😅  If we err on the side of fewer false positives all around, that's OK.

mertcancoskuner:
> > * We can remove the `aws_kms_key` check as terraform might be linting these already in case of any non-compliant input which would solve `-cmk` implementation.
> > * I'll add `aws_alb` cases wherever applicable.
> > * I think having fewer false positives is better in terms of overall user experience which I also prefer but having those may increase the rule coverage. What do you think?
> 
> I'm sure there's a way to get the `-cmk` rules to work... but we can remove them for now and put them up in another PR.
> 
> Re: `aws_alb`: great! Thank you 🙏
> 
> Fewer false positives generally makes end users happier, in our experience. I think that makes sense, contrary to my inline comments above. 😅 If we err on the side of fewer false positives all around, that's OK.

Removed `kms_key_id` comparison, fixed lb checks and ignored the unknowns. As you mentioned, if someone sees there is a good fix for cmk checks or unknowns causing false positives after community usage we can revise those as we go. There are two points that I couldn't manage to pull off;

> To do two languages with metavariable-pattern, you may have to break this up into two patterns clauses, one with pattern: content="$STATEMENT" and metavariable-pattern for JSON, and a second one with pattern: content="$STATEMENT" and metavariable-pattern for YAML.

This one (`aws-ssm-document-logging-issues.yaml`) started to give `metavariable-pattern failed because $STATEMENT does not bind to a sub-program` error which I couldn't figure out why

> The associated rule will not flag this one since the rule only checks for the presence of snapshot_retention_limit set to anything. You can filter this out with another pattern-not-inside clause checking for 0.

The solution that we discussed didn't solve the problem. I've tried `metavariable-comparison` with `$COUNT > 0` which resulted in failure as well.

minusworld:
@mertcancoskuner I went through and tried to fix up the remaining rules. There's only one left now that is giving me a bit of trouble. I think it can be written, so I'll try to get to it early next week.

minusworld:
Turns out there's an issue matching YAML documents in HCL: https://github.com/returntocorp/semgrep/issues/4582

minusworld (approved):

PR #1698, opened 2022-01-05
Fix parse error that is causing semgrep CI to fail

aryx:
See test failures here:
https://github.com/returntocorp/semgrep/runs/4713621483?check_suite_focus=true

test plan:
semgrep-core -lang ruby -dump_ast rails-skip-forgery-protection.rb
does not return any parse error anymore


aryx:
Not sure why the semgrep-rules CI didn't catch the error.
It's a partial parse error (tree-sitter can recover for certain errors), but still we should be more
strict in semgrep-rules and accept only files we can fully parse.

TinMarkovic (approved):

PR #1701, opened 2022-01-06
Add confidence levels to about 150 rules

minusworld:
This PR is an experiment adding confidence levels in the metadata section to about 150 rules.

The goal is to see if this metric makes sense to users on the receiving end. Confidence is usually a component of risk or vulnerability assessments, but is largely absent from SAST tools, likely due its similarity to severity.

Working definitions:

 Level | Definition | Examples | Example rule
-- | -- | -- | --
Very high | This code is definitely a security vulnerability | Near-100% certainty that user input is used unescaped in the WHERE clause of a SQL query, eval, or similar | python.django.security.injection.code.user-eval.user-eval
High | This code is likely a security vulnerability | User input flows into the WHERE clause of a SQL query, eval, or similar | go.lang.security.injection.tainted-sql-string.tainted-sql-string
Medium | This code could lead to a vulnerability if other conditions are present | Manually constructed SQL query with non-constant variables | python.django.security.audit.raw-query.avoid-raw-sql
Low | This code is unlikely to be a vulnerability, or is highlighted for informational purposes | Constant string SQL query | None

Another option worth considering is something like "kinds" of rules. For example, a taint rule is a "bug finding" kind of rule. Checking for non-constant SQL queries is a "secure guardrails" kind of rule. I anticipate confusion using typical H/M/L levels, so maybe having clearer descriptions for a confidence-like field would be better.


adamberman (approved):
This looks good to me, but there are a bunch of other non-confidence level changes in here (licensing changes, severity changes, etc) that I don't feel qualified to comment on.

minusworld:
> This looks good to me, but there are a bunch of other non-confidence level changes in here (licensing changes, severity changes, etc) that I don't feel qualified to comment on.

Thanks! In general, I think the way the licensing is applied with broad strokes in the registry itself but is overwritten if a license field is present. I guess I made that change while I was in there.

For severity, It's been the wild west, so I've also been using a severity definition to edit those as I go so that, hopefully, they aren't conflated with confidence.

Answers above for posterity :D

PR #1702, opened 2022-01-07
Added formatting stuff

enncoded:
1. Added a github actions workflow to format future rules. Runs on PR and Push for both `develop` and `release`

### DIFF
--- a/.github/workflows/format-rules.yaml
+++ b/.github/workflows/format-rules.yaml
@@ -17,14 +17,20 @@ jobs:
           node-version: "17.3.0"
       - name: install prettier
         run: npm install prettier
+      - name: check for unformatted rules
+        id: check
+        continue-on-error: true
+        run: npx prettier --check "./**/*.yaml"
       - name: format unformatted rules
+        id: format
+        if: steps.check.outcome == 'failure'
         run: npx prettier --write $(npx prettier --list-different "./**/*.yaml")
       - name: commit formatted files
+        id: commit
+        if: steps.format.outcome == 'success'
         run: |
-          date > generated.txt
           git config user.name github-actions
           git config user.email github-actions@github.com
           git add .
           git commit -m "Formatted new rules"
           git push
-
--- a/.github/workflows/format-rules.yaml
+++ b/.github/workflows/format-rules.yaml
@@ -1,0 +1,30 @@
+name: format-rules
+
+on:
+  push:
+    branches:
+      - develop
+      - release
+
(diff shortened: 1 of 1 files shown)


.github/workflows/format-rules.yaml
```diff
@@ -0,0 +1,25 @@
...
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - uses: actions/setup-node@v2
+        with:
+          node-version: "17.3.0"
+      - name: install prettier
+        run: npm install prettier
+      - name: format unformatted rules
+        run: npx prettier --write $(npx prettier --list-different "./**/*.yaml")
```
minusworld, line 26: Will this workflow file write changes back to the repository? I think that will require making a commit. Have you tested this on GitHub itself?

.github/workflows/num-rules.yaml
```diff
@@ -0,0 +1,22 @@
+name: num-rules
```
minusworld: Do you know why this appears as "added?" I think this file was already in the repository. I guess maybe it got formatted?

minusworld: Oh, I see, it was renamed from `yml` to `yaml`.

minusworld:
@lejliu I left a question about whether the workflow actually applies the autoformatting back to the repository.

Also, it might be beneficial to split this up into multiple PRs, two at minimum:
1. For the automatic formatting workflow
2. Reformatting everything according to the new formatter
- - You may also consider doing this in smaller batches. Maybe 2-3 PRs total. (Large amounts of changed files makes the GH UI choke on my machine/browser)

.github/workflows/format-rules.yaml
```diff
@@ -0,0 +1,30 @@
...
+      - name: commit formatted files
+        run: |
+          date > generated.txt
```
minusworld: What's the thinking behind this file?

enncoded: Oh that's a good point. That's not necessary. I'll remove that. I also noticed that the action fails when no files are changed, so I'll add a check for that too in my next push.

minusworld: 👍  Thanks!

minusworld:
I wanted to see if we could test the workflow before committing and found https://github.com/nektos/act.

I made a few modifications based on how the tests went.

minusworld (approved):

PR #1703, opened 2022-01-07
Leland matrixify technology

enncoded:
1. The `matrixify-graph` script now produces graphs that show owasp coverage based on both the `language` and `technology` metadata fields
2. Fixed inconsistent naming in CWE metadata fields
3. Fixed bugs in both of the scripts. The produced graphs are now much more accurate

### DIFF
--- a/matrixify_graph.py
+++ b/matrixify_graph.py
@@ -1,3 +1,10 @@
+#!/usr/bin/python3
+
+# Generates graphs showing our owasp coverage per language and technology
+# based on the json output from matrixify.py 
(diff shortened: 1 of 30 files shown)


matrixify_graph.py
```diff
@@ -6,15 +13,18 @@
 import requests
 import pandas as pd
 import matplotlib.pyplot as plt
+#import pprint
```
minusworld: Can remove this probably

minusworld (approved):
lgtm

PR #1705, opened 2022-01-08
new messages

colleend:
New messages for all rules that have run on semgrep-app. These have been detailed [here](https://linear.app/r2c/issue/RULES-244/reevaluate-rules-with-high-frequency-rule-findings).


c/lang/security/insecure-use-string-copy-fn.yaml
```diff
@@ -5,7 +5,7 @@ rules:
   - pattern: strncpy(...)
   message: >-
     Finding triggers whenever there is a strcpy or strncpy used.
-    This is an issue because strcpy or strncpy can lead to buffer overflow vulns.
+    This is an issue because strcpy or strncpy can lead to buffer overflow vulnerabilities and therefore crashes or executing arbitrary code.
```
minusworld: I think this sentence is a bit confusing.

"This is an issue because strcpy and strncpy do not check the boundaries of memory. This can lead to buffer overflows, which can cause program crashes and potentially lets an attacker to inject code into the program, compromising the system."

Maybe something like that?

typescript/react/security/audit/react-no-refs.yaml
```diff
@@ -7,13 +7,15 @@ rules:
   - pattern: createRef(...)
   - pattern: useRef(...)
   message: >-
-    `ref` usage found, refs give direct DOM access and may create a possibility for XSS
+    `ref` usage found, refs give direct DOM access and may create a possibility for XSS, which could cause account compromise. Instead, avoid direct DOM manipulation or use DOMPurify to sanitize HTML before writing it into the page.
```
minusworld: How does this cause account compromise?

colleend: I was thinking xss could lead to cookie compromise, which could possibly lead to session takeover (and wrote account compromise as a broader category of session takeover 😅 ). I'll say `which could cause sensitive information such as user cookies to be exfiltrated by an attacker` instead 👍

c/lang/security/insecure-use-string-copy-fn.yaml
colleend: sounds good!

minusworld (approved):

colleend:
🙏  sorry just saw the approval now, thanks for updating the formatting and fixing conflicts!

PR #1707, opened 2022-01-09
New rules to detect bad random

SPoint42:


SPoint42:
work on 0.77....
@minusworld ; qhat the hell on this error of unit tests ?

csharp/lang/security/crypto/random.cs
```diff
@@ -0,0 +1,6 @@
+var rnd = new Random();
```
minusworld, line 1: It looks like this line is missing the `// ruleid:` test annotation. This is probably the cause of the CI failures.

minusworld (changes requested):
I think there just needs to be a test annotation on the new rule's test file.

Also, it looks like there are a bunch of old commits that are also in this PR. Would you mind updating this PR to branch from the latest `develop` branch on semgrep-rules?

You can add the original repo as an upstream remote and pull in the latest changes. Then branch off that and add the new rule. As it is now, I'm worried about overwriting more recent changes with the old commits.

Thanks! Let me know if you have any questions.

SPoint42:
I will have a look to your comments

PR #1708, opened 2022-01-10
Fix semgrep-rule-lints to actually error

minusworld:
semgrep-rule-lints wasn't actually erroring due to lack of `--error` flag


colleend (approved):
ah, good catch!

PR #1710, opened 2022-01-10
Add metadata to php tainted-filename

minusworld:


enncoded (commented):
Should we also add A10:2021-Server-Side Request Forgery as an OWASP tag?

enncoded (approved):

PR #1714, opened 2022-01-11
Add an ocaml rule to detect useless comparisons

mjambon:
This was modeled after the existing rule `useless-eq` that we already have for ocaml... after getting [caught](https://github.com/returntocorp/semgrep/pull/4488#pullrequestreview-847602344) by @IagoAbal during code review.

### DIFF
--- a/ocaml/lang/correctness/useless-compare.ml
+++ b/ocaml/lang/correctness/useless-compare.ml
@@ -1,0 +1,14 @@
+let test a b =
+  (* ruleid:useless-compare *)
+  let c = compare (a+b) (a+b) in
+  if c <> 0 then c
+  else
+    compare a b
+
+let test a b =
+  (* ruleid:useless-compare *)
+  let c = Int.compare (a+b) (a+b) in
+  if c <> 0 then c
+  else
+    compare a b
+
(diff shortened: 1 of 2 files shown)


mjambon:
CI says
```
FileNotFoundError: [Errno 2] No such file or directory: '/usr/local/lib/python3.9/site-packages/semgrep/rule_schema.yaml'
```
:thinking:

IagoAbal (approved):
Don't know about the CI failure, but the rule looks good to me!

mjambon:
Here's an example of weird code where the rule produces a false positive:
```
utop # let count = let n = ref 0 in fun () -> incr n; !n;;
val count : unit -> int = <fun>

utop # compare (count ()) (count ());;
- : int = 1
```
^ almost always evaluates to `1` (or `-1` depending on argument evaluation order).

IagoAbal:
> Here's an example of weird code where the rule produces a false positive:

IMO that is fine, I don't think I would write the code that way. I would use intermediate `let`s to store the results of side-effectful functions.

aryx:
I know why, it's because it's symbolic link. I'll fix it.

aryx:
CI fix here: https://github.com/returntocorp/semgrep/pull/4498

colleend:
🥳 🥳 🥳

PR #1718, opened 2022-01-11
Only run push-dev-to-release on returntocorp repo

minusworld:
This will also run on forks: https://github.com/jamiees2/semgrep-rules/actions/runs/1683842680

Let's have it only run on our repo


colleend (approved):
ah good catch, thanks for fixing!

PR #1729, opened 2022-01-13
Add initial YAML breakdown of rule maturity framework

clintgibler:
This is a WIP PR for us to discuss how we want to proceed on.


enncoded (approved):

scripts/cwe_to_metacategory.yml
```diff
@@ -0,0 +1,118 @@
...
+      - "CWE-352: Cross-Site Request Forgery (CSRF)"
+    XSS:
+      - "CWE-79: Improper Neutralization of Input During Web Page Generation ('Cross-site Scripting')"
+      - "CWE-80: Improper Neutralization of Script-Related HTML Tags in a Web Page"
+      - "CWE-279: Improper Neutralization of Input During Web Page Generation ('Cross-site Scripting')"
+      - "CWE-116: Improper Encoding or Escaping of Output"
+    "SQL Injection":
+      - "CWE-89: Improper Neutralization of Special Elements used in an SQL Command ('SQL Injection')"
+    "Command Injection":
+      - "CWE-78: Improper Neutralization of Special Elements used in an OS Command ('OS Command Injection')": 24,
```
zyannes: Should this line have a value (`24`)?

clintgibler: It should not, fixed in latest push 🙏

stats/matrixify.py
```diff
@@ -92,12 +89,25 @@ def get_framework(path: str, rule: Dict[str, Any]) -> str:
     # if lang == hcl:
     return s[s.index(lang)+1].strip()
 
+# Reads 'cwe_to_metacategory.yml' to construct a map to convert a CWE to a metacategory
+def create_metacategory_map(path: str) -> Dict[str, str]:
```
zyannes, line 93: Nice! This will be really helpful to avoid manually updating in the backend.

stats/matrixify.py
```diff
@@ -92,12 +89,25 @@ def get_framework(path: str, rule: Dict[str, Any]) -> str:
...
+    cwe_mc_map = {} # {cwe, metacategory}
+
+    with open(path, 'r') as mc_map_file:
+        mc_map = yaml.safe_load(mc_map_file)
+
+        for mc in mc_map:
+            for cwe in mc_map[mc]:
+                cwe_mc_map[cwe] = mc
```
zyannes: super-nit: Can be replaced with the following one-liner
```suggestion
    return {cwe: mc for mc, cwes in mc_map for cwe in cwes}
```

stats/matrixify.py
```diff
@@ -92,12 +89,25 @@ def get_framework(path: str, rule: Dict[str, Any]) -> str:
...
+
+    return cwe_mc_map
+
 def is_security(path: str) -> bool:
     return "security" in path
 
 def is_rule(path: str) -> bool:
     _, ext = os.path.splitext(path)
-    return "yaml" in ext or "yml" in ext
+    return ("yaml" in ext or "yml" in ext) and '/scripts/' not in path
```
zyannes: nit: Although unlikely, the parenthesized conditions could match extensions that are not explicitly "yaml" or "yml". Maybe consider the following
```suggestion
    return ext in (".yaml", ".yml") and '/scripts/' not in path
```

clintgibler: Nice, thanks!

stats/matrixify.py
clintgibler: Ah neat. I like that it's more concise, but to be honest is harder for me to read and reason about :/ Will accept though 👍

PR #1730, opened 2022-01-14
four brakeman rules

colleend:
Four rules correlates to:
* [check_detailed_exceptions.rb](https://github.com/presidentbeef/brakeman/blob/main/lib/brakeman/checks/check_detailed_exceptions.rb)
* [check_dynamic_finders.rb](https://github.com/presidentbeef/brakeman/blob/main/lib/brakeman/checks/check_dynamic_finders.rb)
* [check_mail_to.rb](https://github.com/presidentbeef/brakeman/blob/main/lib/brakeman/checks/check_mail_to.rb)

_Test Plan:_
* check that test-develop passes, and that `semgrep -f <rule name.yaml> <rule name.rb>` passes the `ok` and `ruleid` tests.

### DIFF
--- a/ruby/rails/security/audit/dynamic-finders.yaml
+++ b/ruby/rails/security/audit/dynamic-finders.yaml
@@ -1,0 +1,29 @@
+rules:
+- id: dynamic-finders 
+  metadata:
+    owasp: 'A1: Injection'
+    cwe: "CWE-89: Improper Neutralization of Special Elements used in an SQL Command"
+    source-rule-url: https://github.com/presidentbeef/brakeman/blob/main/lib/brakeman/checks/check_dynamic_finders.rb 
+    category: security
+    technology:
+    - rails
+  message: >-
+    Discovered an application that uses MySQL and find_by_* dynamic finders 
+    on potentially sensitive fields. This could lead to 0=0 comparisons where queries will always return true,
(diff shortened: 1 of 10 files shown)


ruby/rails/security/audit/dynamic-finders.yaml
```diff
@@ -0,0 +1,29 @@
...
+  metadata:
+    owasp: 'A1: Injection'
+    cwe: "CWE-89: Improper Neutralization of Special Elements used in an SQL Command"
+    source-rule-url: https://github.com/presidentbeef/brakeman/blob/main/lib/brakeman/checks/check_dynamic_finders.rb 
+    category: security
+    technology:
+    - rails
+  message: >-
+    Discovered an application that uses MySQL and find_by_* dynamic finders 
+    on potentially sensitive fields. This could lead to 0=0 comparisons where queries will always return true,
```
minusworld: How does this lead to 0=0 comparisons?

minusworld (changes requested):
One small question on a rule message

ruby/rails/security/audit/dynamic-finders.yaml
colleend: The vulnerability is in MySQL integer conversion, which could cause "0" to match any string -- I'll write that instead, as it seems to explain the situation better

minusworld (approved):

PR #1733, opened 2022-01-14
Adding dependency check rule

zzzteph:
The rule for searching dependencies in packages that are not under the scope and have variant versions. Without proper configuration (package-json, etc.) it may lead in some cases to dependency confusion vulnerability or usage of 3d party malicious code.


zzzteph:
Anything wrong with test JSON?

minusworld:
Hey @zzzteph, the issue here is that the rule uses the `paths` key which means it only searches files named `package.json`, and the test file is named after the rule.

This is a problem we've had for a while now but haven't had a chance to fix yet... 😓 

The easiest solution is a hacky one... You can change the `paths` key to search `*package.json` and name the test file to be `ruleid.package.json`, which I think will work. If you don't get to it in the next hour, I will try this after I finish up what I'm working on.

zzzteph:
> Hey @zzzteph, the issue here is that the rule uses the `paths` key which means it only searches files named `package.json`, and the test file is named after the rule.
> 
> This is a problem we've had for a while now but haven't had a chance to fix yet... 😓
> 
> The easiest solution is a hacky one... You can change the `paths` key to search `*package.json` and name the test file to be `ruleid.package.json`, which I think will work. If you don't get to it in the next hour, I will try this after I finish up what I'm working on.

If you can do this, it will be awesome🙂 But it's, as you said - a hack). I checked the error with the JSON validator and seems error in the JSON parser, that makes usage of 
//comment 
will make JSON not valid. I'm not sure, but maybe it's a problem?

minusworld:
@zzzteph Looks like the file path was all it needed.

Once the license/cla is signed we can merge this in! Thanks!

zzzteph:
> @zzzteph Looks like the file path was all it needed.
> 
> Once the license/cla is signed we can merge this in! Thanks!

@minusworld  seems I forget to add an email to the initial pull request, and CLA will remain in pending status.  Maybe need to reopen the pull request with a valid configuration.

colleend:
Hi! That sounds good -- perhaps you could try the steps in [this](https://r2c-community.slack.com/archives/CK86BJ5DW/p1642497219072400) thread (thanks for bringing that up in the Slack), and if that doesn't work then we can close this PR and reopen another PR with an email added to it 👍

zzzteph:
Hi @minusworld @colleend, tried something with "rebase" magic, hope it helps. At least cla is signed right now.

minusworld (approved):

minusworld:
Hmm, someone else just reported an issue that might be related to the failing tests: https://github.com/returntocorp/semgrep/issues/4540

😢 

We'll investigate and figure out what's next.

colleend:
hmm, this might actually be due to the `paths` include key in the file (I think tests are currently configured to check that the test file adheres to the schema followed by the `paths` include key). I'll see if I can try to fix this sometime this week.

zzzteph:
> If you can do this, it will be awesome🙂 But it's, as you said - a hack). I checked the error with the JSON validator and seems error in the JSON parser, that makes usage of
> //comment
> will make JSON not valid. I'm not sure, but maybe it's a problem?

Sure, thank you very much!


> If you can do this, it will be awesome🙂 But it's, as you said - a hack). I checked the error with the JSON validator and seems > error in the JSON parser, that makes usage of
> will make JSON not valid. I'm not sure, but maybe it's a problem?

I'm not sure but when the test filename(**dependency_check.json**=>**dependency_check.package.json**) was changed, the test succeeded.

minusworld:
Aha, there we go. Merging. Thanks @zzzteph !

PR #1737, opened 2022-01-17
Detect SQL injection with Laravel ORM methods

Sjord:
Currently, some tests fail. 

This doesn't match while it should. It doesn't match because `addSelect` is called on `$query` and not on `DB::table`. Does semgrep know that `$query` originates from `DB::table`?

```
// ruleid: laravel-sql-injection
$users = $query->addSelect($tainted)->get();
```

This matches while it shouldn't. I don't understand why, because we match only the first parameter as tainted:

```
// ok: laravel-sql-injection
$user = DB::table('users')->where('name', $tainted)->first();
```

Same here:

```
// ok: laravel-sql-injection
$orders = DB::table('orders')->selectRaw('price * ? as price_with_tax', [$tainted]);
```


colleend:
Hi Sjord! Thanks for creating this rule, we really appreciate it 😄 . 

For 

```
// ruleid: laravel-sql-injection
$users = $query->addSelect($tainted)->get();
```
Semgrep won't take $query to be `DB::table` -- you would have to add in something like 
```
          $Q = DB::table(...)->select($COLUMN);
          ...
          $USERS = $Q->addSelect($COLUMN)->get();
```
in order to get that to match. 

For

```
// ok: laravel-sql-injection
$user = DB::table('users')->where('name', $tainted)->first();
``` 
the `pattern-inside: DB::table(...)->where($COLUMN, ...)` actually matches this (I think in PHP it doesn't matter where the ellipsis is, it matches everything left and right of the ellipsis). This could definitely be classified as a bug -- feel free to file this as an issue (or I can file it for you!). 

The same issue applies for 

```
// ok: laravel-sql-injection
$orders = DB::table('orders')->selectRaw('price * ? as price_with_tax', [$tainted]);
```
the ellipsis for `- pattern-inside: DB::table(...)->selectRaw($EXPRESSION, ...)` will match everything left and right of the metavariable. 

Feel free to tag me in the issue filed or tell me if you want me to file the issue!

Sjord:
> the pattern-inside: DB::table(...)->where($COLUMN, ...) actually matches this

The problem was that I had multiple variables as sinks, which doesn't work as I expected. My rule has this:

```
  pattern-sinks:
  - patterns:
    - pattern-either:
      - pattern: $SQL
      - pattern: $EXPRESSION
      - pattern: $COLUMN
      - pattern: $COLUMNS
    - pattern-either:
      - pattern-inside: DB::table(...)->whereRaw($SQL, ...)
```

The rule `DB::table(...)->whereRaw($SQL, ...)` uses an `$SQL` variable, but is compared with all of the sinks, not only the `$SQL` sink. So semgrep also evaluates the rule with the `$EXPRESSION` sink. `$EXPRESSION` is not captured, so it'll match anything.

colleend:
Ah, I see, that makes sense. Thanks for fixing this, and good catch! We've been making sure that there's adequate metadata in all our rules, so we've been requiring the 'technology' and 'category' fields in metadata to be filled out 👍

colleend (approved):
Thanks so much for putting in the time to add this rule! We appreciate it 🙏

PR #1743, opened 2022-01-20
remove --quiet for better testing errors

colleend:
Removing quiet from from the `semgrep-rules-test-develop` workflow. 

This leads to better errors if `test.py` itself fails. 

_To Test:_
* change the name of a semgrep-rule rule id to something that doesn't exist, like 'dne'.
* run `semgrep --test <folder file exists in>`
* then run `semgrep --test --quiet <folder file exists in>`

You can see that `--quiet` silences the error.


minusworld (approved):

ievans:
thank you @colleend !

PR #1748, opened 2022-01-25
Leland rails rules

enncoded:


colleend:
Would you mind adding the brakeman link (such as https://github.com/presidentbeef/brakeman/blob/main/lib/brakeman/checks/check_header_dos.rb) as a `source-rule-url` in metadata?) (You can reference this here: https://github.com/returntocorp/semgrep-rules/commit/2cc7a752dcf62ab7a6bce35d3c2feff8e48ecc1a)

ruby/rails/security/audit/rails-check-page-caching-cve.yaml
```diff
@@ -0,0 +1,52 @@
+rules:
```
colleend, line 1: For this rule, we also want a ruby rule that catches instances of code like 
```
class BooksController < ApplicationController
caches_page :show
end
```
(where show is vulnerable).
Good work on catching the page-caching gems though 👍

ruby/rails/security/audit/rails-check-render-dos.yaml
```diff
@@ -0,0 +1,52 @@
+rules:
```
colleend: same for this one: we would also want to catch cases of `render :text` 👍

ruby/rails/security/audit/rails-check-response-splitting.yaml
```diff
@@ -0,0 +1,33 @@
+rules:
+  - id: rails-check-response-splitting
```
colleend, line 2: nice!

ruby/rails/security/injection/rails-check-json-parsing-rce.yaml
```diff
@@ -0,0 +1,67 @@
+rules:
```
colleend, line 1: this one is only for CVE-2013-0333 and not for CVE-2013-0269 right?

colleend (changes requested):
Thanks for doing this!

ruby/rails/security/injection/rails-check-json-parsing-rce.yaml
enncoded: Yes, I'll be working on one for 0269 next when I have time to do more Brakeman rules!

colleend (approved):
added in the references to brakeman, lgtm!

PR #1754, opened 2022-01-28
Improve owasp crypto, random, and hash statistics

colleend:
This gets the semgrep score for the OWASP categories:
* Weak Encryption Algorithm -> 100%
* Weak Hashing Algorithm -> 69%
* Weak Randomness -> 100%

To test this, run `semgrep --test .` on the crypto folder and see that all tests pass.

Then, run 
`semgrep -f semgrep-rules/java/lang/security/audit/crypto/ --json > results/Benchmark_1.2-semgrep.json` in the `BenchmarkJava` repository, and then
`createScorecards.sh`

You will see something like:
Screen Shot 2022-01-27 at 4 32 29 PM

_PS:_
The reason why hashes aren't at 100% is because they have this little snippet:
```
String algorithm = benchmarkprops.getProperty("hashAlg1", "SHA512");
java.security.MessageDigest md = java.security.MessageDigest.getInstance(algorithm);         
```
where the `getProperty` function returns `MD5`. tricky!!


inkz (approved):
lgtm

PR #1760, opened 2022-02-01
Leland rule feed initial 5

enncoded:


javascript/ecdsa/security/audit/ecdsa-signature-forgery.yaml
```diff
@@ -0,0 +1,30 @@
+rules:
+  - id: ecdsa-signature-forgery
+    patterns:
+      - pattern-either:
+          - pattern-inside: |
```
kurt-r2c, line 5: Are these meant to be two separate `pattern-inside` rules?  This reads to me like you could potentially have:

```
var SomeLib = require('definitely-not-starkbank-ecdsa')
var ECDSA =  SomeLib.Ecdsa
ECDSA.verify(someArgs)
```

Which would then match as long as the `starkbank-ecdsa` library was present somewhere in the project

javascript/node-fetch/security/audit/node-fetch-cookie-leak.yaml
```diff
@@ -0,0 +1,42 @@
...
+    patterns:
+      - pattern-either:
+        - pattern: |
+            fetch(..., { ..., headers: {..., 'Cookie': ..., ...}, ... })
+        - pattern: |
+            const $HEADERS = {..., 'Cookie': ..., ...}
+            ...
+            fetch(..., { headers: $HEADERS }, ...)
+        - pattern: |
+            const $OPTIONS = {..., headers: $HEADERS, ...}
```
kurt-r2c, line 12: this case doesn't guarantee cookies are sent, does it?  While $HEADERS might be set in one of the other `pattern-either` definitions, this pattern can match `options` with no `Cookie` key.  We should add a test case for same.

kurt-r2c (changes requested):

javascript/ecdsa/security/audit/ecdsa-signature-forgery.yaml
enncoded: Good catch! I'll fix that

javascript/node-fetch/security/audit/node-fetch-cookie-leak.yaml
enncoded: Fixed! I'll send you the updated rule and test on slack

PR #1761, opened 2022-02-01
add `net-webconfig-debug` rule for ASP.NET

inkz:
ref: https://github.com/github/codeql/tree/main/csharp/ql/src/Security%20Features/CWE-011


colleend (approved):
lgtm!

PR #1763, opened 2022-02-01
Java try-with-resources now matches

emjin:
Update this test now that Java's try-with-resources matches


colleend (approved):
Emma's change in semgrep-core will make this match -- approving because otherwise semgrep-core tests won't pass 👍 (yay circular dependencies)

PR #1765, opened 2022-02-02
Fix OCaml rules involving incr and decr

### LINKED ISSUE #1762: ocaml.lang.best-practice.ref.ocamllint-ref-incr uses incorrect pattern
**Describe the bug**
The current pattern for the rule is `$X := $X + 1`, but that's now OCaml `ref` and `:=` work. The current test for it `x := x + 1` isn't even well-typed! If `$X` is a ref to be assigned, then on the right-hand side it needs to be dereferenced as `!$X`.
For some reason the rule `$X := !$X + 1` doesn't find anything though.

**To Reproduce**
https://semgrep.dev/s/DyQP

**Expected behavior**
The rule should have the pattern `$X := !$X + 1` and it should find the match.

**Priority**
How important is this to you?

mjambon:
Fixes https://github.com/returntocorp/semgrep-rules/issues/1762

### DIFF
--- a/ocaml/lang/best-practice/ref.ml
+++ b/ocaml/lang/best-practice/ref.ml
@@ -1,6 +1,6 @@
 let test x =
   (* ruleid:ocamllint-ref-incr *)
-  x := x + 1;
+  x := !x + 1;
   (* ruleid:ocamllint-ref-decr *)
-  x := x - 1;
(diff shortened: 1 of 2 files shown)


inkz (approved):
lgtm

PR #1767, opened 2022-02-02
add SslCertificateTrust configuration correctness rule + tests

kurt-r2c:
@clintgibler @minusworld @lejliu 

Background:

MSDN docs for `CreateForX509Collection` and `CreateForX509Store` discourage setting `sendTrustInHandshake=true`

https://docs.microsoft.com/en-us/dotnet/api/system.net.security.sslcertificatetrust.createforx509collection?view=net-6.0#remarks
https://docs.microsoft.com/en-us/dotnet/api/system.net.security.sslcertificatetrust.createforx509store?view=net-6.0#remarks

This is an easy win for C# correctness coverage, even if a cursory grep.app turns up 0 public-facing usage of this static method.

### DIFF
--- a/csharp/lang/correctness/sslcertificatetrust/sslcertificatetrust-handshake-no-trust.yaml
+++ b/csharp/lang/correctness/sslcertificatetrust/sslcertificatetrust-handshake-no-trust.yaml
@@ -12,7 +12,8 @@ rules:
   languages: [csharp]
   metadata:
     references: 
-      - https://docs.microsoft.com/en-us/dotnet/api/system.net.security.sslcertificatetrust.createforx509collection?view=net-6.0
+      - https://docs.microsoft.com/en-us/dotnet/api/system.net.security.sslcertificatetrust.createforx509collection?view=net-6.0#remarks
+      - https://docs.microsoft.com/en-us/dotnet/api/system.net.security.sslcertificatetrust.createforx509store?view=net-6.0#remarks
     cwe: "CWE-200: Exposure of Sensitive Information to an Unauthorized Actor"
     owasp: "A3: Sensitive Data Exposure"
     category: correctness
--- a/csharp/lang/correctness/sslcertificatetrust/sslcertificatetrust-handshake-no-trust.yaml
+++ b/csharp/lang/correctness/sslcertificatetrust/sslcertificatetrust-handshake-no-trust.yaml
@@ -7,7 +7,7 @@ rules:
     - metavariable-regex:
         metavariable: $METHOD
         regex: CreateForX509(Collection|Store)
-  fix: $METHOD($COLLECTION,false)
+  fix: SslCertificateTrust.$METHOD($COLLECTION,false)
   message: Sending the trusted CA list increases the size of the handshake request and can leak system configuration information.
   languages: [csharp]
   metadata:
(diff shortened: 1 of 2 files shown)


clintgibler:
Nice! Since this is recommending a fairly straightforward argument change, want to try adding an autofix?

kurt-r2c:
Now with 100% more autofix!

enncoded:
Should we also add https://docs.microsoft.com/en-us/dotnet/api/system.net.security.sslcertificatetrust.createforx509store?view=net-6.0#remarks to the `references` section in the metadata?

csharp/lang/correctness/sslcertificatetrust/sslcertificatetrust-handshake-no-trust.yaml
```diff
@@ -0,0 +1,23 @@
+rules:
+- id: correctness-sslcertificatetrust-handshake-no-trust
+  patterns:
+    - pattern-either:
+      - pattern: SslCertificateTrust.$METHOD($COLLECTION,sendTrustInHandshake=true)
+      - pattern: SslCertificateTrust.$METHOD($COLLECTION,true)
+    - metavariable-regex:
+        metavariable: $METHOD
+        regex: CreateForX509(Collection|Store)
+  fix: $METHOD($COLLECTION,false)
```
clintgibler: Does this need to be `SslCertificateTrust.$METHOD($COLLECTION,false)`? I'm not sure, test locally and see what happens.

csharp/lang/correctness/sslcertificatetrust/sslcertificatetrust-handshake-no-trust.yaml
```diff
@@ -0,0 +1,23 @@
...
+  message: Sending the trusted CA list increases the size of the handshake request and can leak system configuration information.
+  languages: [csharp]
+  metadata:
+    references: 
+      - https://docs.microsoft.com/en-us/dotnet/api/system.net.security.sslcertificatetrust.createforx509collection?view=net-6.0#remarks
+      - https://docs.microsoft.com/en-us/dotnet/api/system.net.security.sslcertificatetrust.createforx509store?view=net-6.0#remarks
```
clintgibler, line 16: Looking at both of these docs, it appears that both methods have exactly 2 args, is that the case? I just want to make sure that $COLLECTION will be sufficient, as if there are >2 potential args then it won't match some calls.

clintgibler (approved):
Maybe sure autofix pattern works then 🚢  it!

csharp/lang/correctness/sslcertificatetrust/sslcertificatetrust-handshake-no-trust.yaml
kurt-r2c: It does, playground didn't replace correctly until I fixed this.

csharp/lang/correctness/sslcertificatetrust/sslcertificatetrust-handshake-no-trust.yaml
kurt-r2c: Yes, these methods have exactly two args, though I suppose `$COLLECTION` might not capture an inline collection instantiation?

PR #1768, opened 2022-02-03
SQLi rules for Scala/Slick

inkz:
from: https://find-sec-bugs.github.io/bugs.htm#SCALA_SQL_INJECTION_SLICK

### DIFF
--- a/scala/slick/security/scala-slick-sql-non-literal.yaml
+++ b/scala/slick/security/scala-slick-sql-non-literal.yaml
@@ -14,6 +14,8 @@ rules:
   severity: WARNING
   metadata:
     source-rule-url: https://find-sec-bugs.github.io/bugs.htm#SCALA_SQL_INJECTION_SLICK
+    references:
+    - https://scala-slick.org/doc/3.3.3/sql.html#splicing-literal-values
     category: security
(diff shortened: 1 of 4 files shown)


scala/slick/security/scala-slick-sql-non-literal.yaml
```diff
@@ -0,0 +1,22 @@
...
+  - pattern-inside: |
+      import slick.$DEPS
+      ...
+  message: >-
+      Detected a formatted string in a SQL statement. This could lead to SQL
+      injection if variables in the SQL statement are not properly sanitized.
+      Avoid using `#$variable` and use `$variable` in `sql"..."` strings instead.
+  languages: [scala]
+  severity: WARNING
+  metadata:
```
kurt-r2c, line 15: maybe add a `references` here to https://scala-slick.org/doc/3.3.3/sql.html#splicing-literal-values to explain why the `pattern-regex` is `\#\$`?  I am new to Scala and having the "why" attached to the rule would be useful to me as a new Scala dev.

kurt-r2c (approved):
One non-blocking comment, LGTM otherwise

scala/slick/security/scala-slick-sql-non-literal.yaml
inkz: yep, cool idea, thanks :+1:

PR #1771, opened 2022-02-03
add Double.Epsilon correctness rule & tests

kurt-r2c:
@clintgibler @lejliu 

fyi @minusworld @inkz

Rule for equality comparisons using `Double.Epsilon` for non-zero `Double` values which is discouraged by .NET documentation: https://docs.microsoft.com/en-us/dotnet/api/system.double?view=net-6.0#testing-for-equality


csharp/lang/correctness/double/double-epsilon-equality.yaml
```diff
@@ -0,0 +1,30 @@
...
+    - pattern-not-inside: |
+        double $V2 = 0;
+        ...
+    - pattern-not-inside: |
+        $V1 = 0;
+        ...
+    - pattern-not-inside: |
+        $V2 = 0;
+        ...
+  message: Double.Epsilon is defined by .NET as the smallest value that can be added to or subtracted from a zero-value Double.  It is unsuitable for equality comparisons of non-zero Double values.
```
minusworld: If I want to compare non-zero Double values, what should I use instead?

kurt-r2c: You should use .NET's built-in equality rather than re-implementing it!  `Double.Epsilon`'s value is actually different across frameworks and systems (on ARM such as the M1, it's actually 0) - https://docs.microsoft.com/en-us/dotnet/api/system.double.epsilon?view=net-6.0#platform-notes

Custom double equality is "here there be dragons" behavior and should be avoided at all costs

minusworld: I assume this is the usual <, >, etc.? Can we add that as part of the message?

kurt-r2c: @minusworld I think latest commit handles your comments, please verify

minusworld: Perfect 👍🏼

minusworld (approved):

PR #1774, opened 2022-02-04
sqli and cmdi rules

colleend:
SQLi rule and CMDi rule for OWASP java benchmark -- this improves the score from: 
* 2%-> 12% for cmdi
* 0% -> 18% for sqli

_To test:_
Run `semgrep --test .` on the `java/lang/security/audit` folder and check that all tests pass.

Run 
`semgrep -f ~/Desktop/semgrep_repos/semgrep-rules/java/lang/security/audit --json > results/Benchmark_1.2-semgrep.json`
`createScorecards.sh`

and check the scorecards for the benchmark.

### DIFF
--- a/java/lang/security/audit/tainted-cmd-from-http-request.yaml
+++ b/java/lang/security/audit/tainted-cmd-from-http-request.yaml
@@ -1,0 +1,54 @@
+rules:
+  - id: tainted-cmd-from-http-request 
+    message: >-
+      Detected input from a HTTPServletRequest going into a 'ProcessBuilder' or 'exec' command. 
+      This could lead to command injection if variables passed into the exec commands are not properly sanitized. 
+      Instead, avoid using these OS commands with user-supplied input, or, if you must use these commands, use a whitelist of specific values.
+    languages: [java]
+    severity: WARNING
+    mode: taint
+    pattern-sources:
+    - patterns:
+      - pattern-either:
+        - pattern: |
+            (HttpServletRequest $REQ)
+        - patterns: # this pattern is a hack to get the rule to recognize `map` as tainted source when `cookie.getValue(user_input)` is used.
+          - pattern-inside: |
+              (javax.servlet.http.Cookie[] $COOKIES) = (HttpServletRequest $REQ).getCookies(...);
+              ...
+              for (javax.servlet.http.Cookie $COOKIE: $COOKIES) {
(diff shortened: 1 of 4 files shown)


java/lang/security/audit/tainted-cmd-from-http-request.yaml
```diff
@@ -0,0 +1,54 @@
...
+            (HttpServletRequest $REQ)
+        - patterns: # this pattern is a hack to get the rule to recognize `map` as tainted source when `cookie.getValue(user_input)` is used.
+          - pattern-inside: |
+              (javax.servlet.http.Cookie[] $COOKIES) = (HttpServletRequest $REQ).getCookies(...);
+              ...
+              for (javax.servlet.http.Cookie $COOKIE: $COOKIES) {
+                ...
+              }
+          - pattern: |
+              $COOKIE.getValue(...)
```
inkz, line 23: why sources described differently while both rules tackling HTTP requests? can it be merged together? but I am out of context here, so just asking.
Amazing job :raised_hands:

inkz (approved):

java/lang/security/audit/tainted-cmd-from-http-request.yaml
colleend: good question! for the sqli rule, I kept getting a large number of FP when using just (HttpServletRequest $REQ) -- most likely because there were a lot of test cases where the source would go into a sink we couldn't catch due to tricky code meant to confuse taint analysis 😓 . However, for the cmdi rule, there were many TP cases that had something like `(getValueFrom(request))`, and therefore using `(HttpServletRequest $REQ)` led to a better score (there's not so many tricky FP that taint catches).

minusworld: Will we miss anything in real code by using the more-specific sources?

kurt-r2c: I think we should bias for fewer FPs, even if it means missing some black swan TPs.

colleend: It's definitely possible that we'll miss something in real code, but most of the cases should be caught (as most of the requestservlets from java I've seen on github are used like `req.<func>()`). +1 to kurt's point 👍

PR #1783, opened 2022-02-08
Fix pipenv rule written for pip

underyx:
verified that the original suggestion is incorrect:

```
❯ pip install --ignore-pipfile

Usage:   
  pip install [options] <requirement specifier> [package-index-options] ...
  pip install [options] -r <requirements file> [package-index-options] ...
  pip install [options] [-e] <vcs project url> ...
  pip install [options] [-e] <local project path> ...
  pip install [options] <archive url/path> ...

no such option: --ignore-pipfile
```

### DIFF
--- a/generic/ci/security/use-frozen-lockfile.generic
+++ b/generic/ci/security/use-frozen-lockfile.generic
@@ -5,7 +5,7 @@ WORKDIR /app
 RUN yarn install
 # trailing space
 # ruleid: use-frozen-lockfile-yarn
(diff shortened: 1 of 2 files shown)


ievans (approved):
Should't we detect `pip install` and suggest `pipenv install`?

underyx:
> Should't we detect `pip install` and suggest `pipenv install`?

Might be too much effort to have a reasonable fix rate on that :/

PR #1790, opened 2022-02-09
Leland cs rules

enncoded:


csharp/lang/security/cryptography/X509Certificate2-privkey.yaml
```diff
@@ -0,0 +1,35 @@
...
+      references:
+        - https://docs.microsoft.com/en-us/dotnet/api/system.security.cryptography.x509certificates.x509certificate2.privatekey
+      category: security
+      technology:
+        - .net
+      confidence: MEDIUM
+    message: >-
+      X509Certificate2.PrivateKey is obsolete. Use a method such as GetRSAPrivateKey() or GetECDsaPrivateKey().
+      Alternatively, use the CopyWithPrivateKey() method to create a new instance with a private key. Further,
+      if you set this field to `null` or set it to another key without deleting it first, the private key will
```
minusworld: Which field?

csharp/lang/security/memory/memory-marshal-create-span.yaml
```diff
@@ -0,0 +1,22 @@
+rules:
+  - id: memory-marshal-create-span
+    severity: WARNING
+    languages:
+      - C#
+    metadata:
+      cwe:
+        - "CWE-125: Out-of-bounds Read"
```
minusworld, line 8: This is OK, but a warning is that I think up 'til now CWE has always been a string. I think this will cause issues in the app

csharp/lang/security/http/http-listener-wildcard-bindings.yaml
```diff
@@ -0,0 +1,27 @@
...
+        - "A05:2017 - Broken Access Control"
+        - "A01:2021 - Broken Access Control"
+      references:
+        - https://docs.microsoft.com/en-us/dotnet/api/system.net.httplistener?view=net-6.0 
+      category: security
+      technology:
+        - .net
+      confidence: MEDIUM
+    message: >-
+      Prefix: $PREFIX
```
minusworld: What's the issue here?

csharp/lang/security/memory/memory-marshal-create-span.yaml
enncoded: I remember seeing some rules that had multiple CWEs when I was working on the matrixify script when I first started. What would happen in the app?

minusworld: Just based on previous experience, I think it will cause issues much like the OWASP update. I don't currently see any other list-based CWEs. I think this highlights the need for a metadata schema for rules in this repository. Will talk to some others about this

csharp/lang/security/http/http-listener-wildcard-bindings.yaml
```diff
@@ -0,0 +1,28 @@
...
+        - "A01:2021 - Broken Access Control"
+      references:
+        - https://docs.microsoft.com/en-us/dotnet/api/system.net.httplistener?view=net-6.0
+      category: security
+      technology:
+        - .net
+      confidence: MEDIUM
+    message: >-
+      The top level wildcard bindings $PREFIX leaves your application open to security vulnerabilities and give attackers more control over where traffic is routed.
+      If you must use wildcards, consider using subdomain wildcard binding. For example, you can use "*.asdf.gov" if you own all of "asdf.gov".
```
minusworld, line 19: Nice :D

minusworld (approved):

PR #1798, opened 2022-02-09
standard lib XML parser XXE rules

kurt-r2c:
reference: https://www.jardinesoftware.net/2016/05/26/xxe-and-net/

Rules for the "big 3" .NET parsers:

- `XmlTextReader`
- `XmlReaderSettings` as related to `XmlReader`
- `XmlDocument`


inkz (approved):
lgtm :+1:

PR #1800, opened 2022-02-10
add AWS Lambda SQLi rules for Python

inkz:
tainted SQLi rules for:
* sqlalchemy
* pymysql
* pymssql
* mysql
* psycopg2


kurt-r2c (approved):
These are awesome.  Thanks Amazon for enforcing a standard Lambda entry point.

PR #1804, opened 2022-02-10
add basic path traversal rules for Path.Combine

kurt-r2c:
taint rule for public method string arguments passed to `Path.Combine` without either overwriting with `Path.GetFileName` or comparison of `Path.GetFileName` with the original string.

### DIFF
--- a/csharp/lang/security/filesystem/unsafe-path-combine.yaml
+++ b/csharp/lang/security/filesystem/unsafe-path-combine.yaml
@@ -1,0 +1,40 @@
+rules:
+  - id: unsafe-path-combine
+    mode: taint
+    pattern-sources:
+      - patterns:
+          - pattern: $ARG
+          - pattern-inside: |
+                public $T $M(...,string $ARG,...){...}
+    pattern-sinks:
+      - patterns: 
+        - pattern: $ARG
+        - pattern-inside: |
+            Path.Combine(...)
+    pattern-sanitizers:
+      - patterns:
+        - pattern: $ARG
+        - pattern-either:
+          - pattern-inside: |
+              $ARG = Path.GetFileName(...);
+              ...
+          - pattern-inside: |
+              <... Path.GetFileName(...) != $ARG ...>;
+              ...
+    message: String method argument $ARG passed to Path.Combine without sanitization.
+    languages:
+      - csharp
+    severity: WARNING
+    metadata:
+      category: security
+      license: MIT
+      references:
+        - https://www.praetorian.com/blog/pathcombine-security-issues-in-aspnet-applications/
+        - https://docs.microsoft.com/en-us/dotnet/api/system.io.path.combine?view=net-6.0#remarks
+      technology:
+        - .net
+      cwe:
+        - "CWE-22: Improper Limitation of a Pathname to a Restricted Directory ('Path Traversal')"
+      owasp:
+        - A05:2017 - Broken Access Control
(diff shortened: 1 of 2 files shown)


csharp/lang/security/filesystem/unsafe-path-combine.yaml
```diff
@@ -0,0 +1,40 @@
...
+            Path.Combine(...)
+    pattern-sanitizers:
+      - patterns:
+        - pattern: $ARG
+        - pattern-either:
+          - pattern-inside: |
+              $ARG = Path.GetFileName(...);
+              ...
+          - pattern-inside: |
+              <... Path.GetFileName(...) != $ARG ...>;
```
kurt-r2c, line 24: PA team has pointed out that this approach is quite slow - maybe we just accept that we don't catch this case?

minusworld: Could we measure how slow? Is it like, screeching halt kind of slow? Or step away to make coffee kind of slow?

csharp/lang/security/filesystem/unsafe-path-combine.yaml
```diff
@@ -0,0 +1,40 @@
...
+              ...
+    message: String method argument $ARG passed to Path.Combine without sanitization.
```
minusworld: Why is this bad? I assume it's a path traversal, but others might not know that

csharp/lang/security/filesystem/unsafe-path-combine.yaml
kurt-r2c: it's "get a coffee" slow, but this is uselessly noisy on a real project:

```
kboberg@Kurts-Air Squirrel.Windows % git ls-files | xargs cat | wc -l
cat: test/fixtures/packages/Castle.Core.3.2.0/ASL: No such file or directory
cat: Apache: No such file or directory
cat: Software: No such file or directory
cat: Foundation: No such file or directory
cat: License.txt: No such file or directory
cat: vendor/nuget: Is a directory
  752284
```

```
PS /Users/kboberg/source/Squirrel.Windows> measure-command {semgrep --config s/returntocorp:unsafe-path-combine}
Fetching rules from https://semgrep.dev/registry ...
Running 1 rules...
ran 1 rules on 43 files: 21 findings

Days              : 0
Hours             : 0
Minutes           : 0
Seconds           : 2
Milliseconds      : 441
Ticks             : 24417182
TotalDays         : 2.82606273148148E-05
TotalHours        : 0.000678255055555556
TotalMinutes      : 0.0406953033333333
TotalSeconds      : 2.4417182
TotalMilliseconds : 2441.7182
```

csharp/lang/security/filesystem/unsafe-path-combine.yaml
kurt-r2c: Good point, this message is bad, actually.  Will fix

csharp/lang/security/filesystem/unsafe-path-combine.yaml
minusworld: That's not too bad 👍

kurt-r2c: There was an edge case I wasn't handling, fixed and the results are better now.  Turns out there's lots of permissive path handling in CLI applications since you arguably already control the environment.

minusworld (approved):

PR #1810, opened 2022-02-11
Update suppressed-exception-handling-finally-break.yaml

dlukeomalley:
@DrewDennison @obi1kenobi Improving the help text here. Can you provide links to sources explaining this behavior for us to add?

### DIFF
--- a/python/correctness/suppressed-exception-handling-finally-break.yaml
+++ b/python/correctness/suppressed-exception-handling-finally-break.yaml
@@ -34,7 +34,7 @@ rules:
             ...
             return ...
   message: >-
-      having a `break`, `continue`, or `return` in a `finally` block will cause strange behaviors including an execption to be not caught
+      Having a `break`, `continue`, or `return` in a `finally` block will cause strange behaviors, like exceptions not being caught.
   metadata:
     category: 'correctness'
     technology: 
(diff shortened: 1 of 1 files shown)


python/correctness/suppressed-exception-handling-finally-break.yaml
```diff
@@ -34,7 +34,7 @@ rules:
             ...
             return ...
   message: >-
-      having a `break`, `continue`, or `return` in a `finally` block will cause strange behaviors including an execption to be not caught
```
dlukeomalley, line 37: Had a typo.

minusworld (approved):
Thanks!

python/correctness/suppressed-exception-handling-finally-break.yaml
```diff
@@ -34,8 +34,10 @@ rules:
...
+      Having a `break`, `continue`, or `return` in a `finally` block will cause strange behaviors, like exceptions not being caught.
   metadata:
+    references:
+    - https://www.python.org/dev/peps/pep-0601/#rejection-note
```
obi1kenobi, line 41: This PEP got rejected, so its suggestions are not implemented. I think it's still a good idea to include it, though, because it explains why this requires a lint check as opposed to being handled within the language itself.

The link I'm suggesting below is the Python language reference that defines this behavior, and which the rejected PEP 601 wanted to amend.
```suggestion
    - https://docs.python.org/3/reference/compound_stmts.html#the-try-statement
    - https://www.python.org/dev/peps/pep-0601/#rejection-note
```

minusworld: Perfect, even better! :D

PR #1812, opened 2022-02-11
AWS Lambda SQLi rule for GO

inkz:


go/aws-lambda/security/database-sqli.yaml
```diff
@@ -0,0 +1,51 @@
...
+      - pattern-inside: |
+          func $HANDLER($CTX $CTXTYPE, $EVENT $TYPE, ...) {...}
+          ...
+          lambda.Start($HANDLER, ...)
+      - patterns:
+          - pattern-inside: |
+              func $HANDLER($EVENT $TYPE) {...}
+              ...
+              lambda.Start($HANDLER, ...)
+          - pattern-not-inside: |
```
kurt-r2c, line 46: This is because the `Context` argument isn't user-controllable, correct?

Recommend adding a negative test case for this for regression purposes in the future.

kurt-r2c (changes requested):
Let's standardize adding negative test cases for `pattern-not*` cases.  Test suites should help us with regression as well as correctness.

go/aws-lambda/security/database-sqli.yaml
inkz: :ok_hand:

inkz:
@kurt-r2c You're right about `Context` and I agree with you about testing. Good test coverage is important, but I think it is not always possible to support test for `pattern-not`, `pattern-not-inside` in one test file, e.g. :
```yaml
- pattern-inside: |
    import "foo"
    ...
- pattern-not-inside: |
    import "bar"
    ...
```

go/aws-lambda/security/database-sqli.go
```diff
@@ -81,6 +82,39 @@ func handler(request events.APIGatewayProxyRequest) (events.APIGatewayProxyRespo
...
+
+	defer db.Close()
+
+  	// ok: database-sqli
+	results, err := db.Query("select e.emp_no, e.first_name, e.last_name " +
+		"from employees e, departments d, dept_emp de " +
+		"where de.emp_no = e.emp_no " +
+		"and de.dept_no = d.dept_no " +
+		"and d.dept_name = 'Marketing' " +
+		"and e.last_name LIKE '" + searchCriteria + "%';")
```
kurt-r2c, line 101: Nice 👍

kurt-r2c (approved):

PR #1816, opened 2022-02-13
Fix confusion between ruby and rails

wonda-tea-coffee:
In some documents, the In some documents, the word "ruby" was used when it should have been "rails".


0xDC0DE (approved):
Thanks for fixing these messages!

PR #1817, opened 2022-02-14
Add AWS Lambda tainted SQLi rule for Java

inkz:
shamelessly took it from: https://github.com/returntocorp/semgrep-rules/blob/develop/java/lang/security/audit/sqli/tainted-sql-from-http-request.yaml
but it works on real code! :smile:


0xDC0DE (approved):
This looks good to me!

PR #1820, opened 2022-02-16
Add autofix properties to rules from `p/ci`

inkz:


python/cryptography/security/insecure-cipher-mode-ecb.yaml
```diff
@@ -1,6 +1,9 @@
 rules:
   - id: insecure-cipher-mode-ecb
     pattern: cryptography.hazmat.primitives.ciphers.modes.ECB(...)
+    fix-regex:
+      regex: ECB
```
kurt-r2c: This will break AES functionality - ECB does not require an IV (all blocks are encrypted identically) and ECB mode will be missing correct IV initialization for CBC mode.

python/cryptography/security/insecure-hash-algorithms.yaml
```diff
@@ -1,6 +1,9 @@
 rules:
   - id: insecure-hash-algorithm-md5
     pattern: cryptography.hazmat.primitives.hashes.MD5(...)
+    fix-regex:
+      regex: MD5
```
kurt-r2c, line 5: This may have side-effects like cache-busting: MD5 is cheap/fast and unsuitable for cryptographic operations but is totally fine for things like cache membership.

python/requests/security/disabled-cert-validation.yaml
```diff
@@ -23,3 +23,6 @@ rules:
       - pattern: requests.request(..., verify=False, ...)
       - pattern: requests.get(..., verify=False, ...)
       - pattern: requests.post(..., verify=False, ...)
+    fix-regex:
+      regex: verify(\s)*=(\s)*False
```
kurt-r2c, line 27: This is correct but developers will hate this - SSL verification is often disabled in test environments because provisioning machine keys for SSL requires some domain expertise & dev environment setup scripting.

kurt-r2c (commented):
Autofixing cryptography is hazardous for data integrity - unless the implementer used an agile crypto scheme (if they are running afoul of these rules, they almost certainly did not) changes of this nature will cause any part of the app dependent on cryptography to fail.

Since these are CI rules (which are commonly used as onboarding/test rules) we should be careful about introducing autofixes with potentially severe side effects.

kurt-r2c (approved):
LGTM

PR #1821, opened 2022-02-16
WIP - metavariable-analysis entropy test rules

kurt-r2c:
This depends on https://github.com/returntocorp/semgrep/pull/4672 to function!

Initial set of some entropy-powered secret rules.


clintgibler:
Nice! What about looking for just `"..."` in every language Semgrep supports?

kurt-r2c:
Does...does that work?  Let me try the constant approach and see.  The fun part about secrets is they often show up in config files/etc without quotes.

kurt-r2c:
answer: definitely works.  Added rule

edit - tests lied to me!  Parser errors are not passes!  Reverted.

experiments/metavariable-analysis/entropy/string-literal-entropy.yaml
```diff
@@ -0,0 +1,37 @@
+rules:
+  - id: string-literal-entropy
+    patterns:
+      - pattern: "..."
+      - metavariable-analysis:
+          analyzer: entropy
+          metavariable: "..."
+    message: Found a high-entropy string.
+    languages:
+      - bash
```
kurt-r2c: @clintgibler @minusworld @lejliu @inkz @colleend this is a (slightly) altered list from 
`semgrep-core --help | grep "bash" | sort | uniq | sed "s/, /\r\n- /g" | pbcopy`

Please help sanity check for correctness in case I missed something or deleted the wrong version of a language (there are shortname dupes for many of these)

PR #1824, opened 2022-02-17
Note that strcpy_s is an optional part of C11

langston-barrett:
See https://stackoverflow.com/questions/57915149/using-c11-standard-with-clang-for-use-of-strcpy-s


minusworld (approved):

ievans:
Thanks for the contribution! This could be a good candidate for a `fix:` key, too

PR #1825, opened 2022-02-18
Update use-defused-xml.yaml

dlukeomalley:
Add OWASP reference and lead with the official recommendation of the Python documentation. Also, prefer back tic over ' for GitHub comment formatting.


chmccreery (approved):

c/lang/security/insecure-use-string-copy-fn.yaml
```diff
@@ -1,19 +1,17 @@
 rules:
   - id: insecure-use-string-copy-fn
-    pattern-either:
-      - pattern: strcpy(...)
-      - pattern: strncpy(...)
+    patterns:
+      - pattern: $X(...)
+      - metavariable-regex:
```
dlukeomalley: I suspect this is slow compared to the prior implementation.

inkz: yep, gonna be slower, the rest is lgtm! :+1:

dlukeomalley: @inkz If we take this approach, is this a simple autofix for us to add?

dlukeomalley: @aryx Got any ideas for how I could approach this for both autofix + using metavariables in the rule message + being performant?

inkz: @dlukeomalley it is easy to change func name https://semgrep.dev/s/XZ1A
but I am not 100% sure that arguments will be correct.
Also there is nice comment/pr from one of our users on this rule: https://github.com/returntocorp/semgrep-rules/pull/1824

kurt-r2c (approved):
messaging improvements 🙌

c/lang/security/insecure-use-string-copy-fn.yaml
dlukeomalley: I removed this change for now.

inkz (approved):

PR #1827, opened 2022-02-19
fix(py-check): update log credential pattern

Hetang23:
Updating the credential exposure check in logs for all the levels of Python logging with all the possible object of logging object a developer would use in production.


minusworld (approved):
Cool, thanks!

PR #1831, opened 2022-02-23
Add tainted-html-response rule for AWS Lambda

inkz:
if response is HTML content and it is tainted by `event` object it is likely an XSS

### DIFF
--- a/javascript/aws-lambda/security/tainted-html-response.yaml
+++ b/javascript/aws-lambda/security/tainted-html-response.yaml
@@ -1,0 +1,34 @@
+rules:
+- id: tainted-html-response
+  severity: WARNING
+  message: >-
+    Detected user input flowing into a manually constructed HTML string. You may be
+    accidentally bypassing secure methods
+    of rendering HTML by manually constructing HTML and this could create a cross-site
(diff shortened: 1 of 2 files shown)


javascript/aws-lambda/security/tainted-html-response.yaml
```diff
@@ -0,0 +1,42 @@
...
+      - pattern-inside: |
+          function $FUNC ($EVENT, ...) {...}
+          ...
+          exports.handler = $FUNC
+      - pattern-inside: |
+          $FUNC = function ($EVENT, ...) {...}
+          ...
+          exports.handler = $FUNC
+    - pattern: $EVENT
+  pattern-sinks:
```
kurt-r2c, line 38: We might want to add sanitizers for known Javascript HTML sanitizer libraries like DOMPurify

kurt-r2c (approved):

javascript/aws-lambda/security/tainted-html-response.yaml
inkz: agree, I'll test this rule on real targets and try to collect list of popular sanitizers

PR #1833, opened 2022-02-23
rule: MongoDB NoSQLi from HTTP request (tainted)

aabashkin:
**Note:**
One test (line 50) still failing. Not sure if this is an issue with the tainting engine or if I've configured the rule incorrectly, would appreciate some review.

### DIFF
--- a/java/lang/security/audit/sqli/tainted-mongodb-nosqli-from-http-request.java
+++ b/java/lang/security/audit/sqli/tainted-mongodb-nosqli-from-http-request.java
@@ -1,0 +1,77 @@
+import javax.servlet.http.HttpServletRequest;
+import javax.servlet.http.HttpServletResponse;
+import javax.servlet.RequestDispatcher;
+import javax.servlet.ServletException;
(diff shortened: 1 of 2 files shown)


kurt-r2c:
@aabashkin your test is failing on line 50 because your method signature doesn't contain either of your taint sources:

```java
    private DBCursor findInsecure(String taintedInput) throws UnknownHostException {
        MongoClient mongoClient = new MongoClient();
        DB testDB = mongoClient.getDB("testDB");
        testCollection = testDB.getCollection("testCollection");

        String taintedQueryString = "{ 'input' : '" + taintedInput + "'}";
        // ruleid:tainted-nosqli-from-http-request
        DBObject taintedQuery = (DBObject) JSON.parse(taintedQueryString);

        DBCursor result = testCollection.find(taintedQuery);

        return result;
    }
 ```
 
 Taint mode is currently only intra-procedural and doesn't support your second test case yet: https://semgrep.dev/docs/writing-rules/data-flow/taint-mode/#minimizing-false-positives-via-sanitizers

PR #1838, opened 2022-02-24
WIP - language definition consistency rules

kurt-r2c:
Adding some rules to autofix Semgrep rules with consistent language definitions in the `language` rule definition key.

Some open questions:

- should we standardize on `terraform` or `hcl`? Terraform gurus may prefer `hcl`, people with less domain expertise will probably find `terraform` more accessible
- `c++` vs `cpp`? `c++` is "more correct", `cpp` is more HTTP-friendly

FYI @pieterdc1


ievans:
Do you want to make the "canonical" name defined in https://github.com/returntocorp/semgrep-langs and use that to compile these rules?

kurt-r2c:
per @ievans suggestion - using `semgrep-langs` as "source of truth" for lang configs. Once this is approved + merged, will open second PR with autofixes applied.

yaml/semgrep/consistency/lang-consistency-typescript.yaml
```diff
@@ -0,0 +1,20 @@
...
+              languages: ...
+      - metavariable-regex:
+          metavariable: $X
+          regex: ^typescript$
+    message: >-
+      Found '$X' in language config which diverges from semgrep.dev normalization.
+    languages:
+      - yaml
+    severity: WARNING
+    fix: ts
```
minusworld: Most of the other languages are spelled out - I don't care which we choose, but is it more consistent to spell this and JavaScript out as well?

kurt-r2c: I agree with this, however semgrep.dev does not: 
https://github.com/returntocorp/semgrep-langs/blob/2cb383112a7780e5dfe4d768161c3d4ddb0bac57/lang.json#L229-L231

minusworld: Hm, gotcha 👍

minusworld (approved):

kurt-r2c:
@minusworld @ievans now with more autogeneration from `semgrep-rules`

yaml/semgrep/consistency/lang-consistency-cpp.yaml
```diff
@@ -0,0 +1,17 @@
...
+  metadata:
+    category: correctness
+    technology:
+    - semgrep
+  patterns:
+  - pattern: $X
+  - pattern-inside: 'languages: ...'
+  - metavariable-regex:
+      metavariable: $X
+      regex: ^(c++)$
```
ievans: does this need to be escaped?

yaml/semgrep/consistency/generate_rules.py
```diff
@@ -0,0 +1,56 @@
...
+                        },
+                        {"pattern-inside":"languages: ..."},
+                        {
+                            "metavariable-regex": {
+                                "metavariable":"$X",
+                                "regex":f"^({key_string})$"
+                            }
+                        }
+                    ],
+                    "message": "Found '$X' in language config which diverges from semgrep.dev normalization.",
```
ievans: Maybe say: `"$X is a valid langauge key but for standardization please use " + lang_data['id']"` just a bit clearer to me even though we already have `fix`

yaml/semgrep/consistency/lang-consistency-cpp.yaml
kurt-r2c: yep. will fix in generator script

yaml/semgrep/consistency/generate_rules.py
kurt-r2c: fair point will fix

kurt-r2c:
@ievans pretty sure I got all your comments, please sanity check to make sure

ievans (approved):
Lgtm! Now let's get this working via semgrep-app on semgrep-rules repository, so we get those sweet new autofix comments!

PR #1862, opened 2022-03-04
adding metavariable-entropy to generic secrets rules

colleend:
Adding these to some of the rules with a LOT of findings and large number of negative developer sentiment.

Refer to: https://docs.google.com/document/d/1Q8qCi6imMv_z__TUeRomEgRLsthOmsEw7uihSSrvyTI/edit?usp=sharing for the link to the analysis

To test:
Run `semgrep --test .` in the `generic/secrets/security` directory and check that all tests pass.

### DIFF
--- a/generic/secrets/security/detected-aws-account-id.yaml
+++ b/generic/secrets/security/detected-aws-account-id.yaml
@@ -8,6 +8,18 @@ rules:
           $ACCOUNT_ID : $SECRET
       - pattern: |
           $ACCOUNT_ID => $SECRET
+      - pattern: |
+          $ACCOUNT_ID = "$SECRET"
+      - pattern: |
+          $ACCOUNT_ID : "$SECRET"
+      - pattern: |
+          $ACCOUNT_ID => "$SECRET"
+      - pattern: |
+          "$ACCOUNT_ID" = "$SECRET"
+      - pattern: |
+          "$ACCOUNT_ID" : "$SECRET"
+      - pattern: |
+          "$ACCOUNT_ID" => "$SECRET"
     - metavariable-analysis:
         metavariable: $SECRET
         analyzer: entropy
--- a/generic/secrets/security/detected-aws-account-id.yaml
+++ b/generic/secrets/security/detected-aws-account-id.yaml
@@ -1,8 +1,23 @@
 rules:
   - id: detected-aws-account-id
-    pattern-regex: |-
-      ("|')?(AWS|aws|Aws)?_?(ACCOUNT|account|Account)_?(ID|id|Id)?("|')?\s*(:|=>|=)\s*("|')?[0-9]{12}("|')?
-    languages: [regex]
+    patterns:
+    - pattern-either:
+      - pattern: |
+          $ACCOUNT_ID = $SECRET
+      - pattern: |
+          $ACCOUNT_ID : $SECRET
+      - pattern: |
+          $ACCOUNT_ID => $SECRET
+    - metavariable-analysis:
+        metavariable: $SECRET
+        analyzer: entropy
+    - metavariable-regex:
+        metavariable: $SECRET
+        regex: ^[0-9]{12}$
+    - metavariable-regex:
+        metavariable: $ACCOUNT_ID
+        regex: (AWS|aws|Aws)?_?(ACCOUNT|account|Account)_?(ID|id|Id)?("|')?
+    languages: [generic]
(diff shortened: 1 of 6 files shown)


generic/secrets/security/detected-private-key.yaml
```diff
@@ -1,7 +1,22 @@
...
+        - pattern:
+            -----BEGIN $TYPE PRIVATE KEY-----
+            $KEY
+        - metavariable-regex:
+            metavariable: $TYPE
+            regex: (?i)([dr]sa|ec|openssh|encrypted)?
+      - patterns:
+        - pattern: |
+            -----BEGIN PRIVATE KEY-----
+            $KEY
```
minusworld, line 15: I think $KEY will only read up to a slash, since in `generic` mode metavars only capture words.

A few questions:
1. Could we check this at scale? It might not matter too much that $KEY doesn't capture everything.
2. I -think- there's a feature with `pattern-regex` that will map capture groups to metavariables like `$1`, `$2`, etc. If we need to keep the regex but also use entropy detection, we could use that if it works.

generic/secrets/security/detected-github-token.yaml
```diff
@@ -1,8 +1,18 @@
...
   - id: detected-github-token
-    pattern-regex: |-
-      gh[pousr]_[A-Za-z0-9_]{36,251}
-    languages: [regex]
+    patterns:
+    - pattern-either:
+      - pattern: |
+          $VAR = $SECRET
+      - pattern: |
+          $VAR: $SECRET
```
minusworld, line 8: The secret could also be in a URL somewhere.
If we try the old and new versions of this rule at scale, what do we find?

generic/secrets/security/detected-aws-account-id.yaml
```diff
@@ -1,8 +1,23 @@
 rules:
   - id: detected-aws-account-id
-    pattern-regex: |-
-      ("|')?(AWS|aws|Aws)?_?(ACCOUNT|account|Account)_?(ID|id|Id)?("|')?\s*(:|=>|=)\s*("|')?[0-9]{12}("|')?
-    languages: [regex]
+    patterns:
+    - pattern-either:
+      - pattern: |
+          $ACCOUNT_ID = $SECRET
```
minusworld, line 6: Same as the other rule - this could be in a URL somewhere. Maybe we don't care about it sometimes, but I'm curious how the old and this new rule look when tried at scale.

minusworld (changes requested):
I like the new direction with `generic`, but I wonder about the trade-offs between straight regexes and the new `generic` versions. Could we check out the old and new versions at scale to see how they compare?

colleend:
Good idea --I'll start running these at scale next week to see what results we get first 👍

colleend:
ok cool! finally finished running the rules on a few github repos and triaging :D the results are here: https://www.notion.so/Dev-Sentiment-Rule-Editing-8cfa86531f2c416799a14945ce5e076b. 

A summary of the results is:
1) the results for the `private-key` rule are the same, so this rule should be fine.
2) there are a few TP results in the previous `github-token` rule and not many FP. I'll change the rule to include strings and URL scanning to also capture these TP.
3) there are a few TP results in the previous `aws-account-id` rule (but ALOT more FP). I'll also change the rule to include strings (there were no URL results caught by either rule).

generic/secrets/security/detected-private-key.yaml
colleend: this rule captured the same results as the previous `private-key` rule, so I think we should be ok with this one. I did try using capture groups, but they don't work with metavariable-entropy currently. (and metavariable-ellipsis also don't work)

generic/secrets/security/detected-github-token.yaml
colleend: I'll add in variations for string and url-finding for this one!

generic/secrets/security/detected-aws-account-id.yaml
colleend: from the 400 repos we tested on, there were no URL results. I'll add in string-checking, though 👍

colleend:
I ran the scans again with the edited rules -- getting alot of internal server errors with the github-token one (and therefore no results -- perhaps alot of the repos were taken down somehow?), and much better results with the aws-account-id edited rule.

minusworld:
Awesome. Let's go ahead and do the merge, then, and see how it goes. Thanks!

minusworld (approved):

PR #1870, opened 2022-03-08
New Rule for Terraform/GCP

theJuan1112:
I saw that there were no terraform rules for GCP, so wrote a rule to check that a proper TLS version is being used on the load balancers.


theJuan1112:
I am not sure what is the error that the action is showing.

minusworld:
@theJuan1112 Was this published from the new editor? If so, I suspect the problem is that it's missing the top-level `rules:` key in the YAML file. I'll have to let the editor team know that it needs to add this key when publishing from the editor.

theJuan1112:
@minusworld It was not from the editor. If it's best I could just close this pull request and try again from the editor, if that is better then.

minusworld:
@theJuan1112 No need, I was just wondering if we needed to fix something on our end. I added the missing `rules:` key and it looks all good now!

minusworld (approved):

PR #1871, opened 2022-03-09
Switch to dockerfile parser for multiple-cmd-instructions rule

minusworld:


minusworld:
Note that the `--test` flag doesn't like files with the same name as the rule ID that -don't- have `ruleid` annotations in them. I've included an OK test file, but it doesn't run with `--test` yet. We'll have to change how the `--test` flag works with these kind of files.

inkz (approved):
lgtm

kurt-r2c (approved):
lgtm

PR #1878, opened 2022-03-10
add twirl-html-var rule

inkz:
Raw html in Twirl templates: https://www.playframework.com/documentation/2.8.x/ScalaTemplates#Escaping


kurt-r2c (approved):
lgtm

PR #1883, opened 2022-03-10
Filter request.user.id from django access-foreign-keys

### LINKED ISSUE #1644: `python.django.performance.access-foreign-keys.access-foreign-keys` doesn't exclude `request.user.id`
**Describe the bug**
In Django, the views get a request instance that contains a User object. Since it's common to access `request.user.id`, and the request object is NOT a model instance, this rule triggers a false positive that could perhaps be ignored with `pattern-not: request.user.id`

**To Reproduce**
```python
from django.http import HttpResponse

def cool_view(request):
    return HttpResponse({'user_id': request.user.id})
```

**Expected behavior**
This example should not trigger this rule, since "request" is not a model instance

**Priority**
How important is this to you?

**Additional Context**
N/A

minusworld:
Fixes https://github.com/returntocorp/semgrep-rules/issues/1644

### DIFF
--- a/python/django/performance/access-foreign-keys.yaml
+++ b/python/django/performance/access-foreign-keys.yaml
@@ -1,6 +1,8 @@
 rules:
   - id: access-foreign-keys
-    pattern: $X.user.id
+    patterns:
+    - pattern: $X.user.id
+    - pattern-not: request.user.id
     message: You should use ITEM.user_id rather than ITEM.user.id to prevent running an extra query.
     languages: [python]
     severity: WARNING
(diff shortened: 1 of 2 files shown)


python/django/performance/access-foreign-keys.yaml
```diff
@@ -1,6 +1,8 @@
 rules:
   - id: access-foreign-keys
-    pattern: $X.user.id
+    patterns:
```
kurt-r2c, line 3: maybe add a pattern-inside with a Django import to be relatively sure this is Django related?

```yaml
- pattern-inside: |
    from django.$MOD import $CLASS
    ...
```

kurt-r2c (approved):
Change seems strictly better than old way, 1 suggestion to possibly improve accuracy

PR #1886, opened 2022-03-11
Fix test expectations after returntocorp/semgrep#4792

IagoAbal:
After PR https://github.com/returntocorp/semgrep/pull/4792 fixes PA-947, there is no more false
negative here (note that this should have been a "todoruleid:").

### DIFF
--- a/scala/play/security/tainted-html-response.scala
+++ b/scala/play/security/tainted-html-response.scala
@@ -42,6 +42,16 @@ class XssController extends Controller {
     // ruleid: tainted-html-response
     Ok(views.html.xssString.render(value))
   }
+  
+  def vulnerable8(value:String) = Action {
+    // ruleid: tainted-html-response
+    Ok(views.html.xssHtml.render(Html.apply("Hello "+value+" !")))
+  }
+
+  def vulnerable9(value:String) = Action {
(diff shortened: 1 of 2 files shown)


minusworld:
@inkz Is the changed line still something that shouldn't be detected, or should it be detected now?

inkz:
@minusworld this line should not be detected

inkz:
but awesome that it works now! :partying_face:

IagoAbal:
@inkz 

Given

```yaml
  pattern-sources:
  - patterns:
    - pattern-either:
      - patterns:
        - pattern: $PARAM
        - pattern-inside: |
            def $CTRL(..., $PARAM: $TYPE, ...) = Action {
              ...
            }
```
and

```yaml
  pattern-sinks:
  - pattern-either:
    - pattern: views.html.$TEMPLATE.render(...)
```

how do you justify that this should not match? 

```scala
  def safeTemplate(value:String) = Action {
    Ok(views.html.template.render(value))
  }
```

inkz:
@IagoAbal ah thats my bad! I messed it up, let me quickly fix the rule, I pushed the wrong version :pray:

inkz (approved):

PR #1887, opened 2022-03-11
Make sure $ATTR is filled in with react-http-leak rule

### LINKED ISSUE #1715: react-http-leak doesn't fill $ATTR in most cases
Semgrep reports:

> This HTML element 'img' and attribute '$ATTR' together may load an external resource.

I expect:

> This HTML element 'img' and attribute 'src' together may load an external resource.

[semgrep-rules/react-http-leak.yaml at develop · returntocorp/semgrep-rules](https://github.com/returntocorp/semgrep-rules/blob/develop/typescript/react/security/audit/react-http-leak.yaml)

The message in this rule is:

> This HTML element '$EL' and attribute '$ATTR' together may load an external resource.

But the `$ATTR` variable is only used for attributes of `<object>`, so it's not set for `` and many other tags.

minusworld:
Fixes https://github.com/returntocorp/semgrep-rules/issues/1715

### DIFF
--- a/typescript/react/security/audit/react-http-leak.yaml
+++ b/typescript/react/security/audit/react-http-leak.yaml
@@ -3,34 +3,46 @@ rules:
     pattern-either:
       - patterns:
           - pattern: |
-              <$EL codebase=... />
+              <$EL $ATTR=... />
+          - pattern-not: |
+              <$EL codebase="..." />
           - metavariable-regex:
               metavariable: $EL
               regex: applet|object
-          - pattern-not: |
-              <$EL codebase="..." />
+          - metavariable-regex:
+              metavariable: $ATTR
+              regex: codebase|archive
       - patterns:
           - pattern: |
-              <$EL cite=... />
+              <$EL $ATTR=... />
           - metavariable-regex:
               metavariable: $EL
               regex: blockquote|del|ins|q
+          - metavariable-regex:
+              metavariable: $ATTR
+              regex: cite
           - pattern-not: |
               <$EL cite="..." />
       - patterns:
           - pattern: |
-              <$EL longdesc=... />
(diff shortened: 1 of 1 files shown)


colleend:
@emjin I think the CI `--validate` issues are from a rule that you wrote, can you take a look? 🙏 

UPDATE:

Emma found the issue! Lines 162 and 142 don't have any metavariables that the metavariable regex/patterns are flagging

typescript/react/security/audit/react-http-leak.yaml
```diff
@@ -3,112 +3,170 @@ rules:
...
+              metavariable: $EL
+              regex: video
+          - metavariable-regex:
+              metavariable: $ATTR
+              regex: poster
           - pattern-not: |
               <video poster="..." />
       - patterns:
           - pattern: |
               <command icon=... />
```
colleend: we haven't changed this to $EL and $ATTR!

typescript/react/security/audit/react-http-leak.yaml
```diff
@@ -3,112 +3,170 @@ rules:
...
+              metavariable: $EL
+              regex: html
+          - metavariable-regex:
+              metavariable: $ATTR
+              regex: manifest
           - pattern-not: |
               <html manifest="..." />
       - patterns:
           - pattern: |
               <meta content=... />
```
colleend: haven't changed this either

colleend (changes requested):

kurt-r2c:
@colleend can you recheck this PR for the issue(s) you raised? The force push has messed up the history and the point-in-time commit appears to be missing.

colleend (approved):
cool, just looked over it again and LGTM!

PR #1890, opened 2022-03-14
update react-props-injection rule

inkz:
the rule is to permissive after deep expression improvement from: https://github.com/returntocorp/semgrep/pull/4796


aryx:
This is related to https://github.com/returntocorp/semgrep/issues/4788
cc @lejliu

aryx (approved):

PR #1903, opened 2022-03-17
fix tainted scala rules

inkz:
Small fixes to taint rules for Scala


0xDC0DE (approved):
Looks good. Approved.

PR #1904, opened 2022-03-17
Add ScalaJS eval rule

inkz:
Dangerous eval
* note that it is a **JS** eval https://www.scala-js.org/doc/


0xDC0DE (approved):
That looks good to me!

PR #1906, opened 2022-03-18
Tainted SQL String rules for Scala

inkz:
Scala versions of: 

https://github.com/returntocorp/semgrep-rules/blob/d84ca1d208965041464983d3c2ea1e3329cad1d9/java/spring/security/injection/tainted-sql-string.yaml

https://github.com/returntocorp/semgrep-rules/blob/d84ca1d208965041464983d3c2ea1e3329cad1d9/java/lang/security/audit/sqli/tainted-sql-from-http-request.yaml

### DIFF
--- a/scala/lang/security/audit/tainted-sql-string.scala
+++ b/scala/lang/security/audit/tainted-sql-string.scala
@@ -40,7 +40,7 @@ object Smth {
   }
 
   def call5(name: String) = {
-    // todoruleid: tainted-sql-string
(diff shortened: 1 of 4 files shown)


0xDC0DE:
Looks good to me.
I did notice that the `tainted-sql-string` didn't pass and I had to update to the newest version of semgrep.
This rule does not work on version 0.84, but works on 0.85.

0xDC0DE (approved):

Issue #1881, opened 2022-03-10
command-injection-process-builder (Java) false negatives


lfama:
**Describe the bug**
The rule doesn't match due to missing matching patterns for Windows OS (i.e. "cmd"), leading to false negatives.

**To Reproduce**
Run the rule against the following Java code:
```java
public class Test {
StringBuilder getResponseFromPingCommand(String ipAddress, boolean isValid) throws IOException {
        boolean isWindows = System.getProperty("os.name").toLowerCase().startsWith("windows");
        StringBuilder stringBuilder = new StringBuilder();
        if (isValid) {
            Process process;
            if (!isWindows) {
                process =
                        // Rule matches here
                        new ProcessBuilder(new String[] {"sh", "-c", "ping -c 2 " + ipAddress})
                                .redirectErrorStream(true)
                                .start();
            } else {
                process =
                        // But doesn't match here
                        new ProcessBuilder(new String[] {"cmd", "/c", "ping -n 2 " + ipAddress})
                                .redirectErrorStream(true)
                                .start();
            }
            try (BufferedReader bufferedReader =
                    new BufferedReader(new InputStreamReader(process.getInputStream()))) {
                bufferedReader.lines().forEach(val -> stringBuilder.append(val).append("\n"));
            }
        }
        return stringBuilder;
    }
}
```

**Expected behavior**
The rule should also match Windows "cmd" command.

**Priority**
How important is this to you?
- [ ] P0: blocking me from making progress
- [ ] P1: this will block me in the near future
- [X] P2: annoying but not blocking me

0xDC0DE:
Rule has been updated!

PR #1908, opened 2022-03-21
Add cases for windows OS to command-injection-process-builder

0xDC0DE:
Fix for bug reported here: https://github.com/returntocorp/semgrep-rules/issues/1881


inkz (approved):
lgtm!

Issue #1802, opened 2022-02-10
Incorrect description in missing-noreferrer


Sjord:
The description in [semgrep-rules/missing-noreferrer.yaml at develop · returntocorp/semgrep-rules](https://github.com/returntocorp/semgrep-rules/blob/develop/html/security/missing-noreferrer.yaml) is incorrect:

      This anchor tag with 'target="_blank"' is missing 'noreferrer'.
      A page opened with 'target="_blank"' can access the window object of the origin page.
      This means it can manipulate the 'window.opener' property, which could redirect the origin page to a malicious URL.
      This is called reverse tabnabbing. To prevent this, include 'rel=noreferrer' on this tag.

This is just the description of [missing-noopener](https://github.com/returntocorp/semgrep-rules/blob/develop/html/security/missing-noopener.yaml), but with `opener` replaced by `referer`. This is incorrect. `noopener` indeed protects against exposing the `window.opener`, but `noreferer` does not. It prevents exposing the URL of the current page to the newly opened page.

Sjord:
[This page](https://mathiasbynens.github.io/rel-noopener/) and [CWE-1022](https://cwe.mitre.org/data/definitions/1022.html) indicate that `noreferrer` also disables the `window.opener` value, and in Chrome and Opera that indeed seems to be the case. This is probably why two rules are created, one that checks for noopener and one that checks for noreferrer. But that's incorrect; either attribute solves this issue, but according to semgrep you need both attributes.

kurt-r2c:
Resolved by #1910. Old rules will still be present in the [semgrep.dev](https://semgrep.dev) registry pending support for rule deletion - users should adopt `missing-noopener-or-noreferrer` if they want the "latest and greatest" version of this check.

PR #1910, opened 2022-03-21
Merge two rules into one 

0xDC0DE:
This rule checks whether either of the two attributes is present.

Fix for: https://github.com/returntocorp/semgrep-rules/issues/1802


kurt-r2c (approved):
LGTM

inkz:
I am not sure that changing `ruleid`s is a good thing, we may have users who depend on this rules and changing ids may break their workflow

cc @minusworld

minusworld:
What will happen is that the two old rules will remain, and the new rule will also come into existence. This isn't too much of a problem, we just have to remember that we need to delete the old rules once deletion becomes available. You can merge this.

PR #1914, opened 2022-03-22
make rule better according to feedback

colleend:
There's pretty severe negative feedback for the 'formatted-sql-query' rule, so this PR makes the rule message better and changes the rule mode to taint to (hopefully) result in less FP

_To Test:_
run `semgrep -f formatted-sql-query.yaml formatted-sql-query.py`

### DIFF
--- a/python/lang/security/audit/formatted-sql-query.yaml
+++ b/python/lang/security/audit/formatted-sql-query.yaml
@@ -1,7 +1,11 @@
 rules:
   - id: formatted-sql-query
     message: >-
-      Detected possible formatted SQL query. Use parameterized queries instead.
+      Detected possible formatted SQL query. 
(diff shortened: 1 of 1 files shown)


inkz (approved):
:+1:

minusworld:
The sources here look like they are designed specifically for Flask. Will this significantly alter the behavior of the rule?

Additionally, I think that the negative feedback given was that it fired on an untouched piece of code (an issue with the app logic), not necessarily that the finding itself.

colleend:
I see, if it's primarily an issue with the app's targeting of files, we don't need to translate the rule to taint then. I think the `request.$X` will also cover some requests for django, but it's hard to get full source coverage of all python web frameworks. I'll close this PR and we can edit this later if we get more feedback about it.

PR #1915, opened 2022-03-22
Update Scala Command Injection rules

inkz:
Mitigate FPs by adding check for `import sys.process`


kurt-r2c (approved):
Do tests need updating?

inkz:
> Do tests need updating?

Good call, added tests! :+1:

PR #1918, opened 2022-03-23
Add rule for structured logging in C#

Sjord:
String interpolation should be done by the logging library instead of by
$"...", so that querying variables is possible, and log injection is prevented.

Add rules for various logging libraries. Some methods are double (e.g.
$LOG.Debug), since they appear in multiple libraries. <s>I have kept this
intentionally to make it clear which methods belong in which libraries.</s>

Also check that the method is called on something that is named "log" or
something like that. Logger objects could be named differently, but this
reduces the chance of false positives when calling `Trace` or something on an
object that is not a logger.

### DIFF
--- a/csharp/lang/best-practice/structured-logging.cs
+++ b/csharp/lang/best-practice/structured-logging.cs
@@ -1,0 +1,83 @@
+using Microsoft.Extensions.Logging;
+using Serilog;
+using NLog;
+
+class Program
+{
+    public static void SerilogSample()
+    {
+        using var serilog = new LoggerConfiguration().WriteTo.Console().CreateLogger();
+
+        var position = new { Latitude = 25, Longitude = 134 };
+        var elapsedMs = 34;
+
+        // ok: structured-logging
+        serilog.Information("Processed {@Position} in {Elapsed:000} ms.", position, elapsedMs);
+
+        // ruleid: structured-logging
+        serilog.Information($"Processed {position} in {elapsedMs:000} ms.");
+    }
+
+    public static void MicrosoftSample()
+    {
+        var loggerFactory = LoggerFactory.Create(builder => {
+                builder.AddConsole();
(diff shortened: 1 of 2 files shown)


0xDC0DE:
Hi @Sjord! This is great! Thank you for your contribution!

Instead of checking the name, you could use typed metavariables: https://semgrep.dev/docs/writing-rules/pattern-syntax/#typed-metavariables

That looks like this:
```
pattern: (System.Logger $LOG).Debug($"...")
```

Do you mind updating your rule to include the correct type instead of using `metavariable-regex` for the variable name?

Sjord:
Thanks, @pieterdc1. I wasn't aware of that feature. It looks great, but I have some trouble getting it to work in practice in C#.

Semgrep doesn't always infer the full namespace of the type. Whether a variable is of type ILogger or Logging.ILogger or Microsoft.Extensions.Logging.ILogger or nothing depends a lot on how it's defined. Loggers can be of different types, e.g. `Logger` or `ILogger`. In Serilog you can log using a static class instead of an instance. Type inference doesn't work well with [my test code](https://github.com/returntocorp/semgrep-rules/pull/1918/files#diff-b0c5d67a7eaf9a0a933d1f804e5cbe5badb34f5c02e06ec1e9fbdfe4f9599ab0), because I define everything as `var` without an explicit type. But that may be atypical usage of loggers.

I think the metavariable-regex is a bit dirty, but works better in practice. I'll give it some more thought, but it's not clear to me that specifying types would result in a better rule.

kurt-r2c:
I believe .NET type awareness is limited to either explicit includes or .NET standard lib - for e.g. `NLog` Semgrep has no way to infer type data from the factory.  As-is, I think this is the best approach while C# type awareness support improves.

kurt-r2c (approved):

PR #1919, opened 2022-03-23
Add path keys for generic rules that take a long time

minusworld:
Based on results from https://github.com/monkey-test/sentry/runs/5654646913?check_suite_focus=true


kurt-r2c (commented):
LGTM - these changes shouldn't be harmful

kurt-r2c (approved):

PR #1921, opened 2022-03-23
get rid of dupes in Ruby rule

colleend:
This rule led to duplicate findings, remove two lines to get rid of dupes.


kurt-r2c (approved):
🚢

PR #1922, opened 2022-03-23
Add GHA workflow for testing with agent

minusworld:


.github/workflows/semgrep-rules-test-develop.yml
```diff
@@ -9,7 +9,7 @@ on:
 jobs:
   # Note: if you change this test there will likely need to be a
   # corresponding change in returntocorp/semgrep
-  test-latest:
+  test-develop:
```
emjin, line 12: Why did this get changed? Not blocking, just wondering

emjin (approved):
Thanks!

emjin:
Has this been/will this be tested? It should be sufficient to try to release, since semgrep-agent is still behind.

.github/workflows/semgrep-rules-test-develop.yml
minusworld: Oh, so that it had a different name in the GitHub UI. I forgot that I made this change here.

minusworld:
> Has this been/will this be tested? It should be sufficient to try to release, since semgrep-agent is still behind.

Yes, I tested it on this [commit here](https://github.com/returntocorp/semgrep-rules/pull/1922/commits/9f19d5c0027bcca811d95e79c54bdbe4f5811f81). The failure is expected because the `metavariable-analysis` rules are still in `develop`.

image

PR #1924, opened 2022-03-24
change 2 rules to taint

colleend:
* Makes `avoid-insecure-deserialization.yaml` have less FP by making it taint and adding filtering for specific functions
* Improve `Python.flask.security.audit.directly-returned-format-string.directly-returned-format-string`  by making it taint/adding pattern-nots

_To test:_
Run `semgrep --test .` on the directory the files are in. 

Edit: @minusworld should I test this on real world examples? I am getting rate limited again 😭

### DIFF
--- a/python/django/security/audit/avoid-insecure-deserialization.yaml
+++ b/python/django/security/audit/avoid-insecure-deserialization.yaml
@@ -14,38 +14,43 @@ rules:
     languages:
       - python
     severity: ERROR
-    patterns:
-      - pattern-inside: |
-          def $X(..., request, ...):
-            ...
-      - pattern-either:
-          - pattern: |
-              $VAR = <... request.$Y.get(...) ...>
+    mode: taint
+    pattern-sources:
+    - pattern-either:
+      - patterns:
+        - pattern-inside: |
+            def $INSIDE(..., $PARAM, ...):
               ...
-              pickle.$FUNC(<... $VAR ...>)
-          - pattern: pickle.$FUNC(<... request.$Y.get(...) ...>)
+        - pattern: |
+            $PARAM
+      - pattern: 
+          request.$REQFUNC(...)
+      - pattern:
+          request.$REQFUNC.get(...)
+      - pattern: 
+          request.$REQFUNC[...]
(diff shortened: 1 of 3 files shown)


python/django/security/audit/avoid-insecure-deserialization.yaml
```diff
@@ -14,38 +14,43 @@ rules:
...
+    mode: taint
+    pattern-sources:
+    - pattern-either:
+      - patterns:
+        - pattern-inside: |
+            def $INSIDE(..., $PARAM, ...):
               ...
-              pickle.$FUNC(<... $VAR ...>)
-          - pattern: pickle.$FUNC(<... request.$Y.get(...) ...>)
+        - pattern: |
```
minusworld, line 24: For Django, user data almost always comes from the `request` object, so we can limit scope to functions with `request` as a parameter rather than any arbitrary paramter

minusworld:
I think normal -> taint rules probably doesn't need to be tested. We can reasonably infer the behavior, imo.

minusworld (changes requested):
Left one comment about Django user input

python/django/security/audit/avoid-insecure-deserialization.yaml
colleend: do the `request` objects usually get passed as a parameters to functions? I was thinking I could just get rid of that `patterns` clause as long as we have the 
```
- pattern-either:
      - pattern:
           request.$REQFUNC(...)
       - pattern:
           request.$REQFUNC.get(...)
       - pattern:
           request.$REQFUNC[...]
```
in sources

minusworld: I'm not 100% certain of the precise answer, but I've always seen in Django a parameter called `request` used to indicate the incoming request. It's probably not a big deal, but it using a `pattern-inside` clause to scope to a parameter named `request` might help with stray, random variables that also happened to be named `request`. 🤷‍♂️  I'll  let you make the final decision so I'll go ahead and approve :)

minusworld (approved):
thanks!

PR #1925, opened 2022-03-24
editing dockerfile rule after FP on our repos

colleend:
* changed this rule from generic -> dockerfile
* also changed this so that it doesn't fire on the example case (the one in our repos)

_To test:_
* `semgrep --test .` on the directory


minusworld (approved):
niiiice

Issue #1793, opened 2022-02-09
Convert `generic.dockerfile` rules to use Dockerfile parser


minusworld:
**Is your feature request related to a problem? Please describe.**
Dockerfile support was released in Semgrep recently. Current Dockerfile rules use `generic` mode. This leads to issues like https://github.com/returntocorp/semgrep/issues/4642 and others.

**Describe the solution you'd like**
Switch Dockerfile rules to use the Dockerfile parser.

In `generic/dockerfile/`, the following rules need to be updated:

- [x] avoid-apk-upgrade.yaml
- [x] avoid-apt-get-upgrade.yaml
- [x] avoid-dnf-update.yaml
- [x] avoid-latest-version.yaml
- [x] avoid-platform-with-from.yaml
- [x] avoid-yum-update.yaml
- [x] avoid-zypper-update.yaml
- [x] maintainer-is-deprecated.yaml
- [x] missing-apk-no-cache.yaml
- [x] missing-dnf-assume-yes-switch.yaml
- [x] missing-dnf-clean-all.yaml
- [x] missing-image-version.yaml
- [x] missing-no-install-recommends.yaml
- [x] missing-pip-no-cache-dir.yaml
- [x] missing-yum-assume-yes-switch.yaml
- [x] missing-yum-clean-all.yaml
- [x] missing-zypper-clean.yaml
- [x] missing-zypper-no-confirm-switch.yaml
- [x] nonsensical-command.yaml
- [x] prefer-apt-get.yaml
- [x] prefer-copy-over-add.yaml
- [x] prefer-json-notation.yaml
- [x] remove-package-cache.yaml
- [x] remove-package-lists.yaml
- [x] set-pipefail.yaml
- [ ] use-absolute-workdir.yaml
- [x] use-either-wget-or-curl.yaml
- [x] use-shell-instruction.yaml
- [x] use-workdir.yaml
- [ ] alias-must-be-unique.yaml
- [ ] copy-from-own-alias.yaml
- [x] invalid-port.yaml
- [x] missing-assume-yes-switch.yaml
- [x] multiple-cmd-instructions.yaml
- [x] multiple-entrypoint-instructions.yaml
- [x] last-user-is-root.yaml
- [x] missing-user.yaml

0xDC0DE:
Working on this in branch `generic-to-dockerfile` https://github.com/returntocorp/semgrep-rules/tree/generic-to-dockerfile

0xDC0DE:
[...]

PR #1926, opened 2022-03-25
Generic to dockerfile

0xDC0DE:
See progress in ticket here: https://github.com/returntocorp/semgrep-rules/issues/1793


dockerfile/best-practice/missing-zypper-clean.yaml
```diff
@@ -1,11 +1,11 @@
...
   - id: missing-zypper-clean
     severity: WARNING
-    languages: [generic]
+    languages: [dockerfile]
     patterns:
-      - pattern: zypper $COMMAND
-      - pattern-not-inside: RUN ... && zypper clean
-      - pattern-not-inside: RUN ... && \ zypper clean
+      - pattern: RUN ... zypper ...
+      - pattern-not: RUN ... zypper clean
```
kurt-r2c: We removed enforcing the `&&` condition here - intentional?

Edit: If the user isn't using e.g. `||` over `&&` they could break their container build by having a failed `zypper` invoke followed by a successful `zypper clean` which would look "truthy" to output and cause the layer construction to "succeed".

dockerfile/best-practice/prefer-copy-over-add.yaml
```diff
@@ -17,4 +17,4 @@ rules:
       include:
         - "*dockerfile*"
         - "*Dockerfile*"
-    pattern: ADD
+    pattern: ADD $FROM $TO
```
kurt-r2c, line 20: Can this be an ellipsis or do we need to capture these?

dockerfile/best-practice/remove-package-cache.yaml
```diff
@@ -1,15 +1,13 @@
 rules:
   - id: remove-package-cache
     patterns:
-      - pattern: RUN apt-get update
-      - pattern-not-inside: RUN apt-get update ... && apt-get clean
-      - pattern-not-inside: RUN apt-get update ... &&\ apt-get clean
+      - pattern: RUN ... apt-get update ...
+      - pattern-not: RUN ... apt-get clean ...
```
kurt-r2c: same comment as before re: `zypper` - maybe a separate rule for "incorrect control operators" which is out of scope for this PR.

kurt-r2c (commented):
This looks OK. We should consider a followup rule for incorrect operator usage in the `RUN` directive. 

Also, it looks like the migration from `generic/dockerfile/**` to `dockerfile/**` is incomplete.  Did some of these rules not get moved on purpose or is this an oversight?

kurt-r2c:
This PR is a WIP - see linked ticket from @pieterdc1

kurt-r2c:
blocked by: 
https://github.com/returntocorp/semgrep/issues/4866
https://github.com/returntocorp/semgrep/issues/4881
https://github.com/returntocorp/semgrep/issues/4875

dockerfile/best-practice/remove-package-cache.yaml
0xDC0DE: We actually have a rule for that, set-pipefail.

0xDC0DE:
I would opt to merge the ported rules and create separate pull request for the additional rules when those are unblocked.

dockerfile/best-practice/remove-package-cache.yaml
0xDC0DE: Updated test cases to include `||` operator.

kurt-r2c (approved):
LGTM

kurt-r2c:
@pieterdc1 can you resolve the merge conflicts in `avoid-latest-version` and `missing-pip-no-cache-dir`?

PR #1927, opened 2022-03-25
Add confidence levels to scala rules

minusworld:
Experiment to see how useful confidence levels are in a set of new rules


inkz:
@minusworld great work! I changed `tainted-sql-string` confidence to MEDIUM, because there is still a chance to catch a string that is not SQL, that accidentally has smth like "Insert foobar in blablabbla" in it.
I do not remember exact rate (~20%), but I definitely bumped into a few such examples during testing.

inkz (approved):

PR #1933, opened 2022-03-25
fix for fp of python string interpolation

colleend:
* fixes FP of python string interpolation (makes sure that { in regex is read as a character that leads to interpolation). 

TODO: eventually translate this rule into something other than regex because the regex is hard to read. (and add metavariable-entropy to it)


minusworld (approved):
sweet 👍  Once entropy analysis is back online we could use this here too 🤘

PR #1940, opened 2022-03-28
Corrected fix for autoescape-disabled

0xDC0DE:
This fix behaviour was incorrect.

The fix is adding `autoescape-disabled=True` to the arguments.

But the pattern was also catching cases where `autoescape-disabled` was configured with an incorrect value. In these cases the fix would add it to the arguments a second time instead of correcting the value.


minusworld (approved):
lgtm

PR #1943, opened 2022-03-28
editing http-req to filter out 127.0.0.1

colleend:
filters out http requests of the form 'http://127.0.0.1', as most webapps served locally are accessed without SSL/TLS

_To test:_
`semgrep --test .` in the directory and check that all tests pass.

### DIFF
--- a/problem-based-packs/insecure-transport/go-stdlib/http-request.yaml
+++ b/problem-based-packs/insecure-transport/go-stdlib/http-request.yaml
@@ -4,10 +4,16 @@ rules:
       - pattern-either:
           - pattern: |
               http.$FUNC("=~/[hH][tT][tT][pP]://.*/", ...)
-          - pattern: |
-              $CLIENT := &http.Client{...}
-              ...
-              client.$FUNC("=~/[hH][tT][tT][pP]://.*/", ...)
+          - patterns:
+            - pattern-inside: |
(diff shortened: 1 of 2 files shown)


problem-based-packs/insecure-transport/go-stdlib/http-request.yaml
```diff
@@ -4,10 +4,16 @@ rules:
...
+          - patterns:
+            - pattern-inside: |
+                $CLIENT := &http.Client{...}
+                ...
+            - pattern: |
+                client.$FUNC("=~/[hH][tT][tT][pP]://.*/", ...)
+      - pattern-not:
+          http.$FUNC("=~/[hH][tT][tT][pP]://127.0.0.1.*/", ...)
+      - pattern-not:
+          client.$FUNC("=~/[hH][tT][tT][pP]://127.0.0.1.*/", ...)
```
kurt-r2c, line 16: might also be helpful to ignore `http://localhost/` but we can do that as a future enhancement if necessary.

kurt-r2c (approved):

problem-based-packs/insecure-transport/go-stdlib/http-request.yaml
colleend: yep, good point

PR #1944, opened 2022-03-29
identical_id: Add more wildcards to match in more places

underyx:


kurt-r2c (approved):
Per conversation with @minusworld we are OK with accepting the leading ellipsis in this case due to smaller size of targets. By convention, rule files should not have more than a couple of rule patterns.

PR #1948, opened 2022-03-30
Edit more rules

colleend:
This is the last of the batch! Edited a few more rules to reduce FP and negative developer sentiment:
* `generic/dockerfile/best-practice/missing-zypper-clean.yaml` -- converted this to dockerfile rule, and placed the pattern inside `RUN` to hopefully eliminate more FP
* `java/lang/security/audit/crypto/no-null-cipher.yaml` -- add specific functions the Cipher must call in order for it to lead to more potentially vulnerable situations
* `python/boto3/security/hardcoded-token.yaml` -- added metavariable-analysis to this

_Test Plan:_
Run `semgrep --test .` on the directories the rules are in.

### DIFF
--- a/generic/dockerfile/best-practice/missing-zypper-clean.yaml
+++ b/generic/dockerfile/best-practice/missing-zypper-clean.yaml
@@ -1,9 +1,9 @@
 rules:
   - id: missing-zypper-clean
     severity: WARNING
-    languages: [generic]
+    languages: [dockerfile]
     patterns:
-      - pattern: zypper $COMMAND
+      - pattern: RUN ... zypper $COMMAND ...
       - pattern-not-inside: RUN ... && zypper clean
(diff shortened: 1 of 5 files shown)


generic/dockerfile/best-practice/missing-zypper-clean.yaml
```diff
@@ -1,9 +1,9 @@
 rules:
   - id: missing-zypper-clean
     severity: WARNING
-    languages: [generic]
+    languages: [dockerfile]
     patterns:
-      - pattern: zypper $COMMAND
+      - pattern: RUN ... zypper $COMMAND ...
```
kurt-r2c, line 6: this may cause merge conflicts with https://github.com/returntocorp/semgrep-rules/pull/1926

kurt-r2c (approved):

PR #1949, opened 2022-03-30
add localhost to pattern-not

colleend:
* adds localhost filtering to http-request rule

_To Test:_
Run `semgrep --test .`

### DIFF
--- a/problem-based-packs/insecure-transport/go-stdlib/http-request.yaml
+++ b/problem-based-packs/insecure-transport/go-stdlib/http-request.yaml
@@ -14,6 +14,10 @@ rules:
           http.$FUNC("=~/[hH][tT][tT][pP]://127.0.0.1.*/", ...)
       - pattern-not:
           client.$FUNC("=~/[hH][tT][tT][pP]://127.0.0.1.*/", ...)
+      - pattern-not:
+          http.$FUNC("=~/[hH][tT][tT][pP]://localhost.*/", ...)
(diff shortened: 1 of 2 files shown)


problem-based-packs/insecure-transport/go-stdlib/http-request.yaml
```diff
@@ -14,6 +14,10 @@ rules:
           http.$FUNC("=~/[hH][tT][tT][pP]://127.0.0.1.*/", ...)
       - pattern-not:
           client.$FUNC("=~/[hH][tT][tT][pP]://127.0.0.1.*/", ...)
+      - pattern-not:
+          http.$FUNC("=~/[hH][tT][tT][pP]://localhost.*/", ...)
```
kurt-r2c, line 18: crafted subdomains can bypass this check (and the `127.0.0.1` check):

https://127.0.0.1.requestcatcher.com/
https://localhost.requestcatcher.com/

Probably not important here as users aren't *trying* to evade this regex but maybe a future improvement.

kurt-r2c (approved):

problem-based-packs/insecure-transport/go-stdlib/http-request.yaml
colleend: that's a good point -- hopefully users aren't trying to evade this, but we never truly know 🤔

PR #1950, opened 2022-03-31
Improve the stalebot config

mjambon:
This changes the stalebot config so as to not auto-close issues but require a low-priority label to avoid auto-flagging. This is similar to the config we've been using on the semgrep repo in recent months.


kurt-r2c (approved):
approved per internal discussion

PR #1952, opened 2022-03-31
split bad deserialization rule because a fix exists for YAML

0xDC0DE:


ruby/lang/security/bad-deserialization-yaml.yaml
```diff
@@ -0,0 +1,38 @@
...
+          ...
+          YAML.load(..., $FILE, ...)
+      - pattern-not-inside: |
+          YAML.load(..., File.read("...", ...), ...)
+    fix: Psych.safe_load($...ARGS)
+    message: >-
+      Unsafe deserialization from YAML. Objects in Ruby can be serialized into strings,
+      then later loaded from strings. However, uses of load and object_load can cause remote code execution.
+      Loading user input with YAML can potentially be dangerous. Use JSON in a secure fashion instead.
+      However, loading YAML from a static file is not dangerous and should not be flagged.
```
kurt-r2c, line 27: File writes in the same directory as the static file might be problematic, but we can consider that as a future enhancement.

kurt-r2c (approved):

PR #1966, opened 2022-04-02
fix: reduce false positives on i18next rules

ceefour:
## Link to an issue, if relevant

Improvements for previous rules by myself: #1936 and #1938.


colleend:
Hi! Thanks for submitting this fix -- you can try a regex like: 
`([A-Za-z\n ]+[A-Za-z]+[A-Za-z\n ]+)` 
in `typescript/react/portability/i18next/jsx-not-internationalized.yaml`. This needs to be done because there's always whitespace before the metavariable `$MESSAGE` is captured, so we need to make sure that there's whitespace in the capture groups. 

Would you also mind suggesting an alternative to the patterns in the rule messages too? Thanks for your work!

minusworld (dismissed):

colleend (approved):
thanks for the contribution!

colleend:
@minusworld do you happen to know why CLA isn't passing?

PR #1970, opened 2022-04-03
java.spring.security.cve.cve-2022-22965 improvements

DDuarte:
Less false negatives and less false positives

I'm quite curious if there's a better way to write this rule. I've tried using `metavariable-pattern` for the argument type and argument annotation but couldn't get it to work.

### DIFF
--- a/java/spring/security/cve/cve-2022-22965.yaml
+++ b/java/spring/security/cve/cve-2022-22965.yaml
@@ -103,7 +103,7 @@ rules:
             - pattern: PutMapping
             - pattern: DeleteMapping
             - pattern: PatchMapping
-    message: Semgrep found a match
+    message: Method $METHOD potentially vulnerable to CVE-2022-22965 due to plain binding of a vulnerable POJO type.
     languages:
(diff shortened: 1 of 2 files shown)


java/spring/security/cve/cve-2022-22965.yaml
```diff
@@ -1,43 +1,117 @@
...
+      - metavariable-pattern:
+          metavariable: $MAP
+          pattern-either:
+            - pattern: RequestMapping
+            - pattern: GetMapping
+            - pattern: PostMapping
+            - pattern: PutMapping
+            - pattern: DeleteMapping
+            - pattern: PatchMapping
+    message: Semgrep found a match
```
kurt-r2c:
```suggestion
    message: Method $METHOD potentially vulnerable to CVE-2022-22965 due to plain binding of a vulnerable POJO type.
```

kurt-r2c (changes requested):

kurt-r2c (approved):

PR #1973, opened 2022-04-04
Improvements to ajv-allerrors-true [JavaScript] rule

LewisArdern:
Improving the AJV rule based on potential false negatives it would have caused. This rule checks to make sure it is within an object, vs the first argument of said object. 

Cleaned up some of the language too since it now correctly matches the argument name not lowercase property.


inkz (approved):
lgtm

PR #1976, opened 2022-04-05
Detect call to socket.close() which may be skipped because of exception

Sjord:
If socket.shutdown() raises an exception, socket.close() is not called and won't clean up the socket, which results in memory leaks.

From https://www.paulsprogrammingnotes.com/2021/12/python-memory-leaks.html

### DIFF
--- a/python/correctness/socket-shutdown-close.yaml
+++ b/python/correctness/socket-shutdown-close.yaml
@@ -4,6 +4,18 @@ rules:
     - pattern: |
         $SOCK.shutdown(socket.$A)
         $SOCK.close()
+    - pattern-not-inside: |
+        try:
+            ...
+        except ...:
+            ...
+            $SOCK.close()
+    - pattern-not-inside: |
+        try:
+            ...
+        finally:
+            ...
+            $SOCK.close()
   message: >-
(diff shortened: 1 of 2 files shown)


python/correctness/socket-shutdown-close.yaml
```diff
@@ -0,0 +1,20 @@
+rules:
+- id: socket-shutdown-close
+  patterns:
+    - pattern: |
+        $SOCK.shutdown(socket.$A)
+        $SOCK.close()
```
kurt-r2c, line 6: You may want to consider handling the redundant construction (if you can dream it, a developer has done it)

```python
try:
  socket.shutdown(socket.SHUT_WR)
  socket.close()
except OSError:
  #dirty shutdown, close anyway
  socket.close()
```

kurt-r2c (approved):

python/correctness/socket-shutdown-close.yaml
Sjord: Thanks, I added an `pattern-not-inside` for this.

kurt-r2c (approved):

Issue #1385, opened 2021-07-22
[Rule] HTML construction with JS template strings


minusworld:
**Rule Description**

*What does this rule intend to find?*
Detect HTML constructed in JS template strings.

e.g.,
```js
return `<div style=${style.color}>${content}</div>`
```

**Additional information**

*More information that would help someone write this rule!*

Blocked on https://github.com/returntocorp/semgrep/issues/3578

**PR Checklist**
- [ ] This ticket has links, references, or examples.
- [ ] The rule has **true positive** and **true negative** test cases in a file that matches the rule name.

> If the rule is `my-rule`, the test file name should be `my-rule.js`.
>
> True positives are marked by comments with `ruleid: <my-rule>` and true negatives are marked by comments with `ok: <my-rule>`.

- [ ] The rule has a good message. A good message includes:

> 1. A description of the pattern (e.g., missing parameter, dangerous flag, out-of-order function calls).
> 1. A description of why this pattern was detected (e.g., logic bug, introduces a security vulnerability, bad practice).
> 1. An alternative that resolves the issue (e.g., use another function, validate data first, discard the dangerous flag).

- [ ] After the PR has been reviewed and approved, merge the rule and close this ticket! Thanks for contributing!

Sjord:
I tried creating a rule like this, but ran into a semgrep issue: [Metavariables don't match in JavaScript template strings · Issue #3786 · returntocorp/semgrep](https://github.com/returntocorp/semgrep/issues/3786)

Sjord:
​

PR #1995, opened 2022-04-07
Rule for template strings with interpolation and HTML

Sjord:
E.g.
```
return `<div style=${style.color}>${content}</div>`
```

Closes issue #1385


colleend (approved):
lgtm! thanks for the contribution!

PR #2001, opened 2022-04-08
Fix typo in README

parsiya:
I changed this and removed `ed` from the README.

`we recommended appending audit to your namespace` -> `we recommend`.

I can also replace the next sentence but I did not because it's not wrong.

`that is specifically aiming to detect a vulnerability` -> `that is specifically aimed at detecting a vulnerability`.


ievans (approved):
Ty @parsiya !

PR #2012, opened 2022-04-11
editing sha1 message

colleend:


Sjord:
Recommending HMAC didn't make much sense, since HMAC is a built on top of another hash function. It can work both with SHA1 or SHA256. PBKDF2 is similar, in that it is built on top of HMAC and can work with any hash function. I am not sure if it makes sense to recommend in this context. On the other hand, if someone is hashing passwords with SHA1, you don't want them to switch to SHA256. Does recommending bcrypt make more sense than PBKDF2?

Should we recommend SHA3, or one of the BLAKE functions?

0xDC0DE:
I'll approve this for now. But it looks like we might need to revisit and come up with best recommendations for hashing algorithms in different contexts.

0xDC0DE (approved):

colleend:
@Sjord makes a good point (its confusing recommending HMAC/PBKDF2 when you can use SHA-1 underneath). My thought process in recommending that is to say that you should think of using another encryption algorithm instead of the SHA functions in password hashing situations and in the scenario you're using the SHA1 function with `H(secret || message)`, which leads to length-extension attacks, though that might actually be fit to be its own separate rule (it just might lead to duplicate findings when running in addition to this rule which is hard). 

We'll definitely revisit this and standardize our crypto recommendations based on situation (as we have this issue with other crypto rules too), but will merge for now. 

Thanks for the discussion, this is super helpful!

PR #2016, opened 2022-04-13
Let detect-insecure-websocket match word boundary

### LINKED ISSUE #2004: detect-insecure-websocket incorrectly triggers on aws URLs
I have an URL like `aws://${accountId}/eu-west-1`, and [detect-insecure-websocket](https://github.com/returntocorp/semgrep-rules/blob/develop/javascript/lang/security/detect-insecure-websocket.yaml) thinks it is an insecure websocket.

Sjord:
So that it matches ws://, and not anythingws://.

Closes #2004.


colleend (approved):
thanks for doing this, lgtm!

PR #2022, opened 2022-04-15
Add `dynamodb-request-object` rule

inkz:
The rule to highlight when data from AWS Lambda event object is passed directly to DynamoDB client as a query parameters

### DIFF
--- a/python/aws-lambda/security/dynamodb-filter-injection.yaml
+++ b/python/aws-lambda/security/dynamodb-filter-injection.yaml
@@ -1,5 +1,5 @@
 rules:
-- id: dynamodb-request-object
+- id: dynamodb-filter-injection
   mode: taint
   metadata:
     cwe: "CWE-943: Improper Neutralization of Special Elements in Data Query Logic"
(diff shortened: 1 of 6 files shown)


kurt-r2c (approved):

python/aws-lambda/security/dynamodb-filter-injection.yaml
```diff
@@ -0,0 +1,47 @@
...
+    - pattern: |
+        {...}
+  pattern-sinks:
+  - patterns:
+    - pattern: $SINK
+    - pattern-either:
+      - pattern-inside: $TABLE.scan(..., ScanFilter = $SINK, ...)
+      - pattern-inside: $TABLE.query(..., QueryFilter = $SINK, ...)
+    - pattern-either:
+      - patterns:
```
kurt-r2c, line 35: this `patterns` statement is a great example of well-written `pattern-inside` usage.

kurt-r2c (approved):

python/aws-lambda/security/dynamodb-filter-injection.yaml
inkz: thank you!

PR #2023, opened 2022-04-15
Add fs-directory-listing rule (Go)

lfama:
Simple rule to check if 'http.FileServer' is used which can lead to directory listing vulnerability.


colleend (approved):
lgtm, thanks for this submission!

PR #2025, opened 2022-04-19
First batch of rules for the Pyramid framework

p4p3r:
This PR contains rules for the Pyramid Python framework that cover:
- Cookie flag security
- CSRF configuration (global and per-view)
- SQLi related to SQLAlchemy usage (similar to existing rules for SQLAlchemy)
- XSS related to a few cases of Response object creation without using a template renderer

### DIFF
--- a/python/pyramid/audit/set-cookie-samesite.yaml
+++ b/python/pyramid/audit/set-cookie-samesite.yaml
@@ -15,7 +15,9 @@ rules:
                   ...
       - pattern-not: $RESPONSE.set_cookie(..., samesite='Lax', ...)
       - pattern-not: $RESPONSE.set_cookie(..., **$PARAMS)
-      - pattern: $RESPONSE.set_cookie(..., samesite=$SAMESITE, ...)
+      - pattern-either:
+          - pattern: $RESPONSE.set_cookie(..., samesite=$SAMESITE, ...)
+          - pattern: $RESPONSE.set_cookie(...)
     message: >-
       Found a Pyramid cookie without the samesite option correctly set. Pyramid
       cookies should be handled securely by setting samesite='Lax' in response.set_cookie(...).
--- a/python/pyramid/audit/set-cookie-samesite.yaml
+++ b/python/pyramid/audit/set-cookie-samesite.yaml
@@ -14,6 +14,7 @@ rules:
                   $RESPONSE = pyramid.httpexceptions.HTTPFound(...)
                   ...
       - pattern-not: $RESPONSE.set_cookie(..., samesite='Lax', ...)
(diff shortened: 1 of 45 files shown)


0xDC0DE (approved):
I think [this](https://github.com/returntocorp/semgrep-rules/pull/2025/files#diff-016244119020973a7b2457d11f1b228f3e90dfe055fdeb348f290bd525805a59) rule can be split into two rules that can have fix options (but I'm unsure about the `**params` option)

You could have a rule for the missing httpflag, and one for the httpflag being set to false.
Example here: https://github.com/returntocorp/semgrep-rules/tree/develop/python/jinja2/security/audit

0xDC0DE:
Same argument for [python/pyramid/audit/authtkt-cookie-secure.yaml](https://github.com/returntocorp/semgrep-rules/pull/2025/files#diff-d80e994da8034265493ae4bb814233f3b28fa1e7a9ecd0bb26c259ac58179d39)

0xDC0DE:
And this one python/pyramid/audit/set-cookie-httponly.yaml

0xDC0DE:
But really, I love that you already wrote a lot of fixes for the other rules!

0xDC0DE:
I think you can add a fix to this rule as well: python/pyramid/audit/set-cookie-samesite.yaml

python/pyramid/audit/set-cookie-samesite.yaml
```diff
@@ -0,0 +1,31 @@
...
+                  $RESPONSE = $REQUEST.response
+                  ...
+          - pattern-inside: |
+              def $VIEW(...):
+                  ...
+                  $RESPONSE = pyramid.httpexceptions.HTTPFound(...)
+                  ...
+      - pattern-not: $RESPONSE.set_cookie(..., samesite='Lax', ...)
+      - pattern-not: $RESPONSE.set_cookie(..., **$PARAMS)
+      - pattern: $RESPONSE.set_cookie(..., samesite=$SAMESITE, ...)
```
0xDC0DE: It looks like this rule only checks for the case where `samesite` is set, but incorrectly. Is the default value secure?

0xDC0DE (approved):

python/pyramid/audit/set-cookie-samesite.yaml
p4p3r: Fixed in 4ea7a800e8b951fceca2d6876a6e51c060a17fbe

p4p3r:
@pieterdc1 I split all the cookie flags rules into default-value/unsafe-value and added fixes to all the rules.

0xDC0DE (approved):
I like your naming scheme for the split rules, I will be using that from now on.
`...-unsafe-default` and `...-unsafe-value`.

Issue #2024, opened 2022-04-18
False positive for rule java.lang.security.audit.formatted-sql-string


dgian:
**Describe the bug**
`java.lang.security.audit.formatted-sql-string` rule is matched when using `execute` named in a method, outside of SQL context.

**To Reproduce**
https://semgrep.dev/s/0WKr

**Expected behavior**
Should not report `execute` as a match for SQL injection

**Priority**
How important is this to you?
- [ ] P0: blocking me from making progress
- [x] P1: this will block me in the near future
- [ ] P2: annoying but not blocking me

colleend:
Should be fixed by the PR mentioned above! Feel free to let us know if there are more issues 😄

dgian:
> Should be fixed by the PR mentioned above! Feel free to let us know if there are more issues smile

Many thanks @colleend . This was very quick! 👍🏽 
Do you have an idea of when is this going to be included in a registry release? Thanks again

colleend:
Hi! It will most likely take until Monday or Tuesday for these changes to be reflected in the registry.

PR #2026, opened 2022-04-19
edit rule to make sure there's not too many FPs

colleend:
should solve https://github.com/returntocorp/semgrep-rules/issues/2024. 

* this will reduce FPs with the usage of typed metavariables. However, there will be cases we don't catch as a result (because we're specifying the types). 

_To test:_
run `semgrep --test .` in the directory.

### DIFF
--- a/java/lang/security/audit/formatted-sql-string.java
+++ b/java/lang/security/audit/formatted-sql-string.java
@@ -67,6 +67,7 @@ public class SqlExample2 {
 
     public List<AccountDTO> findAccountsById(String id) {
         String jql = "from Account where id = '" + id + "'";
+        EntityManager em = emfactory.createEntityManager();
         // ruleid:formatted-sql-string
         TypedQuery<Account> q = em.createQuery(jql, Account.class);
         return q.getResultList()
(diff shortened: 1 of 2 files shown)


java/lang/security/audit/formatted-sql-string.java
```diff
@@ -117,3 +119,16 @@ public void tableConcat() {
...
+
+// This whole operation has nothing to do with SQL
+public class FalsePositiveCase {
+    private ApiClient apiClient; // imagine an ApiClient class that contains a method named execute
+
+    public void test(String parameter) throws ApiException {
+        com.squareup.okhttp.Call call = constructHttpCall(parameter); // Create OKHttp call using parameter from outside
+        apiClient.execute(call);
+        apiClient.execute(call);
+        apiClient.run(call); // proof that 'execute' name is causing the false-positive
```
kurt-r2c, line 133: These should have `//ok: formatted-sql-string` annotations

kurt-r2c (changes requested):
Let's be explicit about negative test cases - this is particularly helpful when refactoring rules later on

java/lang/security/audit/formatted-sql-string.java
colleend: ah, good catch 👍

kurt-r2c (approved):

PR #2046, opened 2022-04-27
increasing historical semgrep version

colleend:
This action:

* runs every week
* increases the semgrep-version in `semgrep-rules-test-historical.yml` by 1
* creates a PR with version increase

This automates the need to increase the semgrep-version manually in `semgrep-rules-test-historical.yml` every week. This makes sure the historical semgrep-version test correlates to ~1 month behind our current semgrep version test.

_To Test:_
Go to `actions`
Run `increase-historical-semgrep-script.yml` manually.
Check that this action makes a PR that increases the version number for `semgrep-rules-test-historical.yml`. 

Example PR this creates: https://github.com/returntocorp/semgrep-rules/pull/2047

NOTE:
This workflow uses a PAT that I've generated. Perhaps it would be better to generate a PAT with a company-wide account instead. @minusworld what do you think?


enncoded (approved):

enncoded:
LGTM, but I'll hold off on merging it so someone else can look it over as well beforehand

colleend:
@minusworld this is the PR I mentioned! Would you mind taking a look? 🙏

minusworld:
wdyt about instead of incrementing the version (we might not release every week), we download the latest version, get the version, and subtract some number (4?) from the minor version? Or something simlar?

colleend:
ohhh that's a good idea, though it might be easier doing that in the semgrep-historical-test yml before running the test commands. I'll try that out!

colleend:
@minusworld changed the original script to use a version 3 off of current semgrep-version. Not sure why it's running `validate-r2c-registry-metadata` currently, but this should be good.

enncoded:
Ooh I just merged @minusworld's PR to add `validate-r2c-registry-metadata`. Might need to have that check ignore other workflows files.

.github/workflows/semgrep-rules-test-historical.yml
```diff
@@ -16,15 +16,26 @@ jobs:
...
+      run: |
+        SEMGREP_VERSION=`semgrep --version`
+        SEMGREP_VERSION_SHORT=${SEMGREP_VERSION:2:2}
+        SEMGREP_OLD_VERSION="0.$(($SEMGREP_VERSION_SHORT-3)).0"
+        echo "SEMGREP_OLD_VERSION=$SEMGREP_OLD_VERSION" >> $GITHUB_ENV
+    - name: validate rules on historical semgrep version
       run: |
-        docker run --rm -v ${GITHUB_WORKSPACE}/semgrep-rules:/src returntocorp/semgrep:0.87.0 semgrep --validate --config /src
-    - name: test with semgrep v0.87.0 
+        docker run --rm -v ${GITHUB_WORKSPACE}/semgrep-rules:/src returntocorp/semgrep:${SEMGREP_OLD_VERSION} semgrep semgrep --validate --config /src
```
enncoded: Is the `semgrep semgrep` on line 38 intentional?

colleend: It was, I was getting a deprecation error with semgrep, but I don't think that's the actual problem 😓  I'll remove it!

colleend:
Ah gotcha, that makes sense. It's complaining about lack of `owasp` key on some of the recently merged laravel rules, so that might be the issue 😓

minusworld:
That check's not required atm, so it can be ignored for this PR

.github/workflows/semgrep-rules-test-historical.yml
```diff
@@ -16,15 +16,26 @@ jobs:
...
+      run: pip3 install semgrep
     - name: delete stats directory
       run: rm -rf semgrep-rules/stats
     - name: delete fingerprints directory
       run: rm -rf semgrep-rules/fingerprints
-    - name: validate rules on semgrep v0.87.0
+    - name: grab historical semgrep version
+      run: |
+        SEMGREP_VERSION=`semgrep --version`
+        SEMGREP_VERSION_SHORT=${SEMGREP_VERSION:2:2}
```
minusworld, line 33: nit: This is technically the Semver "minor version". No need to change unless you want to

minusworld (approved):
lgtm! Have you been able to test it out?

colleend:
yep, the version-grabbing has been working for me locally 👍

PR #2050, opened 2022-04-29
add `tainted-deserialization` rule for Ruby

inkz:
simple rule to catch tainted deserialization calls in AWS Lambda functions

### DIFF
--- a/ruby/aws-lambda/security/tainted-deserialization.yaml
+++ b/ruby/aws-lambda/security/tainted-deserialization.yaml
@@ -1,6 +1,7 @@
 rules:
 - id: tainted-deserialization
   mode: taint
+  languages: [ruby]
   message: >-
     Deserialization of a string tainted by `event` object found. Objects in Ruby can be serialized into strings,
     then later loaded from strings. However, uses of `load` can cause remote code execution.
--- a/ruby/aws-lambda/security/tainted-deserialization.yaml
+++ b/ruby/aws-lambda/security/tainted-deserialization.yaml
@@ -1,0 +1,35 @@
+rules:
+- id: tainted-deserialization
+  mode: taint
+  message: >-
+    Deserialization of a string tainted by `event` object found. Objects in Ruby can be serialized into strings,
+    then later loaded from strings. However, uses of `load` can cause remote code execution.
+    Loading user input with MARSHAL or CSV can potentially be dangerous. Use JSON in a secure fashion instead.
+  metadata:
+    references:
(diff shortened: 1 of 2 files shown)


enncoded:
I feel like the message essentially telling people they have no choice but to switch to a new format that they're not already using might be offputting, especially if it's a fairly big codebase that's already built around using yaml or something similar. I think adding some more explanation would be helpful. Two options I can think of:

1. Adding something similar to the sentence `If you need to deserialize untrusted data, you should use JSON as it is only capable of returning 'primitive' types` from [this page](https://ruby-doc.org/core-3.1.2/doc/security_rdoc.html) would be helpful.

2. Suggest safe alternate methods and libraries for deserialization for the file formats that they may already be using, such as the ones outlined [here](https://codeql.github.com/codeql-query-help/ruby/rb-unsafe-deserialization/). Paraphrasing this sentence seems particularly useful: `Using JSON.parse and YAML.safe_load instead, as in the following example, removes the vulnerability. Similarly, calling Oj.load with any mode other than :object is safe, as is calling Oj.safe_load. Note that there is no safe way to deserialize untrusted data using Marshal.`

ruby/aws-lambda/security/tainted-deserialization.yaml
```diff
@@ -0,0 +1,36 @@
+rules:
+- id: tainted-deserialization
+  mode: taint
+  languages: [ruby]
+  message: >-
+    Deserialization of a string tainted by `event` object found. Objects in Ruby can be serialized into strings,
+    then later loaded from strings. However, uses of `load` can cause remote code execution.
+    Loading user input with MARSHAL or CSV can potentially be dangerous. Use JSON in a secure fashion instead.
```
enncoded: I feel like the message essentially telling people they have no choice but to switch to a new format that they're not already using might be offputting, especially if it's a fairly big codebase that's already built around using yaml or something similar. I think adding some more explanation would be helpful. Two options I can think of:

1. Adding something similar to the sentence `If you need to deserialize untrusted data, you should use JSON as it is only capable of returning 'primitive' types` from [this page](https://ruby-doc.org/core-3.1.2/doc/security_rdoc.html) would be helpful.

2. Suggest safe alternate methods and libraries for deserialization for the file formats that they may already be using, such as the ones outlined [here](https://codeql.github.com/codeql-query-help/ruby/rb-unsafe-deserialization/). Paraphrasing this sentence seems particularly useful: `Using JSON.parse and YAML.safe_load instead, as in the following example, removes the vulnerability. Similarly, calling Oj.load with any mode other than :object is safe, as is calling Oj.safe_load. Note that there is no safe way to deserialize untrusted data using Marshal.`

enncoded (approved):

PR #2054, opened 2022-04-29
Add a schema for validating metadata

minusworld:
This PR adds a schema for validating the fields in `metadata` sections that the Semgrep registry ecosystem has come to expect. 

This will help with:
* Ensuring rules in the registry are tagged appropriately so that rules and associated findings get routed to the right place
* Ensures a baseline set of metadata all rules have so that OSS users can reliably parse them
* Allows the Semgrep registry to additionally validate the metadata of incoming rules (from the playground or other sources)

This PR adds:
* A metadata schema. This schema requires a small subset of fields (references, technology, category, and for security rules, owasp and cwe) and validates the format of several others.
* A script that will check the metadata section of rules against the schema
* A GHA workflow that validates incoming rules' metadata with this script


minusworld:
Oh yeah, and here's an example of the output:
```
('bash/curl/security/curl-eval.yaml', 'required', '', "'references' is a required property. Please include at least one URL with more information about this rule in a metadata field called 'references'.")
('bash/curl/security/curl-eval.yaml', 'required', '', "'cwe' is a required property. Please include a 'cwe' metadata field for security rules. Format: CWE-nnn, where nnn is the CWE number. See https://cwe.mitre.org/ for more info.")
```

enncoded (approved):

PR #2059, opened 2022-05-03
Fixing OWASP saxparserfactory rule

coheigea:
As per https://cheatsheetseries.owasp.org/cheatsheets/XML_External_Entity_Prevention_Cheat_Sheet.html#jaxp-documentbuilderfactory-saxparserfactory-and-dom4j if doctypes are disabled, no further features are required.


contrib/owasp/java/xxe/saxparserfactory.yaml
```diff
@@ -23,16 +23,30 @@ rules:
       - pattern-not-inside: |
           $RETURNTYPE $METHOD(...) {
             ...
-            $XXX.setFeature("http://apache.org/xml/features/disallow-doctype-decl", true);
-            $XXX.setFeature("http://xml.org/sax/features/external-general-entities", false);
-            $XXX.setFeature("http://xml.org/sax/features/external-parameter-entities", false);
+            $SPF.setFeature("http://xml.org/sax/features/external-general-entities", false);
```
minusworld: Semgrep won't understand that these settings can be set in any order. You'll have to either: make a `pattern-not-inside` for each ordering, or reconfigure the rule so that there are three `pattern-not-inside`s for each setting.

Something like this:

```
patterns:
- pattern-not-inside: |
    $SPF.setFeature(<f1>);
    ...
- pattern-not-inside: |
    $SPF.setFeature(<f2>);
    ...
- pattern: $SAXPARSER.parse(...);
```

coheigea: @minusworld I decided to simplify the PR and just fix the issue which is that only doctypes are required to be disabled to be secure, the other two features aren't needed.

minusworld (approved):

PR #2063, opened 2022-05-03
Add Spring taint rules

minusworld:
Add 5 taint rules for Spring

### DIFF
--- a/java/spring/security/injection/tainted-file-path.yaml
+++ b/java/spring/security/injection/tainted-file-path.yaml
@@ -42,6 +42,9 @@ rules:
       - pattern-either:
           - pattern-either:
               - pattern: new FileInputStream(...)
+              - pattern: new java.io.FileInputStream(...)
               - pattern: new FileReader(...)
+              - pattern: new java.io.FileReader(...)
               - pattern: new File(...)
+              - pattern: new java.io.File(...)
               - pattern: $SOMETHING.getResourceAsStream(...)
--- a/java/spring/security/injection/tainted-file-path.yaml
+++ b/java/spring/security/injection/tainted-file-path.yaml
@@ -13,8 +13,8 @@ rules:
     metadata:
       cwe: "CWE-23: Relative Path Traversal"
       owasp:
-        - A03:2021
-        - A01:2017
+        - A03:2021 - Injection
+        - A01:2017 - Injection
       references:
       - https://owasp.org/www-community/attacks/Path_Traversal
       category: security
--- a/java/spring/security/injection/tainted-file-path.yaml
+++ b/java/spring/security/injection/tainted-file-path.yaml
@@ -3,14 +3,20 @@ rules:
     languages:
       - java
     severity: ERROR
-    message:  TODO
+    message: >-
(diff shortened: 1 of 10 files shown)


java/spring/security/injection/tainted-file-path.yaml
```diff
@@ -0,0 +1,51 @@
...
+      Detected user input creating a file path. An attacker could control
+      the location of this file, to include going backwards in the directory
+      with '../'. To address this, ensure that user-controlled variables in file
+      paths are sanitized. You may also consider using a utility method such as
+      org.apache.commons.io.FilenameUtils.getName(...) to only retrieve the file
+      name from the path.
+    metadata:
+      cwe: "CWE-23: Relative Path Traversal"
+      owasp:
+        - A03:2021
```
LewisArdern, line 16: missing names after owasp

java/spring/security/injection/tainted-file-path.yaml
```diff
@@ -0,0 +1,47 @@
...
+                    ...
+                  }
+          - metavariable-regex:
+              metavariable: $REQ
+              regex: (RequestBody|PathVariable|RequestParam)
+          - pattern: $SOURCE
+    pattern-sinks:
+      - pattern-either:
+          - pattern-either:
+              - pattern: new FileInputStream(...)
```
kurt-r2c, line 44: we might want to make these explicit, e.g. `java.io.FileInputStream`

java/spring/security/injection/tainted-sql-string.yaml
```diff
@@ -52,6 +58,19 @@ rules:
...
+                    - pattern: String.format($VAR, ...)
               - metavariable-regex:
                   metavariable: $SQLSTR
                   regex: (?i)(select|delete|insert|create|update|alter|drop)\b
+    pattern-sanitizers:
+    - pattern-either:
+      - pattern: System.out.println(...)
+      - pattern: $LOG.info(...)
+      - pattern: $LOG.warning(...)
+      - pattern: $LOG.debug(...)
```
kurt-r2c: This will incorrectly sanitize the (not uncommon) case of:

```java
string query = "SELECT * FROM users WHERE id = '" + user_id + "'";
Logger.debug("running SQL query: " + query);
...
connector.execute(query);
```

SQL injection often comes with other code smell like inappropriate logging.

kurt-r2c (commented):
One thought we should consider before shipping (and one nitpick about typechecking)

java/spring/security/injection/tainted-sql-string.yaml
minusworld: Ah, good thinking. I think I can switch the sanitizers to just use pattern-nots without adversely affecting things

java/spring/security/injection/tainted-file-path.yaml
minusworld: I'm not actually clear on the behavior differences between the two. Should I just add both? Or only the explicit ones?

kurt-r2c (approved):

PR #2066, opened 2022-05-04
Create rule for dangerous spring boot actuator endpoints

Gby56:
The original rule, spring-actuator-fully-enabled, only checks for fully enabled actuator endpoints.
We want to make a separate rule that checks for specific dangerous endpoints, which should catch way more things.


minusworld:
`management.endpoints.web.exposure.include=*` is pretty common out in the wild. I found ~90 instances of this in a few hundred repositories. What does this setting do, and does it make sense to include in this rule?

(Not a perfect search, but here's what comes back in GitHub: https://github.com/search?l=Java+Properties&q=management.endpoints.web.exposure.include%3D%5C%2A&type=Code)

Gby56:
I agree I was a bit worried on how widespread that config would be, exposure is not the same as enabling something. Maybe there could be another condition to match unauthenticated exposure but that's more complex.

https://docs.spring.io/spring-boot/docs/2.1.11.RELEASE/reference/html/production-ready-endpoints.html
The wildcard means that if anyone turns on one of the features it will get exposed by default on web, so it is indeed a bit more dangerous but having some pre-requisite could help, say, combining web * with enabled heapdump

minusworld:
Also, I tried filtering out comments (lines beginning with `#`) but that appears to be malfunctioning at the moment. I filed https://github.com/returntocorp/semgrep/issues/5148 to hopefully address this 😅

minusworld:
> I agree I was a bit worried on how widespread that config would be, exposure is not the same as enabling something. Maybe there could be another condition to match unauthenticated exposure but that's more complex.
> 
> https://docs.spring.io/spring-boot/docs/2.1.11.RELEASE/reference/html/production-ready-endpoints.html The wildcard means that if anyone turns on one of the features it will get exposed by default on web, so it is indeed a bit more dangerous but having some pre-requisite could help, say, combining web * with enabled heapdump

Does it make sense to factor out that case into its own rule? That way we can address the special cases for it separately. 🤷🏽

Gby56:
@minusworld I'll rework the rule to have a better signal yes, I do think the web exposure wildcard has a use case, just like the wildcard enabled, but it's a bit too much for that one. I'll only include specific endpoints here then.

Gby56:
it's confusing, removing the "      - pattern: management.endpoints.web.exposure.include=*" does not remove the finding in the example, so this group:
```
      - pattern: management.endpoints.web.exposure.include=...heapdump...
      - pattern: management.endpoints.web.exposure.include=...threaddump...
      - pattern: management.endpoints.web.exposure.include=...trace...
      - pattern: management.endpoints.web.exposure.include=...logfile...
      - pattern: management.endpoints.web.exposure.include=...env...
      - pattern: management.endpoints.web.exposure.include=...jolokia...
```
is quite fuzzy
and adding this `    - pattern-not: management.endpoints.web.exposure.include=*` doesn't fix it either 🤔 
EDIT: switching to regex helped

Gby56:
I was hoping to do something like that to combine several rules:
```
rules:
  - id: spring-actuator-dangerous-endpoints
    patterns:
      - pattern-regex: management.endpoints.web.exposure.include=(jolokia|heapdump|threaddump|jolokia|trace|env|logfile)
      - pattern-regex: management.endpoint.(jolokia|heapdump|threaddump|jolokia|trace|env|logfile).enabled=true
      - pattern: management.endpoints.enabled-by-default=true
```
That way the rule would only trigger IF:

- some web exposure is present
- some management endpoint is enabled
- enabled by default is also true (could be good to put it in an either group instead)

```
  - id: spring-actuator-dangerous-endpoints
    pattern-either:
      - patterns:
        - pattern-regex: management.endpoints.web.exposure.include=(jolokia|heapdump|threaddump|jolokia|trace|env|logfile)
        - pattern-regex: management.endpoint.(jolokia|heapdump|threaddump|jolokia|trace|env|logfile).enabled=true
      - pattern: management.endpoints.enabled-by-default=true
```
That would give a good signal imo, maybe metavariable is needed here

minusworld:
Hm, OK. There might be some issues with how `generic` is behaving here.

Let me take a look and see if I can identify a working path forward for what you've proposed.

minusworld:
I'm not quite sure how Spring Boot Actuator works, so I'm relying on you for context-specific understanding 😅 

I think if I understood you correctly, you want to alert when and endpoint is both ENABLED and EXPOSED. If so, this is what I worked on: https://semgrep.dev/s/GX31

Notably, `generic` ellipses only capture 10 lines, so adding more will hopefully catch it.

PR #2067, opened 2022-05-04
javascript taint rules

LewisArdern:


minusworld:
@enncoded @LewisArdern What's the difference between the two `sendFile` rules?

LewisArdern:
> @enncoded @LewisArdern What's the difference between the two `sendFile` rules?

@minusworld  - One is more noisy due to either join-mode or deep semgrep the one which is important:

* https://github.com/returntocorp/semgrep-rules/pull/2067/files#diff-b04e022ed229a997ccb8d43397e1662db6df933ecc6c17fbe8c4c73c8fa983ba - this is less noisy
* https://github.com/returntocorp/semgrep-rules/pull/2067/files#diff-9b1f22506ec5e01d367bc475315b26a032588a4c55e5fe30507a4a4815b622e8 - this looks for type string which may not come from user-input, but without it we will miss critical findings.

minusworld (approved):

minusworld:
Failing historical test is OK because the rule still runs on the older version. It just doesn't find all the same things.

PR #2071, opened 2022-05-05
Fixing false positive for JWT rule

coheigea:
The JWT detection rule has a false positive on a java expression that starts with eyJ and has a "." in it. I fixed it by adding minimum characters both to the header (Base-64 URL encoded version of {"alg":""} which is a required header for JWS) and the payload (BASE-64 URL encoded version of {"":""}).


coheigea:
The linting error does not seem to be related to this PR

minusworld (approved):

PR #2072, opened 2022-05-05
Remove owasp, cwe, and references metalints because this is handled by the schema validator now

minusworld:


kurt-r2c (approved):

kurt-r2c:
Dogfooding is good, schema validation is better

PR #2076, opened 2022-05-07
Add more acceptable cases for python requests use-raise-for-status rule

djmattyg007:
Sometimes the actual request will be made inside of a context manager, but you don't want to check the status until after the context manager is closed. For example, your context manager may have resulted in opening a file or a different network socket, and closing it ASAP is desirable. You're still checking the status of the request though, which should be fine.


minusworld:
The important test here is `semgrep-rules-test-develop`.

It looks like the rule is reporting on line 108, which isn't marked with `ruleid`. If it should be alerting here, that needs to be added. If it doesn't, then the rule is flagging on a line it shouldn't.

djmattyg007:
@minusworld I've updated the PR with a couple of small improvements. The relevant tests should be passing now.

minusworld (approved):

PR #2081, opened 2022-05-10
Reduce findings span - part 1

p4p3r:
The purpose of this PR is to reduce the number of lines of the findings reported by rules that have shown to be matching on average very large ranges.

Marked as draft while the impact is assessed.


python/jwt/security/jwt-hardcode.py
```diff
@@ -1,25 +1,27 @@
...
     # ruleid: jwt-python-hardcoded-secret
-    encoded = jwt.encode({'some': 'payload'}, 'secret', algorithm='HS256')
+    encoded = jwt.encode({"some": "payload"}, "secret", algorithm="HS256")
     return encoded
 
+
 def bad2():
     # ruleid: jwt-python-hardcoded-secret
-    encoded = jwt.encode({'some': 'payload'}, secret_const, algorithm='HS256')
+    encoded = jwt.encode({"some": "payload"}, secret_const, algorithm="HS256")
```
0xDC0DE, line 19: Hmmm. Single quoted strings are allowed in python, no? Should the rule capture both cases?

python/lang/correctness/baseclass-attribute-override.py
```diff
@@ -12,8 +10,16 @@ def method2(self, args):
 
 class B:
     def method1(self, args):
-        print('hello there')
+        print("hello there")
```
0xDC0DE: Should there be test cases for both single and double quoted strings?

python/lang/correctness/baseclass-attribute-override.py
```diff
@@ -12,8 +10,16 @@ def method2(self, args):
...
 
+
+# ruleid: baseclass-attribute-override
 class C(A, B):
     def __init__():
-        print('initialized')
+        print("initialized")
```
0xDC0DE, line 19: Should there be test cases for both single and double quoted strings?

0xDC0DE (commented):
Deleting and renaming files will not properly propagate to the registry.

0xDC0DE (approved):

python/jwt/security/jwt-hardcode.py
p4p3r: I didn't notice these changes... vscode autoformatted and replaced single quotes with double quotes...

python/lang/correctness/baseclass-attribute-override.py
p4p3r: This case of single quotes doesn't really matter because the rule is about class inheritance, the print statement is not involved in the pattern.

python/lang/correctness/baseclass-attribute-override.py
p4p3r: This case of single quotes doesn't really matter because the rule is about class inheritance, the print statement is not involved in the pattern.

python/jwt/security/jwt-hardcode.py
p4p3r: Added an additional test case in cbdad21b350a0cbe56155a93d13a7055e6072970

p4p3r:
> Deleting and renaming files will not properly propagate to the registry.

Oh well then it means this PR cannot be merged as it is!

PR #2082, opened 2022-05-11
Improve prototype pollution rules

dottif:
Fix a problem with the function rule, which may not detect some vulnerabilities because it does not consider the case of a “named” function declaration if it is directly assigned to the exports object (or also any other object). 
An example is https://github.com/eggjs/extend2/blob/ddefb57595a369eeec88d001786161389e1ee406/index.js#L26
If the function is declared and then exported the rule can detect it.
```javascript
//Not detected
var x;
x = function vuln() { ... vuln() ... }

module.exports = function vuln() { ... vuln() ...};


//Detected
function vuln() { ... vuln() ... };

const vuln = function() { ... vuln() ... };
```
Also update the pattern-not-inside patterns of the assignment rule to avoid some false positives, as in the pattern-sanitizers of the function rule.


minusworld:
@inkz and @LewisArdern, would you mind taking a look at this one?

LewisArdern:
@minusworld I'll take this on and verify

LewisArdern:
@dottif  I looked seeing how this would work in taint for the express example https://semgrep.dev/s/b1Rd  I noticed that from the current JS file, it seems that one of the rules which is meant to be safe, seems unsafe? 
```
    items[req.query.name] = req.query.text;
```
This technically should be an issue? Unless I am missing something

Overall your rule changes look good! I made one change to reduce a false positive, and left one comment if you could address it

LewisArdern:
I would love to read your thesis too, feel free to DM me over at lewis@r2c.dev would like to send you some swag for adding semgrep into your analysis!

javascript/lang/security/audit/prototype-pollution/prototype-pollution-function.yaml
```diff
@@ -68,6 +68,24 @@ rules:
...
+              - pattern-inside: |
+                  $OBJ = function $NAME(...) {
+                    ...
+                    $NAME(...)
+                    ...
+                  }
+              - pattern-inside: |
+                  $OBJ = function $NAME(...) {
+                    ...
+                    $THIS.$NAME(...)
```
LewisArdern: Can this be this.$NAME? may reduce false positives if its actually intended to be this

LewisArdern (changes requested):
Left a few comments and made one edit, please let me know your thoughts when you get a chance!

javascript/lang/security/audit/prototype-pollution/prototype-pollution-function.yaml
dottif: > @dottif I looked seeing how this would work in taint for the express example https://semgrep.dev/s/b1Rd I noticed that from the current JS file, it seems that one of the rules which is meant to be safe, seems unsafe?
> 
> ```
>     items[req.query.name] = req.query.text;
> ```
> 
> This technically should be an issue? Unless I am missing something
> 
> Overall your rule changes look good! I made one change to reduce a false positive, and left one comment if you could address it

The example of the `ok1` function is not vulnerable.
The user can't control the property name (it's hardcoded to "id"), as in the first example, and setting it to `__proto__`, so the variable `items` can't then refer to `Object.prototype`.

The `app.post` case is vulnerable, the check should be done on `id` not on `req.query.name`, also in `let items = req.session.todos[id1];` the `1` is a typo?

Also, by rewriting the rule that way (targeting only express cases) then it will not consider other assignments cases like
```javascript
\\two step
function foo(key1, key2, value) {
  var target = {};
  var tmp = target[key1];
  tmp[key2] = value;
}
```
Also, right now the assignment rule does not consider “direct” assignments that could lead to prototype pollution, and can only detect “two-step" assignments. I don't know if there is a way to match also the direct assignments without causing too many false positives.
```javascript
\\direct assignments 
function foo1(obj, a, b, val) {
  obj[a][b] = val;
}
function foo2(obj, a, b, c, val) {
  obj[a][b][c] = val;
}
```

> Can this be this.$NAME? may reduce false positives if its actually intended to be this

Very probably not, this one can be removed.

LewisArdern: Thanks for the informative response!

LewisArdern (approved):

LewisArdern:
Merged! Thank you for you addition to semgrep @dottif

Issue #1421, opened 2021-08-17
django nontext-field-must-set-null-true false positive on django.contrib.gis.db.models


tino:
**Describe the bug**
The `rule:python.django.correctness.nontext-field-must-set-null-true.nontext-field-must-set-null-true` triggers false positives when `models` is imported from `django.contrib.gis.db`.

**To Reproduce**
```python
from django.contrib.gis.db import models

class MyModel(models.Model):
    shape = models.PolygonField(dim=3, srid=4326)
    grouper_id = models.CharField(max_length=50, blank=True)  # <-- fails on this line
```
output:
```bash
❯ semgrep --config https://semgrep.dev/p/django app/geodata/models.py
using config from https://semgrep.dev/p/django. Visit https://semgrep.dev/registry to see all public rules.
downloading config...
running 35 rules...
100%...
app/geodata/models.py
severity:error rule:python.django.correctness.nontext-field-must-set-null-true.nontext-field-must-set-null-true: null=True should be set if blank=True is set on non-text fields.
65:    grouper_id = models.CharField(max_length=50, blank=True)
```

**Expected behavior**
Shouldn't fail

**Priority**
How important is this to you?
- [ ] P0: blocking me from making progress
- [ ] P1: this will block me in the near future
- [x] P2: annoying but not blocking me (I'll `#nosemgrep` for now :)

Version: 0.61.0

tino:
Oh btw, how does one disable this single rule? I mean there is no way to do `#nosemgrep python.django.correctness.nontext-field-must-set-null-true.nontext-field-must-set-null-true` without going over my 88 line-length...

tino:
I noticed this also fails for subclassed fields. For example with fields that serialize in some way (json, encryption, etc), this also fails:
```python
class EncryptedField(models.TextField):
    # implement en/decryption in .get_prep_value() & .from_db_value
   ...

class FakeModel(models.Model):
  # ok: nontext-field-must-set-null-true
  fieldText = EncryptedField(blank=True)
```

Can we update the rule so it is smart enough to deal with this?

tino:
Thanks @ievans. Anything I can do to make this more clear? Or what should I expect in terms of answers here?

minusworld:
> Oh btw, how does one disable this single rule? I mean there is no way to do `#nosemgrep python.django.correctness.nontext-field-must-set-null-true.nontext-field-must-set-null-true` without going over my 88 line-length...

[...]

PR #2086, opened 2022-05-16
Filter django.contrib.gis.db.models from results

minusworld:


minusworld:
Addresses https://github.com/returntocorp/semgrep-rules/issues/1421

colleend:
@minusworld I think it's just missing the 'rules': key at the top

p4p3r (approved):
LGTM

PR #2093, opened 2022-05-23
added debug rule + all cookie rules for laravel

colleend:
These rules were written for the sake of getting laravel up to tier 1 (specifically for the 'debug code' and 'cookie security' bullet points).

This follows https://cheatsheetseries.owasp.org/cheatsheets/Laravel_Cheat_Sheet.html for best recommendations. 

_To test:_
* run `semgrep --test .` on `php/laravel/security` and check that all tests pass.

_TODO:_
* add tests for default values (ex: catch ` config('app.debug', 'true');`)


php/laravel/security/laravel-active-debug-code.php
```diff
@@ -0,0 +1,23 @@
...
+
+    // ok: laravel-active-debug-code
+    putenv("APP_DEBUG=false");
+
+    // ok: laravel-active-debug-code 
+    $_ENV['APP_DEBUG'] = 'false';
+
+    // this is ok because it retrieves the value from the env file instead of setting it directly.
+    // ok: laravel-active-debug-code
+    $value = config('app.debug', 'true');
```
kurt-r2c, line 22: `true` is a default value here - should we still flag it? Failing to set it in the environment will cause this to "fail open" to debug behavior.

php/laravel/security/laravel-cookie-http-only.yaml
```diff
@@ -0,0 +1,41 @@
...
+  - pattern-not-inside: |
+      return [
+        ...,
+        'http_only' => true,
+        ...
+      ];
+  - pattern-not-inside: |
+      return [
+        ...,
+        'http_only' => env('$NAME', $DEFAULT),
```
kurt-r2c, line 21: Similar to debug code, this could potentially fail open to `false`

php/laravel/security/laravel-cookie-long-timeout.yaml
```diff
@@ -0,0 +1,35 @@
...
+  - pattern-inside: |
+      return [
+        ...,
+        'lifetime' => $TIME,
+        ...
+      ];
+  - pattern-not-inside: |
+      return [
+        ...,
+        'lifetime' => env("$VAR", $DEFAULT),
```
kurt-r2c, line 15: we should compare the default to the lifetime as well and flag defaults that are wrong

php/laravel/security/laravel-cookie-null-domain.yaml
```diff
@@ -0,0 +1,40 @@
...
+  - pattern-not-inside: |
+      return [
+        ...,
+        'domain' => null,
+        ...
+      ];
+  - pattern-not-inside: |
+      return [
+        ...,
+        'domain' => env('$NAME', $DEFAULT),
```
kurt-r2c, line 21: should check default for null as well

php/laravel/security/laravel-cookie-same-site.yaml
```diff
@@ -0,0 +1,45 @@
...
+  - pattern-not-inside: |
+      return [
+        ...,
+        'same_site' => 'strict',
+        ...
+      ];
+  - pattern-not-inside: |
+      return [
+        ...,
+        'same_site' => env('$NAME', $DEFAULT),
```
kurt-r2c, line 27: this may ignore bad defaults

php/laravel/security/laravel-cookie-secure-set.yaml
```diff
@@ -0,0 +1,40 @@
...
+  - pattern-not-inside: |
+      return [
+        ...,
+        'secure' => true,
+        ...
+      ];
+  - pattern-not-inside: |
+      return [
+        ...,
+        'secure' => env('$NAME', $DEFAULT),
```
kurt-r2c, line 21: this may ignore bad defaults

kurt-r2c (changes requested):
We should make sure we cover default values as well in `config($VAL,$DEFAULT)` and `env($VAL,$DEFAULT)`

php/laravel/security/laravel-active-debug-code.php
colleend: this is a good question! I opted to not flag because of the impression that we were leaning more towards reducing FP rather than reducing FN, and suspected that flagging this could cause alot of FP reports. This is definitely not good practice, though. What do you think?

colleend: weird, totally thought that I'd replied but perhaps I didn't click send 🤔 . this is definitely a good question -- I was thinking about this while writing the rule, and thought that we're currently prioritizing reducing FP rather than FN so opted to not catch unsafe defaults. Do you think we'll get a lot of negative sentiment from developers due to this?

kurt-r2c: This is a good point. The defaults is a "maybe problem" and perhaps we're okay with this?  We can ship this as-is and revisit if we get lots of "why didn't you find this" reports.

kurt-r2c (approved):

PR #2095, opened 2022-05-24
Reduce findings span - part 2

p4p3r:
The purpose of this PR is to reduce the number of lines of the findings reported by rules that have shown to be matching on average very large ranges.

Marked as draft while the impact is assessed.


python/django/security/audit/django-rest-framework/missing-throttle-config.yaml
```diff
@@ -8,14 +8,17 @@ rules:
           }
       - pattern: |
```
0xDC0DE: Should this be `pattern-inside` ?

0xDC0DE (approved):

python/django/security/audit/django-rest-framework/missing-throttle-config.yaml
p4p3r: The rules works either way.  Do you think `pattern-inside` would be more appropriate?

0xDC0DE: It feels counter-intuitive.
If the match needs to be `REST_FRAMEWORK = ...` and `REST_FRAMEWORK` at the same time, the clauses seems either conflicting (can not be both at the same time) or redundant (one is more restrictive than the other).

p4p3r: I see why `pattern-inside` would make more sense, as I'm using patterns to step-by-step drill down to a specific part "keyword".
Addressed in 72b013cb.

typescript/react/security/audit/react-styled-components-injection.jsx
```diff
@@ -2,28 +2,22 @@ import styled from "styled-components";
 
 function Vulnerable1(userInput) {
 // ruleid: react-styled-components-injection
```
0xDC0DE: I've also come across cases like this. I wonder if there is a better way to do this, instead of making the code a single line. Should we have smarter test syntax?

0xDC0DE (approved):

typescript/react/security/audit/react-styled-components-injection.jsx
p4p3r: Oh actually it works fine with comments inside the curly braces! Fixed in ab710c844482bbf3afa24c1ee22862bc356e3384

colleend:
are we waiting on someone before this PR is merge-able?

p4p3r:
> are we waiting on someone before this PR is merge-able?

@colleend yes, for both this and https://github.com/returntocorp/semgrep-rules/pull/2081 we are waiting on some work from the app team.

colleend:
got it, I'll note that in support queue 👍

PR #2101, opened 2022-05-26
Filter logger-style statements from Scala tainted-sql-string

minusworld:


scala/lang/security/audit/tainted-sql-string.yaml
```diff
@@ -1,68 +1,83 @@
...
+                }
+            - pattern-inside: |
+                def $CTRL(..., $PARAM: $TYPE, ...) = $A {
+                  ...
+                }
+            - pattern-inside: |
+                def $CTRL(..., $PARAM: $TYPE, ...) = $A(...) {
+                  ...
+                }
+  pattern-sinks:
```
kurt-r2c, line 42: This isn't really a "SQL injection" rule as much as it's a "string concatenation" rule with some additional keywords. As a fix, this is probably fine but I think we'll get better results by scrapping this approach and directly targeting Java/Scala ORMs as sinks.

kurt-r2c (approved):

PR #2108, opened 2022-05-30
Fix test expectations after returntocorp/semgrep#5348

IagoAbal:
Relates to PR returntocorp/semgrep#5348
fixing returntocorp/semgrep#5239 (PA-1308)


0xDC0DE (approved):
Looks good to me. Waiting for update to semgrep so that tests pass.

PR #2109, opened 2022-05-30
Change replacement for defaulthttpclient

coheigea:
The current defaulthttpclient rule recommends switching to a deprecated class https://hc.apache.org/httpcomponents-client-4.5.x/current/httpclient/apidocs/org/apache/http/impl/client/SystemDefaultHttpClient.html

Instead we should switch to HttpClientBuilder as per https://hc.apache.org/httpcomponents-client-4.5.x/current/httpclient/apidocs/org/apache/http/impl/client/DefaultHttpClient.html
"Deprecated. (4.3) use [HttpClientBuilder]"


0xDC0DE (approved):
Looks good to me! Thanks for your contribution.

PR #2117, opened 2022-06-07
Add support for HEALTHCHECK CMD ...

thomdixon:
This adds support for `HEALTHCHECK CMD` to the `generic.dockerfile.correctness.multiple-cmd-instructions` rule.


p4p3r:
Hello @thomdixon, first of all thanks for your contribution!

I think this rule can be improved to cover more cases.
For instance, currently the rule is going to report a false positive when running on the following snippet:
```
FROM couchbase:latest
COPY configure-node.sh /opt/couchbase
CMD ["/opt/couchbase/configure-node.sh"]
HEALTHCHECK --interval=5s --timeout=3s CMD curl --fail http://localhost:8091/pools || exit 1
```

Would you mind adding similar `pattern-not` to those you already added to cover all cases?

Thanks!

p4p3r:
Hello @thomdixon, first of all thanks for your contribution!

I think this rule can be improved to cover more cases.
For instance, currently the rule is going to report a false positive when running on the following snippet:
```
FROM couchbase:latest
COPY configure-node.sh /opt/couchbase
CMD ["/opt/couchbase/configure-node.sh"]
HEALTHCHECK --interval=5s --timeout=3s CMD curl --fail http://localhost:8091/pools || exit 1
```

Would you mind adding similar `pattern-not` to those you already added to cover all cases?

Thanks!

thomdixon:
> Hello @thomdixon, first of all thanks for your contribution!
> 
> I think this rule can be improved to cover more cases. For instance, currently the rule is going to report a false positive when running on the following snippet:
> 
> ```
> FROM couchbase:latest
> COPY configure-node.sh /opt/couchbase
> CMD ["/opt/couchbase/configure-node.sh"]
> HEALTHCHECK --interval=5s --timeout=3s CMD curl --fail http://localhost:8091/pools || exit 1
> ```
> 
> Would you mind adding similar `pattern-not` to those you already added to cover all cases?
> 
> Thanks!

Nice catch! Updated to support `HEALTHCHECK` parameters and other orderings, so your false-positive example should now pass. I'll look into the CLA and fixing any errors that show up in the tests.

p4p3r:
Tests are good. You can ignore `semgrep-rules-test-historical` for the time being, it's not behaving properly but the current output is actually ok (these test case failures are actually expected).
You just need to take care of the CLA and then we're good to merge!

p4p3r (approved):
LGTM

thomdixon:
Signed the CLA :) Thanks again for your patience!

PR #2121, opened 2022-06-08
Rule for using pytest snapshots after patching pathlib

mmcqd:
This adds a rule that will catch the use of `snapshot.assert_match` in a pytest test after patching the `Path` object. Doing this can cause unexpected behavior because `assert_match` using path operations to create snapshot files.

### DIFF
--- a/python/lang/correctness/pytest-assert_match-after-path-patch.py
+++ b/python/lang/correctness/pytest-assert_match-after-path-patch.py
@@ -1,4 +1,5 @@
 import pytest
+from pathlib import Path
 
 @pytest.mark.quick
 def test_foo(snapshot, mocker):
--- a/python/lang/correctness/pytest-assert_match-after-path-patch.py
+++ b/python/lang/correctness/pytest-assert_match-after-path-patch.py
@@ -6,6 +6,13 @@ def test_foo(snapshot, mocker):
     #ruleid: pytest-assert_match-after-path-patch
     snapshot.assert_match(foo(), "results.json")
 
+
+@pytest.mark.quick
+def test_fooooo(snapshot, mocker):
+    mocker.patch("pathlib.Path", None)
+    #ruleid: pytest-assert_match-after-path-patch
+    snapshot.assert_match(foo(), "results.json")
+
 @pytest.mark.quick
 def test_bar(snapshot, mocker):
     #ok: pytest-assert_match-after-path-patch
--- a/python/lang/correctness/pytest-assert_match-after-path-patch.py
+++ b/python/lang/correctness/pytest-assert_match-after-path-patch.py
@@ -1,0 +1,12 @@
+import pytest
+
+@pytest.mark.quick
+def test_foo(snapshot, mocker):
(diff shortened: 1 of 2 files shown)


python/lang/correctness/pytest-assert_match-after-path-patch.yaml
```diff
@@ -0,0 +1,19 @@
+rules:
+  - id: pytest-assert_match-after-path-patch
+    patterns:
+      - pattern-inside: |
+          import pytest
+          ...
+      - pattern-inside: |
+          mocker.patch.object(Path, $METHOD, $MOCKED_VALUE)
+          ...
```
underyx:
```suggestion
      - pattern-either:
        - pattern-inside: |
            mocker.patch("pathlib.Path", $MOCKED_VALUE)
            ...
        - pattern-inside: |
            mocker.patch.object(Path, $METHOD, $MOCKED_VALUE)
            ...
```

python/lang/correctness/pytest-assert_match-after-path-patch.py
```diff
@@ -0,0 +1,12 @@
+import pytest
+
+@pytest.mark.quick
+def test_foo(snapshot, mocker):
+    mocker.patch.object(Path, "open", mocker.mock_open(read_data=file_content))
+    #ruleid: pytest-assert_match-after-path-patch
+    snapshot.assert_match(foo(), "results.json")
```
underyx, line 8:
```suggestion
@pytest.mark.quick
def test_foo(snapshot, mocker):
    mocker.patch.object(Path, "open", mocker.mock_open(read_data=file_content))
    #ruleid: pytest-assert_match-after-path-patch
    snapshot.assert_match(foo(), "results.json")


@pytest.mark.quick
def test_fooooo(snapshot, mocker):
    mocker.patch("pathlib.Path", None)
    #ruleid: pytest-assert_match-after-path-patch
    snapshot.assert_match(foo(), "results.json")
```

python/lang/correctness/pytest-assert_match-after-path-patch.yaml
```diff
@@ -0,0 +1,23 @@
...
+        - pattern-inside: |
+            mocker.patch("pathlib.Path", $MOCKED_VALUE)
+            ...
+        - pattern-inside: |
+            mocker.patch.object(Path, $METHOD, $MOCKED_VALUE)
+            ...
+      - pattern:
+          snapshot.assert_match(...)
+    message: >-
+      snapshot.assert_match makes use of pathlib to create files. Patching $METHOD may result in unexpected snapshot behavior
```
kurt-r2c, line 17: `$METHOD` may not be set here - are we okay with this behavior for community adoption?  cc @minusworld @raghavjain3

minusworld: I'm usually OK with this. Best thing might just be to say "Patching $METHOD or pathlib.Path"

minusworld (approved):

PR #2122, opened 2022-06-09
Address feedback for prefer-copy-over-add

p4p3r:
We received feedback related to Dockerfile's `prefer-copy-over-add` rule, which is too noisy.
I updated the rule to focus on the most interesting cases of dangerous ADD usage: URLs and files that Docker will automatically extract.


kurt-r2c (approved):
lgtm

PR #2125, opened 2022-06-12
Update react-missing-noreferrer.yaml

hughdavenport:
This rule shouldn't complain when `noopener` is in place, but `noreferrer` is in place. The linked reference suggests both are suitable for a mitigation of the `window.opener` risk.

### DIFF
--- a/typescript/react/security/audit/react-missing-noreferrer.jsx
+++ b/typescript/react/security/audit/react-missing-noreferrer.jsx
@@ -1,10 +1,10 @@
 // ruleid: react-missing-noreferrer
 var Test1 = <a target='_blank' href="http://example.com/"></a>
 
-// ruleid: react-missing-noreferrer
+// ok: react-missing-noreferrer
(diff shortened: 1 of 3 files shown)


hughdavenport:
The `semgrep-rules-test-with-agent / rules-test-with-agent` job seems to be failing on things outside of this change.

minusworld:
Hey @hughdavenport, thanks! I took the liberty of switching `metavariable-regex` over to `metavariable-pattern` while you were making this change, which I think makes the rule easier to read. I also updated the test annotations to match the new behavior. Thanks again!

minusworld:
I'm not sure what's up with the failing agent test. I will investigate, but it looks to me like it's misbehaving in a strange way.

minusworld:
I'm not sure why that test is failing here and not on other PRs. I looked it over and I didn't see anything strange, and we've been talking about fixing up that test anyway, so I'm going to approve and merge. Thanks!

minusworld (approved):

PR #2126, opened 2022-06-13
update rule to use typed metavariables and public method taint

kurt-r2c:
@minusworld 
cc @ievans


minusworld (approved):
Tested on ~300 repos, returned 21 findings. Seems like an improvement

PR #2127, opened 2022-06-13
Added `InsecureSkipVerify` rule

enncoded:


minusworld:
@enncoded This looks like a more precise version of this rule, here: https://github.com/returntocorp/semgrep-rules/blob/7d2f26cfc8c2a79b4c72823bfc5c0de08eacac21/problem-based-packs/insecure-transport/go-stdlib/bypass-tls-verification.yaml

Is that right?

enncoded:
@minusworld Sorry for the late response. Yes, that's correct. I didn't know about that one my bad

LewisArdern:
@emjin will make a new PR updating old rule

PR #2128, opened 2022-06-14
deprecate react rules

LewisArdern:
Deprecating react rules with negative developer sentiment 

Reasons:
1. Rule offered 'hypothetical' type of vulnerabilities not actual vulnerabilities 
2. Rule is a best practice, and shouldn't have been an audit rule in the first place
3. We were flagging basic usage of react which was upsetting developers


minusworld:
`validate-r2c-registry-metadata` and `-historical` can be skipped. We'll need to update the tests not to match anything.

minusworld (approved):

PR #2130, opened 2022-06-14
Match `new` in grpc rule

nmote:
This rule previously matched `new ...` only by accident, because of a
parsing inconsistency between JS/TS and other languages. I am addressing
that inconsistency in https://github.com/returntocorp/semgrep/pull/5510.

This change updates the rule so that it will continue to function as
desired.


nmote:
I don't think the failures are related to the changes I made in this PR

minusworld (approved):

PR #2136, opened 2022-06-15
improve react audit rules

LewisArdern:
This introduces a still 'audit style' rule, but uses taint to try to help reduce the large number of false positives people have been seeing.

## Dangerously SetinnerHTML 
Reduction from dataset:
Initial rule: 1400 findings
Latest update: 400 findings

1. We still focus on any element with dangerouslySetInnerHTML and { } values
2. We now check for use of a pattern-sanitizer for the 3 common sanitizers iv'e seen used in react apps
3. We do a basic check for no longer usage of consts, which people will deem a false positive

Until deepsemgrep comes to JS I think this is the close to best audit rule we can create, maybe with some additional checks in the future such as { } object entering createElement

https://semgrep.dev/s/oRW1

## react-var-href
Use taint to allow 'anything'
Scope down from:

- We don't care about imports
- We don't care about hard coded strings
- We don't care about things which start start as a string
- We don't care about template literal strings which start with something other than a potentially tainted value
- Add a basic sanitizer to check to see if the value is within an if $value.startsWith

https://semgrep.dev/s/qBg8

## Lowering severity
MEDIUM TO LOW for many of the 'findings' which are very noisy which likely wont lead to a true exploit


LewisArdern:
Should also resolve https://github.com/returntocorp/semgrep-rules/issues/2168

inkz (approved):
stamp!

PR #2138, opened 2022-06-16
Add rule for unsafe usage of gorm methods

lfama:
In order to support some features, gorm doesn't escape user inputs when using some functions (as stated [here](https://gorm.io/docs/security.html#SQL-injection-Methods)). This rule tries to look for usage of those dangerous functions using taint mode (it's the first time I'm using taint mode so I hope I'm not missing something).


LewisArdern:
Hey @lfama thank you so much for your contribution!

I think we need to make the 'sources' a little bit more specific -- there are some things which wont be controllable on the request object from the user, do you approve of the updates to the following sources?

Example: https://semgrep.dev/s/dNDY

Cheers!

lfama:
Hi @LewisArdern,

We could make the sources more specific as you suggested, but in that case it would be hard to list all the possible methods that can be used to read the request parameters (for example your list does not include `ParseForm()`), leading to false negatives.
If you look at few rules already in the registry (like [this ](https://github.com/returntocorp/semgrep-rules/blob/develop/go/lang/security/injection/tainted-sql-string.yaml)and [this](https://github.com/returntocorp/semgrep-rules/blob/develop/go/lang/security/injection/tainted-url-host.yaml)) they are using the same approach. 
Furthermore, I believe that it's unlikely to produce false positives as we are looking for specific GORM methods name inside the GORM import statement.
What do you think?

Thanks!

LewisArdern:
We also need to update old rules too :) but that will be a separate PR. 

I have an internal thread going about using typed metavariables within the sink too e.g.
https://semgrep.dev/s/32zb but we can't currently do that with semgrep
 
but in general, yes the import is important but doesn't mean that there wont be another import that also has that method exposed something like this https://semgrep.dev/playground/s/returntocorp:gorm-dangerous-method-usage would probably be more accurate, but may miss findings too due to the more restrictive nature.

LewisArdern:
Hey @lfama your updates look great, i made some slight changes. Do you have anything else you would like to add or is it good to merge?

lfama:
@LewisArdern It looks good to me!

LewisArdern (approved):
lgtm

LewisArdern:
Thanks for your contribution!

PR #2148, opened 2022-06-20
Update sources for tainted Java spring rules

inkz:
extend scope of sources for Java Spring rules that target server-side vulns


LewisArdern (approved):
lgtm

PR #2152, opened 2022-06-21
Deprecate bracket-object-injection

minusworld:
This rule will be deprecated. It is based on eslint-plugin-security's detect-object-injection rule, but fires on extremely common patterns (approx. 800 times per project according to some scans) and is rarely, if ever, a security problem.

This rule will no longer match anything.


LewisArdern (approved):
lgtm

PR #2156, opened 2022-06-22
fix template-unquoted-attribute-var to have better develop sentiment

LewisArdern:
Re-wrote this rule to only use requests library from flask which should improve developer sentiment


inkz (commented):
maybe create 2 rules? one generic for html and second for python files, where it first finds a hardcoded string that is used as template and then it is scanned as generic with the same patterns used in the 1 rule? (https://semgrep.dev/docs/writing-rules/rule-syntax/#nested-language)

inkz (approved):
added `.py` test

PR #2157, opened 2022-06-22
Fixed sources for some Go rules

enncoded:


LewisArdern:
@enncoded do we wanna update gorm too?

https://github.com/returntocorp/semgrep-rules/blob/d367a61924f5ca6b7d07f89acc824d4888c67242/go/gorm/security/audit/gorm-dangerous-methods-usage.yaml

PR #2158, opened 2022-06-23
fix privileged service

LewisArdern:
Updated this rule to only match the line where the issue occurs, not the extra lines like previously


inkz (approved):
you can also use metavariable-pattern
```
      - metavariable-pattern:
          metavariable: $TRUE
          pattern: true
```

idk what will be more efficient, but the result is gonna be the same so, I approve! :+1:

PR #2159, opened 2022-06-23
Update Python dangerous exec rules to use taint mode

p4p3r:
This PR does the following:
- Removes current dangerous* audit rules for Python (covering various command injection and code injection sinks)
- Replaces them with two sets of taint-mode rules with respectively sources for web frameworks and env variables / cmd line args.
- The current rule ids are used by the rules targeting web frameworks (more critical). The others have similar ids but with an added suffix


python/lang/security/audit/dangerous-asyncio-create-exec-tainted-env-args.yaml
```diff
@@ -0,0 +1,98 @@
...
+  - id: dangerous-asyncio-create-exec-tainted-env-args
+    mode: taint
+    options:
+      symbolic_propagation: true
+    pattern-sources:
+      - patterns:
+          - pattern-either:
+              - patterns:
+                  - pattern-either:
+                      - pattern: os.environ['$ANYTHING']
```
kurt-r2c: this might work as just `os.environ` - are we using `$ANYTHING`?

python/lang/security/dangerous-code-run.yaml
```diff
@@ -0,0 +1,132 @@
+rules:
+  - id: dangerous-interactive-code-run
+    mode: taint
+    options:
+      symbolic_propagation: true
+    pattern-sources:
+      - patterns:
+          - pattern-either:
+              - pattern: flask.request.$ANYTHING
```
LewisArdern: This could be problematic you may want to scope down only addressable things from request e.g. ARGS. headers .body etc.

p4p3r: Fixed in 95b335ce1be919443c93844add4872a675e0be57

python/lang/security/audit/dangerous-asyncio-create-exec-tainted-env-args.yaml
```diff
@@ -0,0 +1,98 @@
...
+                          - pattern-either:
+                              - pattern: sys.argv[...]
+                              - pattern: sys.orig_argv[...]
+                      - patterns:
+                          - pattern-inside: |
+                              $PARSER = argparse.ArgumentParser(...)
+                              ...
+                          - pattern-inside: |
+                              $ARGS = $PARSER.parse_args()
+                          - pattern: <... $ARGS ...>
```
LewisArdern, line 29: should we use focs-metavariable here?

python/lang/security/audit/dangerous-asyncio-create-exec-tainted-env-args.yaml
```diff
@@ -0,0 +1,98 @@
...
+                                  ...
+                              - pattern-inside: |
+                                  $OPTS, $ARGS = getopt.gnu_getopt(...)
+                                  ...
+                          - pattern-either:
+                              - patterns:
+                                  - pattern-inside: |
+                                      for $O, $A in $OPTS:
+                                        ...
+                                  - pattern: $A
```
LewisArdern, line 50: same for this?

python/lang/security/audit/dangerous-asyncio-shell-tainted-env-args.yaml
```diff
@@ -0,0 +1,88 @@
...
+                                        ...
+                                  - pattern: $A
+                              - pattern: $ARGS
+    pattern-sinks:
+      - patterns:
+          - pattern-either:
+              - pattern-inside: $LOOP.subprocess_shell($PROTOCOL, $CMD)
+              - pattern-inside: asyncio.subprocess.create_subprocess_shell($CMD, ...)
+              - pattern-inside: asyncio.create_subprocess_shell($CMD, ...)
+          - pattern: $CMD
```
LewisArdern: same focus-metavariable?

LewisArdern (approved):
approved, feel free to update or not focus

LewisArdern (changes requested):
Sadly we can't change values from audit so we may need to re-do this PR... 
* Keep rules as audit with audit names
* New rules are welcome to stay where they are

p4p3r:
> Sadly we can't change values from audit so we may need to re-do this PR...
> 
> * Keep rules as audit with audit names
> * New rules are welcome to stay where they are

@LewisArdern you mean keeping the old rules but changing the ids by adding for example a `-audit` suffix and add all the new rules already in the PR as they are?

LewisArdern:
Yep, I think thats probably best cause of action, we can deprecate the deleted rules if you want to keep current structure https://github.com/returntocorp/semgrep-rules/commit/b0e7ba3c0dbd04c535eb8ba73aa59f3a2b997114 for inspiration..

python/lang/security/audit/dangerous-asyncio-create-exec-tainted-env-args.yaml
p4p3r: I'd rather keep this with a `pattern` like in the other part of the `pattern-either`

python/lang/security/audit/dangerous-asyncio-create-exec-tainted-env-args.yaml
p4p3r: I don't think in this case focus-metavariable would apply?

python/lang/security/audit/dangerous-asyncio-shell-tainted-env-args.yaml
p4p3r: Addressed in a7df6ca8bd66541db46740c0f5c62e6b8f5bc648

p4p3r:
> Yep, I think thats probably best cause of action, we can deprecate the deleted rules if you want to keep current structure [b0e7ba3](https://github.com/returntocorp/semgrep-rules/commit/b0e7ba3c0dbd04c535eb8ba73aa59f3a2b997114) for inspiration..

I added the old audit rules back with a `-audit` suffix. I'll update the rulepacks once this is merged.

python/lang/security/audit/dangerous-asyncio-exec-tainted-env-args.yaml
```diff
@@ -0,0 +1,86 @@
...
+        control_url: https://github.com/OWASP/ASVS/blob/master/4.0/en/0x13-V5-Validation-Sanitization-Encoding.md#v53-output-encoding-and-injection-prevention-requirements
+        version: "4"
+      references:
+        - https://docs.python.org/3/library/asyncio-eventloop.html#asyncio.loop.subprocess_exec
+        - https://docs.python.org/3/library/shlex.html
+        - https://semgrep.dev/docs/cheat-sheets/python-command-injection/
+      category: security
+      technology:
+        - python
+      confidence: HIGH
```
LewisArdern: I think people are going to argue about confidence of environment variables and OS commands as high confidence, because its usually 'defined by the consumer of the applicaation

LewisArdern (approved):

python/lang/security/audit/dangerous-asyncio-exec-tainted-env-args.yaml
kurt-r2c: I think Lewis is right, however since this is namespaced as an audit rule it's unsuitable for CI (which I also tend to agree with) - things like this should probably be forked and used w/additional app context.

p4p3r: Yeah, I wanted to convey the fact that confidence is high related to the taint flow being real/accurate (vs a simpler non-taint based rule) but I agree it can be misleading. I lowered the confidence for all audit rules in b775beb85fa56f2a688d0d36cdf81ca090bc3d9f

python/lang/security/audit/dangerous-asyncio-create-exec-tainted-env-args.yaml
p4p3r: Addressed in 42002a39d27f0d53b0c028b5f4834775864413bf

PR #2161, opened 2022-06-23
update open redirect rules

### LINKED ISSUE #2149: JS Open redirect rule to support other types of strings
**Describe the bug**
I may be wrong, but it seems that the open redirect rule is missing other ways of ensuring a string is passed.

**To Reproduce**
```js
const url = "/string-path"
window.location.replace(url)
```

**Expected behavior**
In the rule definition it seems that only a directly hard coded string with double quotes `""` is allowed, however, there are more ways to ensure a string, like TypeScript or string literals, casting or .toString()

Is there any other way to make to rule happy while also allowing to define the paths elsewhere? (needed for our react component library)

**Priority**
How important is this to you?

LewisArdern:
Resolves https://github.com/returntocorp/semgrep-rules/issues/2149 which had issues with hard-coded values capturing, rewrote it with client-side sources and taint mode.

### DIFF
--- a/javascript/browser/security/open-redirect.yaml
+++ b/javascript/browser/security/open-redirect.yaml
@@ -19,6 +19,8 @@ rules:
         version: "4"
       category: security
       confidence: HIGH
+      references:
+        - https://cheatsheetseries.owasp.org/cheatsheets/Unvalidated_Redirects_and_Forwards_Cheat_Sheet.html
       technology:
         - browser
       license: Commons Clause License Condition v1.0[LGPL-2.1-only]
--- a/javascript/browser/security/open-redirect.yaml
+++ b/javascript/browser/security/open-redirect.yaml
@@ -56,8 +56,7 @@ pattern-sources:
               $PROPS = new URLSearchParams(location.search)
                 ...
           - pattern-inside: |
-              $PROPS = new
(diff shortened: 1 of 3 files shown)


javascript/browser/security/open-redirect.yaml
```diff
@@ -1,37 +1,86 @@
...
+                  $PROP = new URLSearchParams($WINDOW. ...
+                  .location.search).get('...')
+                   ...
+              - pattern-inside: |
+                  $PROP = new URLSearchParams(location.search).get('...')
+                   ...
+              - pattern-inside: >
+                  $PROP = new
+
+                  URLSearchParams($WINDOW. ... .location.hash.substring(1)).get('...')
```
inkz: why on different lines? :grimacing:

inkz (changes requested):
if you control first argument (`$ARG`) in `new URL($ARG, $BASE)` it is also an open redirect
e.g.
```javascript
(new URL('//bar.com', 'http://www.foo.ru')).toString()
// returns http://bar.com/
```

javascript/browser/security/open-redirect.yaml
LewisArdern: 🤷

PR #2164, opened 2022-06-23
fix detect-eval-with-expression

LewisArdern:
Rewritten with client-side sources for use of dynamic execution of javascript


inkz (approved):
amazing

PR #2169, opened 2022-06-27
Add audit-rules-must-be-info meta rule

underyx:


javascript/express/security/audit/express-cookie-settings.yaml
```diff
@@ -4,7 +4,7 @@ rules:
       Don’t use the default session cookie name
       Using the default session cookie name can open your app to attacks.
       The security issue posed is similar to X-Powered-By: a potential attacker can use it to fingerprint the server and target attacks accordingly.
-    severity: WARNING
+    severity: INFONG
```
underyx, line 7: infong?

underyx: image

colleend:
I see that there's discussion about whether we should change all audit rules to INFO, and the consensus among security researchers is that we should hold off on this. Should I close this PR @underyx @p4p3r?

underyx:
yup

PR #2171, opened 2022-06-28
Open Redirect Taint Update

### LINKED ISSUE #2149: JS Open redirect rule to support other types of strings

LewisArdern:
Resolves https://github.com/returntocorp/semgrep-rules/issues/2149 which had issues with hard-coded values capturing, rewrote it with client-side sources and taint mode.

### DIFF
--- a/javascript/browser/security/open-redirect-from-function.yaml
+++ b/javascript/browser/security/open-redirect-from-function.yaml
@@ -7,6 +7,7 @@ rules:
       (XSS)  with JavaScript URIs. It is recommended to validate
       user-controllable  input before allowing it to control the redirection.
     metadata:
+      confidence: LOW
       cwe: "CWE-601: URL Redirection to Untrusted Site ('Open Redirect')"
       owasp:
         - A10:2013 - Unvalidated Redirects and Forwards
@@ -26,7 +27,7 @@ rules:
     languages:
       - javascript
       - typescript
-    severity: WARNING
+    severity: INFO
     mode: taint
     pattern-sources:
       - patterns:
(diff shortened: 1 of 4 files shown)


javascript/browser/security/open-redirect-from-function.yaml
```diff
@@ -0,0 +1,57 @@
...
+      - typescript
+    severity: WARNING
+    mode: taint
+    pattern-sources:
+      - patterns:
+          - pattern-either:
+              - patterns:
+                  - pattern: |
+                      function ... (..., $PROP, ...) { ... }
+                  - pattern: $PROP
```
enncoded: Wouldn't this source be really noisy? There's no validation to narrow down the context and there isn't a specific field that's specified from the parameter, so right now any parameter in any function is a taint source.

enncoded (changes requested):

javascript/browser/security/open-redirect-from-function.yaml
enncoded: Resolved by lowering the `confidence` and `severity` in the rule

enncoded (approved):

PR #2172, opened 2022-06-28
Autofix tests for all rules with autofixes

0xDC0DE:
Sorry for the big PR!


javascript/angular/security/detect-angular-trust-as-js-method.yaml
```diff
@@ -2,9 +2,13 @@ rules:
...
     patterns:
       - pattern-either:
-          - pattern: |
-              $SOURCE = $scope.$INPUT;
-              $sce.trustAsJs($SOURCE);
+          - patterns:
+            - pattern-inside: |
+                $SOURCE = $scope.$INPUT;
+                ...
+                $sce.trustAsJs($SOURCE);
```
enncoded: I think you wanna remove this line here, since you have a pattern right after that's this exact line

also in:
* `javascript/angular/security/detect-angular-trust-as-resourceurl-method.yaml`
* `javascript/angular/security/detect-angular-trust-as-url-method.yaml`

enncoded (changes requested):

javascript/angular/security/detect-angular-trust-as-js-method.yaml
0xDC0DE: These actually should probably be taint rules at some point in the future. 
Having only 
```
 - pattern: | 
     $sce.trustAsJs($INPUT);
```
does not verify if this input is originating as user input. i also changed to rule to use a combination of `pattern-inside` and `pattern` so that I could rewrite the right part with the fix. But the fixes of these angular rules are going to be removed anyway, because @LewisArdern agreed that the current fix is probably not a good idea anyway. So I am going to revert these rules to their previous state.

generic/dockerfile/best-practice/use-absolute-workdir.yaml
```diff
@@ -1,12 +1,31 @@
...
+            metavariable: $CMD
+            pattern: WORKDIR
+        - focus-metavariable: $CMD
     message: >-
       Detected a relative WORKDIR. Use absolute paths. This prevents
       issues based on assumptions about the WORKDIR of previous containers.
     severity: WARNING
     languages: [generic]
-    fix: WORKDIR /$WORD
+    #fix: WORKDIR /$VALUE
```
colleend, line 28: is this commented because the fix isn't effective at this time?

python/requests/best-practice/use-response-json-shortcut.fixed.py
```diff
@@ -0,0 +1,10 @@
+import json
+import requests
+
+r = requests.get("https://example.org")
+
+# ok
```
colleend: should this be`ok:python.requests.best-practice.use-response-json-shortcut`?

colleend:
left small small nits, but looks ready to merge after those are changed!

generic/dockerfile/best-practice/use-absolute-workdir.yaml
0xDC0DE: That is correct, there are cases that do not work, and I can't write an effective fix for now. I think because of limitations of the generic parser that is not enough context-aware. I will take a stab at porting this rule to dockerfile parser at a later time.

typescript/react/security/audit/react-http-leak.yaml
```diff
@@ -6,11 +6,13 @@ rules:
...
     metadata:
-      owasp: "A3: Sensitive Data Exposure"
+      owasp: 
+        - "A06:2007 - Information Leakage and Improper Error Handling"
+        - "A01:2021 - Broken Access Control"
       cwe: "CWE-200: Exposure of Sensitive Information to an Unauthorized Actor"
       deprecated: true
       references:
-        - https://github.com/cure53/HTTPLeaks
+        - http"s://github.com/cure53/HTTPLeaks
```
colleend: i think this shouldn't have the extra double quotes

colleend (approved):
small nit, otherwise looks good to me!

colleend:
@enncoded would you mind approving too?

typescript/react/security/audit/react-http-leak.yaml
0xDC0DE: Oh! Good eye! Fixed.

enncoded (approved):

inkz:
@pieterdc1 setting low priority label, so that stale bot stopped disturbing :)

generic/dockerfile/best-practice/use-absolute-workdir.yaml
kurt-r2c, line 28: intentionally commented-out?

kurt-r2c (approved):

generic/dockerfile/best-practice/use-absolute-workdir.yaml
0xDC0DE: Yes, couldn't get the fix to work, so I removed it.

PR #2179, opened 2022-06-29
remove new-func

LewisArdern:
A lot of false positives from this, with not much benefit for developers.


enncoded (approved):
Approving despite metadata check failing because this rule is built not to fire

PR #2186, opened 2022-07-01
feat: Remove dockerfile rules which prevent to upgrade packages

SIPR-octo:
This PR removes a bunch a rules for Dockerfiles which are basically saying the same thing: "Do not upgrade your packages".

These rules, which most probably come from Hadolint, were already removed from a couple of linters (including Hadolint) and from the OWASP recommendations, for the reasons below.

**TL;DR**: 

> "Packages in base images should be up-to-date"

That's unfortunately **very rarely** the case.

> "If packages are out-of-date, consider contacting the base image maintainer"

That sounds good and the sensible thing to do, but it does not always work, and when it does it is not fast.

**Long read**: https://pythonspeed.com/articles/security-updates-in-docker/

Thanks!


colleend:
Thank you for submitting this PR! I read the article and the issue submitted in `hadolint`'s repository, and this sounds like the right thing to do (the article also resolved some of my doubts). We appreciate the time you've taken to help educate and also to submit a PR. 

Unfortunately, our ecosystem is a little tangled in the sense that when we delete rules from the `semgrep-rules` repository, customers can still use them. We're still working on how to truly 'delete' rules from our rules registry, but this will take a quarter or two. Instead, as a faster fix, I submitted a PR to delete these rules from our rulepacks so that the rules are less visible and so that most users will not be running them. 

Again, thanks for your time -- I'm going to close this PR and instead focus on merging the one to remove the rules from our rulepacks. Feel free to let us know if there's anything else on your mind!

PR #2187, opened 2022-07-01
fix: Fix false positives in rule hugo/invalid-base-url

SIPR-octo:
Found a false positive with the rule `invalid-base-url` in `generic.hugo`, which does not seem to work as expected.

Added some tests that the rule should not flag on, as specified in `CONTRIBUTING.md`, then fixed the rule!


enncoded (approved):

enncoded:
Thanks for the contribution!

PR #2188, opened 2022-07-01
Add missed cases of dangerous system calls

cikay:


python/lang/security/audit/dangerous-system-call.py
```diff
@@ -11,6 +11,30 @@
...
+
+# ruleid:dangerous-system-call
+getattr(__import__("os"), "system")(something())
+
+# ruleid:dangerous-system-call
+dynamic_os = __import__("os")
+dynamic_os.system(something())
+
+# ruleid:dangerous-system-call
+getattr(dynamic_os, "system")(something())
```
colleend, line 33: the develop test isn't passing because its matching starting from line 28 -- to fix this, you can add `dynamic_os = __import__("os")` to the line above line 32.

python/lang/security/audit/dangerous-system-call.py
```diff
@@ -11,6 +11,30 @@
...
+
+# ruleid:dangerous-system-call
+dynamic_system = getattr(dynamic_os, "system")
```
colleend, line 37: same thing here!

colleend (changes requested):
Thanks for your contribution! We appreciate it -- currently tests aren't passing because of some line-matching issues. To fix this, you can use something like this: https://semgrep.dev/s/JOwQ.

python/lang/security/audit/dangerous-system-call.py
cikay: Done

python/lang/security/audit/dangerous-system-call.py
cikay: Done

colleend (approved):
awesome, thank you!

PR #2195, opened 2022-07-04
dockerfile/use-workdir: improved match and cleaner message

p4p3r:
Improved both message and match to address developer sentiment.
It requires Semgrep > 0.102.0 (unreleased at the moment).

### DIFF
--- a/dockerfile/best-practice/use-workdir.dockerfile
+++ b/dockerfile/best-practice/use-workdir.dockerfile
@@ -3,7 +3,7 @@ FROM busybox
 # ruleid: use-workdir
 RUN cd semgrep && git clone https://github.com/returntocorp/semgrep
 
-# ruleid: use-workdir
+# ok: use-workdir
 RUN pip3 install semgrep && cd ..
 
 # ok: use-workdir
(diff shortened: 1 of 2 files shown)


colleend (approved):
lgtm :D

p4p3r:
Note: historical semgrep test (v0.100.0) is expected to be failing like this because this PR requires >v0.102.0

colleend:
ah I see -- v103 was released yesterday by nathan! I would be down to merge this and just not put it in any rulesets for now 😅 

it would really be great if rules came with corresponding min. version numbers, but that's a whole 'nother topic 😅

p4p3r:
> I would be down to merge this and just not put it in any rulesets for now 😅

That's seems fine to me. The user that submitted feedback initially should still benefit from the updated rule as soon as they will update to 0.103.

colleend:
ok awesome, that sounds good 👍

colleend:
ah shoot, I just realized that this rule is already in the `dockerfile` rule-pack, and pushing this out means that people who don't have upgraded semgrep versions and who try to run the `dockerfile` pack will encounter failures. Should've checked this before merging, my bad. 

I do remember we agreed with the CLI team to not release rules with new features until after waiting ~2 weeks in order to make sure people have time to upgrade -- is this still true? @minusworld

minusworld:
If it passes the historical tests, then we shouldn't have to worry about it. If it fails, then we should wait until it passes (I think ~3 weeks) @colleend @p4p3r

PR #2197, opened 2022-07-04
Make test historical check only changed files

p4p3r:
Make test historical check only changed files.
Also fix semgrep version check which broke after 0.100.0


colleend (approved):
thank you for doing this 🙏 🙏 this is awesome!!

Issue #2198, opened 2022-07-05
javascript.lang.security.detect-child-process.detect-child-process generates false positives


ianmuscat:
#### Describe the bug
The `javascript.lang.security.detect-child-process.detect-child-process` rule incorrectly reports usage of `spawn()` (and possibly other methods) that are not related to `child_process` as vulnerabilities.

#### To Reproduce
The following code was taken from the [XState documentation](https://xstate.js.org/docs/guides/actors.html#spawning-machines).
```
const parentMachine = createMachine({
  id: 'parent',
  initial: 'waiting',
  context: {
    localOne: null
  },
  states: {
    waiting: {
      entry: assign({
        localOne: () => spawn(remoteMachine)  // <--
      }),
      on: {
        'LOCAL.WAKE': {
          actions: send({ type: 'WAKE' }, { to: (context) => context.localOne })
        },
        'REMOTE.ONLINE': { target: 'connected' }
      }
    },
    connected: {}
  }
});
```

#### Expected behaviour
This rule should not evaluate to true since this code does not use `child_process`.

#### Priority
How important is this to you?
- P1: this will block me in the near future

colleend:
closed with #2199. Let us know if you have any other feedback in the future!

PR #2199, opened 2022-07-05
editing detect child process

colleend:
for https://github.com/returntocorp/semgrep-rules/issues/2198. This will make sure the calls we're flagging are actually from `child_process`. 

_Test plan_
Run `semgrep --test .`


inkz (approved):
@colleend  we now have import equivalences for JS? :open_mouth:

colleend:
yess!!! I was shocked too 😄

PR #2206, opened 2022-07-07
edited username and password rule to be less FP heavy

colleend:
This is for RULES-952 but also because we have had multiple complaints that this rule is very FP-heavy.

What I did: 
* changed the rule from regex rule to a generic rule
* made the 'PROTOCOL' portion match only `http, sql, ftp, smtp` URLS. I wanted to be specific about the protocol to eliminate FPs.
* defined 'usernames' to start with an alphabetical letter, and have between 5-31 other alphanumeric, underscore, or dash characters (more rigid than our previous definition). Usernames can also optionally be between brackets (var names essentially)
* defined 'passwords' to must include 1 number, 1 uppercase letter, one special character (as defined by OWASP). These 'passwords' must also be between 6-32 chars. This will make it so we're more likely to actually capture a password. 
* I did indeed try to use `metavariable-entropy`, but it didn't match on any of my examples so we are back to regexes :D 

_To test:_
run `semgrep -f detected-username-and-password-in-uri.yaml detected-username-and-password-in-uri.txt`

_Note:_
* this has not been tested on OSS repos.

### DIFF
--- a/generic/secrets/security/detected-username-and-password-in-uri.txt
+++ b/generic/secrets/security/detected-username-and-password-in-uri.txt
@@ -1,8 +1,8 @@
 # ruleid: detected-username-and-password-in-uri
-https://username:password@example.com
+https://username:passworD123*@example.com
 
 # ruleid: detected-username-and-password-in-uri
-https://username:**sswor*@example.com
+https://username:**sswor*D12@example.com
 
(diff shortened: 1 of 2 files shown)


inkz (approved):
LGTM! :+1:

PR #2221, opened 2022-07-13
Do not include whitespace in passwords (generic.secrets.security.detected-username-and-password-in-uri)

thomdixon:
# What
This performs a negative lookahead in the `$...PASSWORD` metavariable in order to ensure that it does not contain whitespace. A test is included that was previously considered a username/password combination.

# Why
The current check results in too many false positives where URLs and code examples/Markdown mix (e.g., YAML, Bazel build files, etc).


inkz (approved):
lgtm! Thanks a lot :pray:

PR #2222, opened 2022-07-13
fix url credential + tests

kurt-r2c:
Missing line boundary on user regex, causing multiline things to be captured as a "user" for things like lockfile definitions `package@version`


colleend (approved):
nice! thanks for adding in the beginning/ending enforcers, totally forgot about them 😅

kurt-r2c:
pretty sure I reviewed the PR that made this changes and I missed it too. Happens, and now we have better test cases

PR #2223, opened 2022-07-13
Update semgrep validation rule

kopecs:
`extract` mode rules (https://github.com/returntocorp/semgrep/pull/5621) do not require a message field.

We currently use this rule for the `--validate` flag in the CLI, so we can run the rule just fine, but it'll fail validation. Moreover, I'm not sure we need a rule for this in the first place---we'll catch this just fine with jsonschema in the CLI regardless, so unless we use this elsewhere we can just delete this rule instead.


inkz (approved):
lgtm

PR #2233, opened 2022-07-14
scope regex DOS to ActiveRecord-like behavior and cookie/params

kurt-r2c:
@raghavjain3 fix for Brakeman ReDoS rule.


inkz (approved):
:+1:

PR #2234, opened 2022-07-14
add "pipenv install --deploy" as allowed pattern

camjay:
add "pipenv install --deploy" as an allowed pattern, as this will result in a failure if the lockfile does not match pipfile.

This ensures deterministic deploys in the case of a match, or allows for an alert to update the Pipenv or lockfile and evaluate changes.


kurt-r2c:
see https://pipenv-fork.readthedocs.io/en/latest/advanced.html#using-pipenv-for-deployments for background info

LewisArdern:
@camjay if you sign the CLA we can merge this into develop!

camjay:
I was waiting for employer approval to complete the CLA, but it looks like https://github.com/returntocorp/semgrep-rules/commit/56ca767dd28fd898ac2445aabf42fb8730feeac0 has deprecated this rule and so this PR is no longer needed.

PR #2236, opened 2022-07-15
More rails brakeman rules

colleend:
Rules for: 
* check_model_serialize.rb https://github.com/advisories/GHSA-fhj9-cjjh-27vm
* check_simple_format.rb https://github.com/advisories/GHSA-w37c-q653-qg95

### DIFF
--- a/ruby/rails/security/brakeman/check-simple-format.yaml
+++ b/ruby/rails/security/brakeman/check-simple-format.yaml
@@ -1,0 +1,45 @@
+rules:
+- id: check-simple-format 
+  mode: taint
+  pattern-sources:
+  - pattern-either:
+    - pattern: |
+        cookies[...]
+    - patterns:
+      - pattern: |
+          cookies. ... .$PROPERTY[...]
+      - metavariable-regex:
+          metavariable: $PROPERTY
+          regex: (?!signed|encrypted)
+    - pattern: |
+        params[...]
+    - pattern: |
+        request.env[...]
+  pattern-sanitizers:
+  - pattern-either:
+    - pattern: h(...)
+    - pattern: html_escape(...)
+    - pattern: html_escape_once(...)
+  pattern-sinks:
+  - pattern: simple_format(...)
+  message: >-
+    Discovered usage of the `simple_format` helper, which
+    is vulnerable to XSS in versions 4.0.0 & 4.0.1. This
+    can lead to cookies and other secrets being obtained by
+    attackers. Instead, escape user data before passing it to
+    simple_format or upgrade.
+  languages:
+  - ruby
+  severity: ERROR
+  metadata:
+    source-rule-url: https://github.com/presidentbeef/brakeman/blob/main/lib/brakeman/checks/check_simple_format.rb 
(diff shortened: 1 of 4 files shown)


ruby/rails/security/brakeman/model-serialize.yaml
```diff
@@ -0,0 +1,30 @@
+rules:
+- id: model-serialize
```
kurt-r2c, line 2: this is effectively a duplicate of an SCA rule - @adamberman how do we want to handle this sort of coverage in Brakeman ports?

ruby/rails/security/brakeman/check-simple-format.yaml
```diff
@@ -0,0 +1,45 @@
...
+  pattern-sanitizers:
+  - pattern-either:
+    - pattern: h(...)
+    - pattern: html_escape(...)
+    - pattern: html_escape_once(...)
+  pattern-sinks:
+  - pattern: simple_format(...)
+  message: >-
+    Discovered usage of the `simple_format` helper, which
+    is vulnerable to XSS in versions 4.0.0 & 4.0.1. This
```
kurt-r2c, line 27: This should probably be an SCA rule

colleend: should i move these rules to the SCA repo then? Not sure what our current distinction between what should be in SCA or brakeman is 🤔 (especially since writing/running tests for SCA is a little difficult)

ruby/rails/security/brakeman/model-serialize.yaml
colleend: I looked in the folders of the SCA repo and didn't see this one 😅 perhaps it was named something different?

kurt-r2c: nm, this is the *first* version of messing up DB serialization.  Everything old is new again, as usual.

ruby/rails/security/brakeman/check-simple-format.yaml
kurt-r2c: This is probably OK, it would be higher accuracy with a project-depends-on key but this is probably okay as an audit rule.

Without the rails version context though, this will be noisy as `simple_format` is not a problem unless you are using a vulnerable rails version.

ruby/rails/security/brakeman/model-serialize.yaml
colleend: ah I see, gotcha

ruby/rails/security/brakeman/check-simple-format.yaml
colleend: hmm that's a very good point -- I'm going to move this to the SCA folder then (and will move the model-serialize rule too 😄). It makes sense to me that brakeman rules go in there (especially the ones with known CVEs).

colleend:
@kurt-r2c I'm going to move these rules to SCA (along with another rule I wrote) -- most of these have known CVEs so it Makes Sense (TM)

PR #2246, opened 2022-07-20
Update angular rule

LewisArdern:
Update bypassSecurityTrust rules in Angular to:

- Use taint mode (generic) $X
- Remove false positives "..." and template literal strings
- Update to use regex in sink (faster)
- Use default sanitizers which should improve developer experience


kurt-r2c (approved):
lgtm

PR #2248, opened 2022-07-21
update sql rules

LewisArdern:
* Improved taint for express using typescript and js sources
* Updated the source to come from a function argument for SQLI since thats pretty much what the rule is doing the other rule would flag any type of issue which would add negative develop sentiment

### DIFF
--- a/javascript/express/security/injection/tainted-sql-string.yaml
+++ b/javascript/express/security/injection/tainted-sql-string.yaml
@@ -1,7 +1,8 @@
 rules:
   - id: tainted-sql-string
     severity: ERROR
-    message: Detected user input used to manually construct a SQL string. This is
+    message: >-
+      Detected user input used to manually construct a SQL string. This is
       usually bad practice because manual construction could accidentally result
       in a SQL injection. An attacker could use a SQL injection to steal or
       modify contents of the database. Instead, use a parameterized query which
--- a/javascript/express/security/injection/tainted-sql-string.yaml
+++ b/javascript/express/security/injection/tainted-sql-string.yaml
@@ -1,13 +1,13 @@
 rules:
   - id: tainted-sql-string
     severity: ERROR
-    message: >-
-      Detected user input used to manually construct a SQL string. This is usually
-      bad practice because manual construction could accidentally result in a SQL
-      injection. An attacker could use a SQL injection to steal or modify contents
-      of the database. Instead, use a parameterized query which is available
-      by default in most database engines. Alternatively, consider using an
-      object-relational mapper (ORM) such as Sequelize which will protect your queries.
+    message: Detected user input used to manually construct a SQL string. This is
+      usually bad practice because manual construction could accidentally result
+      in a SQL injection. An attacker could use a SQL injection to steal or
+      modify contents of the database. Instead, use a parameterized query which
+      is available by default in most database engines. Alternatively, consider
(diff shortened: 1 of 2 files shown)


inkz (approved):
good!

javascript/express/security/injection/tainted-sql-string.yaml
```diff
@@ -20,34 +21,56 @@ rules:
...
+              - pattern: $REQ.headers
+      - patterns:
+          - pattern-either:
+              - pattern-inside: >
+                  ({ $REQ }: Request,$RES: Response, $NEXT: NextFunction) =>
+                  {...}
+              - pattern-inside: |
+                  ({ $REQ }: Request,$RES: Response) => {...}
+          - pattern-either:
+              - pattern: params
```
IagoAbal, line 53: Don't you want to make sure that `params` etc are values of `$REQ` ?

javascript/express/security/injection/tainted-sql-string.yaml
```diff
@@ -20,34 +21,56 @@ rules:
...
                   - pattern-either:
-                      - pattern: |
+                      - pattern-inside: |
                           "$SQLSTR" + $EXPR
-                      - pattern: |
-                          "$SQLSTR".concat(...)
-                      - pattern: util.format($SQLSTR, ...)
+                      - pattern-inside: |
+                          "$SQLSTR".concat($EXPR)
+                      - pattern: util.format($SQLSTR, $EXPR)
```
IagoAbal, line 67: It's strange to mix `pattern-inside` and `pattern` in a `pattern-either`. Is that what you want?

javascript/express/security/injection/tainted-sql-string.yaml
```diff
@@ -20,34 +21,56 @@ rules:
...
                   - metavariable-regex:
                       metavariable: $SQLSTR
                       regex: .*\b(?i)(select|delete|insert|create|update|alter|drop)\b.*
               - patterns:
                   - pattern: |
-                      `...${...}...`
+                      `...${$EXPR}...`
                   - pattern-regex: |
                       .*\b(?i)(select|delete|insert|create|update|alter|drop)\b.*
```
IagoAbal, line 75: Is this regex meant to only apply inside
```
`...${$EXPR}...`
```
?

javascript/lang/security/audit/sqli/node-mysql-sqli.yaml
```diff
@@ -1,66 +1,54 @@
...
+      technology:
+        - mysql
+        - mysql2
+        - javascript
+        - nodejs
+      license: Commons Clause License Condition v1.0[LGPL-2.1-only]
+    pattern-sources:
+      - patterns:
+          - pattern-inside: function ... (..., $Y,...) {...}
+          - pattern: $Y
```
IagoAbal, line 32: `focus-metavariable` is better here

javascript/express/security/injection/tainted-sql-string.yaml
LewisArdern: No, its meant to be most likely at the start of the expression, but the 'source should be $EXPR

javascript/lang/security/audit/sqli/node-mysql-sqli.yaml
LewisArdern: noted! i will update

PR #2249, opened 2022-07-21
JS/TS path traversal rules

inkz:


javascript/lang/security/audit/path-traversal/path-join-resolve-traversal.yaml
```diff
@@ -1,53 +1,68 @@
...
+  pattern-sources:
+  - patterns:
+    - pattern: $X
+    - pattern-either:
+      - pattern-inside: |
+          function ... (...,$X,...) {...}
+      - pattern-inside: |
+          function ... (...,{...,$X,...},...) {...}
+      - pattern-inside: |
+          for ($X of $SMTH) {...}
```
kurt-r2c, line 13: how are we feeling about the FP rate of this? Curious if this iterator pattern is disproportionately used for path elements when it's not used on an intrafunction array.

kurt-r2c (approved):

javascript/lang/security/audit/path-traversal/path-join-resolve-traversal.yaml
inkz: yeah, its a good call! if some array config file is used it will be highlighted, I will update the rule :+1:

PR #2250, opened 2022-07-21
improvements to SQL rules

LewisArdern:
* Updated express taint to use focus-metavariable on $REQ
* Reduced regex since it was not needed and could be used in earlier pattern

### DIFF
--- a/javascript/express/security/injection/tainted-sql-string.yaml
+++ b/javascript/express/security/injection/tainted-sql-string.yaml
@@ -36,7 +36,6 @@ rules:
                   - metavariable-regex:
                       metavariable: $METHOD
                       regex: ^(get|post|put|head|delete|options)
-          - focus-metavariable: $REQ
           - pattern-either:
               - pattern: $REQ.query
               - pattern: $REQ.body
@@ -73,4 +72,3 @@ rules:
                       metavariable: $SQLSTR
                       regex: .*\b(?i)(select|delete|insert|create|update|alter|drop)\b.*
           - focus-metavariable: $EXPR
-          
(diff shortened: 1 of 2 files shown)


javascript/express/security/injection/tainted-sql-string.yaml
```diff
@@ -36,6 +36,7 @@ rules:
                   - metavariable-regex:
                       metavariable: $METHOD
                       regex: ^(get|post|put|head|delete|options)
+          - focus-metavariable: $REQ
```
IagoAbal: Hmm, in this case this won't have the effect you want, you're kind of saying that `$REQ` is tainted rather than `$REQ.query`. So let's say that the code uses `req.query`, and later on you use `req.abc`, I think adding this `focus-metavariable` would cause `req.abc` to be tainted because you're specifying that `req` itself is tainted. In short, I would drop this line.

javascript/express/security/injection/tainted-sql-string.yaml
```diff
@@ -44,11 +45,12 @@ rules:
               - pattern: $REQ.headers
       - patterns:
           - pattern-either:
-              - pattern-inside: >
+              - pattern-inside: |
                   ({ $REQ }: Request,$RES: Response, $NEXT: NextFunction) =>
                   {...}
               - pattern-inside: |
                   ({ $REQ }: Request,$RES: Response) => {...}
+          - focus-metavariable: $REQ
```
IagoAbal, line 52: Here this is what you want! Because you want `params` etc to refer to values of `$REQ`.

javascript/express/security/injection/tainted-sql-string.yaml
```diff
@@ -65,12 +66,9 @@ rules:
                       - pattern-inside: |
                           "$SQLSTR".concat($EXPR)
                       - pattern: util.format($SQLSTR, $EXPR)
+                      - pattern: |
+                          `$SQLSTR${$EXPR}...`
```
kurt-r2c, line 70: v nice

kurt-r2c (approved):

PR #2262, opened 2022-07-25
Initial Express Source Cleanup

LewisArdern:
Updated rules to use latest express source/sinks

### DIFF
--- a/javascript/express/security/audit/remote-property-injection.yaml
+++ b/javascript/express/security/audit/remote-property-injection.yaml
@@ -27,18 +27,30 @@ rules:
               - pattern-inside: function ... ($REQ, $RES) {...}
               - pattern-inside: function ... ($REQ, $RES, $NEXT) {...}
               - patterns:
-                  - pattern-either:
-                      - pattern-inside: $APP.$METHOD(..., function $FUNC($REQ, $RES) {...})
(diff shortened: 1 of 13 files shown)


javascript/express/security/audit/remote-property-injection.yaml
```diff
@@ -27,18 +27,30 @@ rules:
               - pattern-inside: function ... ($REQ, $RES) {...}
               - pattern-inside: function ... ($REQ, $RES, $NEXT) {...}
               - patterns:
-                  - pattern-either:
-                      - pattern-inside: $APP.$METHOD(..., function $FUNC($REQ, $RES) {...})
-                      - pattern-inside: $APP.$METHOD(..., function $FUNC($REQ, $RES, $NEXT) {...})
```
0xDC0DE, line 32: You have removed the case with`$NEXT` here, but have left it in in the other rules. Is that on purpose?

0xDC0DE (approved):
Looks good to me!

javascript/express/security/audit/remote-property-injection.yaml
LewisArdern: i think the pattern is the same i just somehow messed up the ordering

PR #2265, opened 2022-07-26
update `path-join-resolve-traversal` rule

inkz:
Small fix to remove "for" loop from sources, discussed here: https://github.com/returntocorp/semgrep-rules/pull/2249#discussion_r926929790


colleend (approved):
cool, thanks for reducing FPs!

PR #2271, opened 2022-07-28
Update Node/Express SQL Rules

LewisArdern:
Update mysql and knex to use better sources

### DIFF
--- a/javascript/lang/security/audit/sqli/node-mysql-sqli.yaml
+++ b/javascript/lang/security/audit/sqli/node-mysql-sqli.yaml
@@ -2,8 +2,8 @@ rules:
   - id: node-mysql-sqli
     severity: WARNING
     message: >-
-      Detected a `$IMPORT` SQL statement that is tainted by user-input. This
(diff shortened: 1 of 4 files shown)


0xDC0DE (approved):
lgtm!

javascript/lang/security/audit/sqli/node-mysql-sqli.yaml
```diff
@@ -29,7 +31,14 @@ rules:
     pattern-sources:
       - patterns:
           - pattern-inside: function ... (..., $Y,...) {...}
-          - focus-metavariable: $Y
+          - pattern: $Y
+          - pattern-not-inside: |
+              function ... (..., $Y: number,...) {...}
+          - pattern-not-inside: $Y.query
```
IagoAbal, line 37: It would be more efficient to keep the `focus-metavariable: $Y` and instead use sanitizers, or else enumerate the cases `$Y.FOO` that are problematic.

PR #2275, opened 2022-07-29
secrets clean-up

LewisArdern:
This cleans up various rules which react hard-coded secrets.
* Adds taint mode to improve capturing
* Improves some rules by removing patterns and using metavariable-regex
* Adds and improves metadata to be very generic, which will allow for automative improvements later..


go/jwt-go/security/jwt.yaml
```diff
@@ -1,19 +1,32 @@
...
-          $Y := $TOKEN.SignedString($X)
-      - pattern: |
-          $TOKEN.SignedString([]byte("..."))
-    message: JWT token is hardcoded
-    languages: [go]
     severity: WARNING
+    patterns:
+      - pattern-either:
+        - pattern: |
+            $X = []byte("$F")
```
kurt-r2c: Are we keeping this search mode to handle interprocedural behavior in OSS Semgrep?

javascript/jose/security/jwt-exposed-credentials.yaml
```diff
@@ -4,230 +4,52 @@ rules:
...
+                    ...
+                - pattern-inside: |
+                    import * as $JWT from "jsonwebtoken"
+                    ...
+                - pattern-inside: |
+                    import {...,$JWT,...} from "jsonwebtoken"
+                    ...
+            - pattern-either:
+                - pattern-inside: |
+                    $JWT.sign($F,...);
```
kurt-r2c: the signing key for jsonwebtoken is the second positional arg: https://github.com/auth0/node-jsonwebtoken/blob/master/sign.js#L82

This rule seems to be for accidentally adding passwords to the signed payload (name is `jwt-exposed-credentials`), but the messaging is for secrets in source.

The case that this finds is *highly* unusual and I think it ought to be two rules:

* rule for finding "password" in JWT payload (info disclosure)
* rule for hardcoded signing key

javascript/jsonwebtoken/security/jwt-exposed-credentials.yaml
```diff
@@ -4,57 +4,52 @@ rules:
...
-        - https://cwe.mitre.org/data/definitions/522.html
       category: security
       technology:
         - jwt
-    message: >-
-      Password is exposed through JWT token payload. This is not encrypted and
-      the password could be compromised. Do not store passwords in JWT tokens.
+        - secrets
+    message: >- 
+      A hard-coded credential was detected. It is not recommended
```
kurt-r2c: See other comment re: exposed-credential rules. I think this was intended to be an info disclosure rule, not a hardcoded secret rule.

kurt-r2c (changes requested):
The JWT exposed credentials rules should be either retired/renamed: right now their messaging does not agree with their pattern behavior/rule names.

go/jwt-go/security/jwt.yaml
LewisArdern: i think i was lazy let me see if i can convert it into taint

LewisArdern: updated to taint

kurt-r2c (approved):
LGTM

PR #2278, opened 2022-08-01
update direct-response-write

LewisArdern:
Updated to use express sources, narrowed down to only obvious sources since its direct reflection. 

Removed the set-content header since it will likely introduce FPs will write a new rule to capture that edge case where text/html is set.


colleend (approved):
lgtm 👍

PR #2280, opened 2022-08-02
Add support for github actions inputs

luisfontes19:
Github actions/workflows can have external [inputs](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#onworkflow_dispatchinputs), which are hold in the `github.event.input` variable.

Added support for those cases as well

### DIFF
--- a/yaml/github-actions/security/run-shell-injection.test.yaml
+++ b/yaml/github-actions/security/run-shell-injection.test.yaml
@@ -3,6 +3,10 @@ name: docker
 
 on:
   workflow_dispatch:
+    inputs:
+      message_to_print:
+        type: string
+        required: false
   push:
     branches:
       - develop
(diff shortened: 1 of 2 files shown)


yaml/github-actions/security/run-shell-injection.test.yaml
```diff
@@ -83,7 +87,9 @@ jobs:
           echo "PR title did not start with 'octocat'"
           exit 1
           fi
-
+      - name: Print a message
```
colleend, line 90: We need to add a `#ruleid: run-shell-injection` for tests to pass and also to annotate that we expect this to match.

colleend (changes requested):
Thanks for submitting this change! We appreciate it -- there's one small addition that is needed 🙏

colleend (approved):
looks good to me, thanks for the PR!

PR #2284, opened 2022-08-02
Tainted shell call fix fp

colleend:
* fix some FP for this rule in cases where `params` is overloaded.
* changed taint source to be specifically accessing `params` keys

_To Test:_
Run `semgrep --test .` in the repository

### DIFF
--- a/ruby/rails/security/audit/avoid-tainted-shell-call.yaml
+++ b/ruby/rails/security/audit/avoid-tainted-shell-call.yaml
@@ -15,7 +15,8 @@ rules:
     severity: ERROR
     mode: taint
     pattern-sources:
-      - pattern: params
+    - pattern-either:
+      - pattern: params[...]
       - pattern: cookies
       - pattern: request.env
     pattern-sinks:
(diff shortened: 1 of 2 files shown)


ruby/rails/security/audit/avoid-tainted-shell-call.yaml
```diff
@@ -1,7 +1,7 @@
 rules:
   - id: avoid-tainted-shell-call
     metadata:
-      owasp: "A1: Injection"
+      owasp: "A01:2017 - Injection"
```
kurt-r2c: should maybe add A03:2021 - Injection as well?

kurt-r2c (approved):
LGTM - one minor metadata suggestion

ruby/rails/security/audit/avoid-tainted-shell-call.yaml
colleend: sounds good to me :D

colleend: done!

kurt-r2c (approved):

PR #2287, opened 2022-08-04
moving away from regex to generic for template href var

colleend:
Received FP report for this rule, so moving it away from regex and more towards generic mode 👍


colleend:
validate is failing on `shell-injection.yaml`, not this rule 👍

minusworld (approved):

PR #2301, opened 2022-08-09
Delete twitter-bot.yaml

minusworld:
The Twitter Bot hasn't been active since May, and no one seems to have noticed. Therefore, I think we should remove it.
RIP Twitter Bot, you served us well 🫡


inkz (approved):
R.I.P. :pray:

PR #2312, opened 2022-08-15
Update `vm-injection` rule

inkz:
merge `vm-injection-*` rules into one + add taint


LewisArdern:
@inkz shouldn't we just update the audit rule which already exists vs 'deprecating' and making a new rule.

LewisArdern (changes requested):
merge this into the audit rule IMO

inkz:
@LewisArdern yes we should:) but it brakes the tests
tests do not allow for having multiple rules in one file and not having test cases for all of them
either deprecate all of them or none of them :(
I decided to create another rule and deprecate existing one

LewisArdern:
I see, got it

LewisArdern (approved):

inkz:
> merge this into the audit rule IMO

@LewisArdern I did it! :smile:  I was wrong, it was possible to merge :+1:

minusworld (approved):

PR #2318, opened 2022-08-16
Exclude constants from file inclusion in PHP

luisfontes19:
Assuming constants are trusted we can exclude them from findings. This can help remove tons of False Positives. 

I changed the logic of the rule to only report findings where a regular variable (starting with $) is found inside the require function.

### DIFF
--- a/php/lang/security/file-inclusion.php
+++ b/php/lang/security/file-inclusion.php
@@ -35,3 +35,6 @@ include_safe(__DIR__ . $user_input);
 
 // ok: file-inclusion
 require_once(CONFIG_DIR . '/constant.php');
+
+// ok: file-inclusion
+require_once( dirname( __FILE__ ) . '/admin.php' );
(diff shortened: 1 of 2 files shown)


minusworld:
Are there any conditions in PHP where a variable is not prefaced with the dollar sign `$`?

luisfontes19:
hi @minusworld 

There are a few yes

You can use the method [define](https://www.php.net/manual/en/function.define.php) to create constants and use it like so:

```php
define("MYVAR", "Hello world.");
echo MYVAR;
```


Constants are usually trusted as they are defined when the app boots but developers can do anything 😅

I had semgrep running on wordpress source code and had quite a few FP's like with methods being called inside the require (like the example I just added to the tests) or others like:

```php
include( ABSPATH . 'wp-admin/edit-form-blocks.php' );
```

minusworld (approved):

PR #2325, opened 2022-08-18
Filter SafeDumper from yaml deserialization

minusworld:


chmccreery (approved):

chmccreery (approved):
Can we also filter out `yaml.SafeLoader` from the `pyyaml.load(...)` function and add a fix to use the `SafeDumper` and `SafeLoader` here? In fact pyyaml.load(...) and pyyaml.dump(...) already complains if you don't use the SafeLoader and SafeDumper so maybe that library is ok now and can be removed from this rule altogether?

minusworld:
Probably want to keep this around for just a while longer while old Python versions phase out, but generally yeah, if the Python interpreter complains about this, there's not much need for a Semgrep rule anymore unless that's just someone's preferred workflow

Issue #1694, opened 2022-01-04
[Rule] Checkov Terraform GCP Rules


minusworld:
**Rule Description**

- [ ] CloudArmorWAFACLCVE202144228
- [ ] CloudStorageLogging
- [ ] CloudStorageSelfLogging
- [ ] GKEAliasIpEnabled
- [ ] GKEBasicAuth
- [ ] GKEBinaryAuthorization
- [ ] GKEClientCertificateEnabled
- [ ] GKEClusterLogging
- [ ] GKEDisableLegacyAuth
- [ ] GKEEnableShieldedNodes
- [ ] GKEEnableVPCFlowLogs
- [ ] GKEEnsureIntegrityMonitoring
- [ ] GKEHasLabels
- [ ] GKEKubernetesRBACGoogleGroups
- [ ] GKELegacyInstanceMetadataDisabled
- [ ] GKEMasterAuthorizedNetworksEnabled
- [ ] GKEMetadataServerIsEnabled
- [ ] KEMetadataServerIsEnabled to handle metadata as list
- [ ] GKEMonitoringEnabled
- [ ] GKENetworkPolicyEnabled
- [ ] GKENodePoolAutoRepairEnabled
- [ ] GKENodePoolAutoUpgradeEnabled
- [ ] GKEPodSecurityPolicyEnabled
- [ ] GKEPrivateClusterConfig
- [ ] GKEPrivateNodes
- [ ] GKEPublicControlPlane
- [ ] GKEReleaseChannel
- [ ] GKESecureBootforShieldedNodes
- [ ] GKEUseCosImage
- [ ] GoogleBigQueryDatasetPublicACL
- [ ] GoogleCloudDNSKeySpecsRSASHA1
- [ ] GoogleCloudDNSSECEnabled
- [ ] GoogleCloudMySqlLocalInfileOff
- [ ] GoogleCloudPostgreSqlLogCheckpoints
- [ ] GoogleCloudPostgreSqlLogConnection
- [ ] GoogleCloudPostgreSqlLogDisconnection
- [ ] GoogleCloudPostgreSqlLogLockWaits
- [ ] GoogleCloudPostgreSqlLogMinDuration
- [ ] GoogleCloudPostgreSqlLogMinMessage
- [ ] GoogleCloudPostgreSqlLogTemp
- [ ] GoogleCloudSqlBackupConfiguration
- [ ] GoogleCloudSqlDatabasePublicallyAccessible
[...]

PR #2327, opened 2022-08-19
Adding migrated GCP checkov rules to GCP terraform checks - Part 1

mertcancoskuner:
This is part 1 of addressing https://github.com/returntocorp/semgrep-rules/issues/1694


terraform/gcp/security/gcp-bigquery-dataset-encrypted-with-cmk.tf
```diff
@@ -0,0 +1,36 @@
...
+  location                    = "EU"
+  default_table_expiration_ms = 3600000
+
+  labels = {
+    env = "default"
+  }
+
+  access {
+    role          = "OWNER"
+    special_group = "allAuthenticatedUsers"
```
minusworld, line 16: Is the issue in this case that there is no encryption, or that allAuthenticatedUsers is used here?

terraform/gcp/security/gcp-compute-boot-disk-encryption.yaml
```diff
@@ -0,0 +1,27 @@
...
+  patterns:
+    - pattern: |
+        resource "google_compute_instance" $ANYTHING {
+          ...
+        }
+    - pattern-not-inside: |
+        resource "google_compute_instance" $ANYTHING {
+          ...
+          boot_disk {
+            disk_encryption_key_raw = ...
```
minusworld, line 12: Are there other encryption methods besides raw keys? Something like the CMK keys in some of the other rules?

terraform/gcp/security/gcp-compute-disk-encryption.yaml
```diff
@@ -0,0 +1,27 @@
...
+  patterns:
+    - pattern: |
+        resource "google_compute_disk" $ANYTHING {
+          ...
+        }
+    - pattern-not-inside: |
+        resource "google_compute_disk" $ANYTHING {
+          ...
+          disk_encryption_key {
+            raw_key = ...
```
minusworld, line 12: Same question as above; commenting here for reference later

terraform/gcp/security/gcp-compute-ip-forward.yaml
```diff
@@ -0,0 +1,22 @@
...
+  patterns:
+    - pattern: resource
+    - pattern-inside: |
+        resource "google_compute_instance" "..." {
+        ...
+        can_ip_forward = true
+        ...
+        }
+  message: >-
+    Ensure that IP forwarding is not enabled on Instances
```
minusworld: What's the danger of IP forwarding? Does it allow for SSRF or something similar?

terraform/gcp/security/gcp-compute-project-os-login.yaml
```diff
@@ -0,0 +1,28 @@
...
+    - pattern: resource
+    - pattern-inside: |
+        resource "google_compute_project_metadata" "..." {
+        ...
+        }
+    - pattern-not-inside: |
+        resource "google_compute_project_metadata" "..." {
+        ...
+        metadata = {
+            enable-oslogin = "True"
```
minusworld, line 13: Any idea why the value for this resource is a string, and the value for the instance version of this rule is a boolean? 🤔

minusworld (changes requested):
Left a few questions

terraform/gcp/security/gcp-bigquery-dataset-encrypted-with-cmk.tf
mertcancoskuner: The issue in this case is that there is no encryption.

terraform/gcp/security/gcp-compute-disk-encryption.yaml
mertcancoskuner: According to [google_compute_instance](https://registry.terraform.io/providers/hashicorp/google/latest/docs/resources/compute_instance) terraform docs `kms_key_self_link` as well.

terraform/gcp/security/gcp-compute-boot-disk-encryption.yaml
mertcancoskuner: According to [google_compute_instance](https://registry.terraform.io/providers/hashicorp/google/latest/docs/resources/compute_instance) terraform docs `kms_key_self_link` as well.

terraform/gcp/security/gcp-compute-ip-forward.yaml
mertcancoskuner: When the IP Forwarding feature is enabled on a virtual machine's network interface (NIC), it allows the VM to act as a  router/packet forwarder and receive traffic addressed to other destinations.

terraform/gcp/security/gcp-compute-project-os-login.yaml
mertcancoskuner: According to [setCommonInstanceMetadata ](https://cloud.google.com/compute/docs/reference/rest/v1/projects/setCommonInstanceMetadata) JSON representation, fields expect String. Examples I've encountered regarding `enable-oslogin` were also using String ([example1](https://yuweisung.medium.com/applying-gcp-os-login-to-terraform-and-ansible-da1d78c016b5), [example2](https://binx.io/nl/2022/01/28/how-to-use-os-login-for-ssh-access-to-vms-on-gcp/)). 

However, [this](https://aquasecurity.github.io/tfsec/v1.6.2/checks/google/compute/project-level-oslogin/) also shows that it can use boolean. Do you think we should cover the boolean case just to be sure? If yes, how can we change the condition to capture both?

minusworld: I guess it won't hurt to cover both cases 🤷 

The easiest way is to add another pattern-not-inside clause for the boolean case. Negative clauses just act like filters, so they will work even if they're on the same level.

mertcancoskuner: Added the use case. Pipeline fails on an irrelevant rule (/src/javascript/angular/security/detect-angular-element-methods.js)

minusworld (approved):

PR #2333, opened 2022-08-20
Fix Typo - detect-non-literal-require.yaml

cydave:
Small typo spotted :)


javascript/lang/security/detect-non-literal-require.yaml
```diff
@@ -9,8 +9,7 @@ rules:
         - javascript
     message: >-
       Detected the use of require(variable). Calling require with a non-literal argument
-      might
-      allow an attacker to load an run arbitrary code, or access arbitrary files.
```
cydave, line 13: "... attacker to load **an** run arbitrary ..." =>  "... attacker to load **and** run arbitrary ..."

PR #2338, opened 2022-08-21
Adding migrated GCP checkov rules to GCP terraform checks - Part 2

mertcancoskuner:
This is part 2 of addressing https://github.com/returntocorp/semgrep-rules/issues/1694


minusworld:
Bump

terraform/gcp/best-practice/gcp-kms-rotation-period.yaml
```diff
@@ -0,0 +1,22 @@
...
+    - pattern: resource
+    - pattern-inside: |
+          resource "google_kms_crypto_key" "..." {
+            ...
+            rotation_period = "$ROTATION"
+            ...
+          }
+    - metavariable-comparison:
+        metavariable: $ROTATION
+        comparison: $ROTATION > 7776000
```
minusworld: I assume this is failing because the test cases have an `s` string at the end. Let me see if there's anything that helps process strings in `metavariable-comparison` that might be undocumented.

minusworld: I've noodled on this for a bit but I can't figure out any way to do the comparison successfully. I've pointed this out to the Semgrep folks so it might require a new feature or modification to an existing one to make work.

In the meantime, what if we took this rule out from this PR, made an individual PR with it, and then moved forward with the rest of the rules?

mertcancoskuner: Sounds good to me!

minusworld: Editor link for preserving the rule:https://semgrep.dev/playground/s/wwbB

terraform/gcp/security/gcp-gke-client-certificate-disabled.yaml
```diff
@@ -0,0 +1,34 @@
...
+        ...
+        }
+    - pattern-not-inside: |
+        resource "google_container_cluster" "..." {
+        ...
+        master_auth {
+          ...
+          client_certificate_config {
+            ...
+            issue_client_certificate = true
```
minusworld: The message below says to ensure that client certificate authentication is disabled. This `pattern-not-inside` filters out resources where `issue_client_certificate` is set to `true`. Is this the opposite of what's supposed to be happening?

minusworld (changes requested):
Left one clarification question, otherwise lgtm

terraform/gcp/security/gcp-gke-client-certificate-disabled.yaml
mertcancoskuner: That was supposed to be the opposite (https://cloud.google.com/kubernetes-engine/docs/how-to/hardening-your-cluster#restrict_authn_methods)

minusworld (approved):

PR #2350, opened 2022-08-25
Find sensitive Bicep parameters without @secure() attribute

Sjord:
Determine from the parameter name whether it is sensitive. I left out "key" for now, because that may trigger too many false positives.

Marking a parameter with @secure() prevents logging and displaying in Azure.

### DIFF
--- a/generic/bicep/security/secure-parameter-for-secrets.bicep
+++ b/generic/bicep/security/secure-parameter-for-secrets.bicep
@@ -1,0 +1,12 @@
+// ok: secure-parameter-for-secrets
+@secure()
+param demoPassword string
+
+// ok: secure-parameter-for-secrets
+param normalParam string
+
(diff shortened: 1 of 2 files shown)


Sjord:
I created a new `bicep` directory, but maybe this should be a subdirectory of `generic`?

minusworld:
> I created a new bicep directory, but maybe this should be a subdirectory of generic?

This is probably better as a subdirectory of `generic`, but otherwise it looks good to me! Ping me here when you make the change. Thanks!

Sjord:
@minusworld I moved it, could you take another look?

p4p3r (approved):
Hey @Sjord, thanks a lot for your contribution! LGTM!
I'm going to merge it into `develop` now, and it will be merged into `release` in the next few days as soon as we cut a new release.

PR #2358, opened 2022-08-26
update test for symbolic propagation of expressions

brandonspark:
**What:**
I am adding a change to propagate arbitrary expressions symbolically, so long as they are assigned to a single time.

In addition, I have also made it so propagation of singly-assigned variables does not only happen in Javascript. Thus, this test now has an extra finding.

**Why:**
This change is with the intention of allowing propagation of more cases.

### DIFF
--- a/java/java-jwt/security/jwt-hardcode.java
+++ b/java/java-jwt/security/jwt-hardcode.java
@@ -48,6 +48,7 @@ abstract class App2
 
     public void bad2() {
         try {
+            // ruleid: java-jwt-hardcoded-secret
             Algorithm algorithm = Algorithm.HMAC256(secret);
             String token = JWT.create()
                 .withIssuer("auth0")


java/java-jwt/security/jwt-hardcode.java
```diff
@@ -48,6 +48,7 @@ public static void main( String[] args )
 
     public void bad2() {
         try {
+            // ruleid: java-jwt-hardcoded-secret
             Algorithm algorithm = Algorithm.HMAC256(secret);
```
IagoAbal, line 52: Shouldn't have this be matching before? Maybe I haven't looked into this carefully enough, but it feels like it should have matched already.

java/java-jwt/security/jwt-hardcode.java
```diff
@@ -48,6 +48,7 @@ public static void main( String[] args )
 
     public void bad2() {
         try {
+            // ruleid: java-jwt-hardcoded-secret
```
IagoAbal, line 51: Yeah wait a sec, should we really be flagging this? This is not guaranteed to be a constant.

brandonspark: How would it not be, if the value of `secret` is never touched again? I may be unclear on Java.

IagoAbal: The problem is that we _do not know_ if `secret` will ever be touched again, given its visibility another object could surely modify it. And const-prop is supposed to be a _must_ analysis, so if we report that something is constant when it may not be, that is a bug.

PR #2363, opened 2022-08-30
Rule to detect secrets in build time arguments in Docker

Sjord:
It is not recommended to use build-time variables for passing secrets like github keys, user credentials etc. Build-time variable values are visible to any user of the image with the docker history command.

### DIFF
--- a/dockerfile/security/secret-in-build-arg.dockerfile
+++ b/dockerfile/security/secret-in-build-arg.dockerfile
@@ -1,0 +1,36 @@
+# ok: secret-in-build-arg
+ARG BASE_IMG=rust
+# ok: secret-in-build-arg
+ARG BASE_TAG=latest
+
+FROM ${BASE_IMG}:${BASE_TAG}
+
+# ok: secret-in-build-arg
+ARG environment
+
+# ok: secret-in-build-arg
+ARG BUILD_PACKAGES="ruby-dev build-base curl-dev git openssl bash"
+
+# ok: secret-in-build-arg
+ARG DEV_PACKAGES="postgresql-dev"
+
+# ok: secret-in-build-arg
+ARG RUBY_PACKAGES="tzdata"
+
+# ok: secret-in-build-arg
+ARG NAME_IS_OK="secret password token key"
+
+# ruleid: secret-in-build-arg
+ARG SECRET_KEY_BASE
+
+# ruleid: secret-in-build-arg
+ARG SECRET_TOKEN
+
+# ruleid: secret-in-build-arg
+ARG DJANGO_SECRET_KEY
+
+# ruleid: secret-in-build-arg
+ARG NEXUS_PASSWORD
+
+# ruleid: secret-in-build-arg
+ARG nexus_password
--- a/dockerfile/security/secret-in-build-arg.yaml
+++ b/dockerfile/security/secret-in-build-arg.yaml
@@ -1,0 +1,27 @@
+rules:
+- id: secret-in-build-arg
+  patterns:
+  - pattern: ARG $ARG
+  - metavariable-regex:
+      metavariable: $ARG
+      regex: (?i).*(password|secret|token)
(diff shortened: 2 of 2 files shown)


dockerfile/security/secret-in-build-arg.yaml
```diff
@@ -0,0 +1,24 @@
...
+    Docker build time arguments are not suited for secrets, because the
+    argument values are saved with the image. Running `docker image history` on the
+    image will show information on how the image was built, including arguments. If
+    these contain secrets, anyone with access to the docker image can access those
+    secrets.
+  metadata:
+    category: security
+    technology:
+      - dockerfile
+    cwe: "CWE-538: Insertion of Sensitive Information into Externally-Accessible File or Directory"
```
p4p3r, line 18: Would you mind adding a reference to the Owasp Top10 too?

```
      owasp:
        - A07:2021 - Identification and Authentication Failures
        - A03:2017 - Sensitive Data Exposure
```

dockerfile/security/secret-in-build-arg.dockerfile
```diff
@@ -0,0 +1,33 @@
+# ok: secret-in-build-arg
+ARG BASE_IMG=rust   # default image
```
p4p3r: I don't think you can have comments on the same line of the ARG command (I see a syntax error when running this in the playground).
I would simply these first four lines to:
```
# default image
# ok: secret-in-build-arg
ARG BASE_IMG=rust   
# default tag
# ok: secret-in-build-arg
ARG BASE_TAG=latest
```

p4p3r (dismissed):

dockerfile/security/secret-in-build-arg.yaml
Sjord: I couldn't really find an applicable entry in the top 10 of 2021. I don't think `A07:2021 - Identification and Authentication Failures` applies, really. But I'll add `A03:2017 - Sensitive Data Exposure`.

dockerfile/security/secret-in-build-arg.dockerfile
Sjord: [Semgrep fails to parse fake comment after ARG in Dockerfile · Issue #6008 · returntocorp/semgrep](https://github.com/returntocorp/semgrep/issues/6008)

Also, I removed the "comments" in the test.

dockerfile/security/secret-in-build-arg.yaml
p4p3r: The "What's changed..." section of https://owasp.org/Top10/ actually seems to suggest that `A03:2017 - Sensitive Data Exposure.` -> `A02:2021 Cryptographic Failures`: what do you think? Still rather broad but it make sense (leaking sensitive data can be considered a lack of cryptographic features)

Sjord: Since there is no cryptography involved in either the problem or the solution, I would say _A02:2021 Cryptographic Failures_ is not a match.

The OWASP top 10 is not meant to be exhaustive. Some vulnerabilities fall into none of the categories.

LewisArdern: We should use whatever OWASP mapping is defined in the CWE category, if there is no memberships then we should not generate the OWASP fields, in this case:

```
CWE-538: Insertion of Sensitive Information into Externally-Accessible File or Directory
```
has:
image

Therefore we should add:
```
owasp:
 - A01:2021 - Broken Access Control
```

We are moving to a process where OWASP will not be mandatory, and this approach takes out the guessing game of what should we add @Sjord

Sjord: I added - A01:2021 - Broken Access Control

Sjord:
@p4p3r Could you take another look at this?

kurt-r2c (approved):

PR #2364, opened 2022-08-31
Catch up with latest semgrep (or vice-versa)

mjambon:
The semgrep repo now points to a specific commit of semgrep-rules (via a git submodule) which doesn't get updated automatically. Today, we had a problem caused by a new behavior in semgrep which would have been fixed by using a more recent version of semgrep-rules. So, we upgraded semgrep-rules. This gave us 8 new errors of various nature that this PR fixes. The tests now pass when using the latest version of semgrep.

Here's a description of the different issues and possible long-term improvements:
* semgrep-core recognizes `ruleid:` but not `ruleid :`. We should agree on whether a space is allowed and use the same spec everywhere.
* there were stray zero-width spaces (normally used to define breaking points within words) in C# targets.They result in partial parsing of the target files, which in this case is totally fine since these characters end up being treated like other ignorable whitespace. It would be good to treat these errors as test failures so as to get the same behavior as semgrep-core tests.
* similarly, there was an unclosed string literal in a typescript target which results in partial parsing of the target.
* one deprecated python rule file didn't have a matching target, which was a problem for semgrep-core. We can fix this in semgrep-core if necessary in the future or if having an empty python target is not desirable.
* two different rules had new matches in places that were marked as "to do", so I removed the `todo` prefix. This was nice.

The CI failure for this PR is:
```
The following test(s) failed with Semgrep v0.108.0 are caused by this PR:
javascript/lang/security/audit/sqli/node-postgres-sqli.js
ruby/rails/security/audit/sqli/ruby-pg-sqli.rb
```
... which correspond to the to-dos that are now implemented. :tada: 

The semgrep repo is now depending on this branch. To avoid surprises with git/github, it's best to either merge it with a merge commit (easier) OR squash and update the submodule accordingly in the semgrep repo.

### DIFF
--- a/csharp/dotnet/security/audit/open-directory-listing.cs
+++ b/csharp/dotnet/security/audit/open-directory-listing.cs
@@ -2,18 +2,18 @@
 public void Configure(IApplicationBuilder app, IHostingEnvironment env)
 {
     var builder = WebApplication.CreateBuilder(args);
-    // ruleid : open-directory-listing
+    // ruleid: open-directory-listing
     builder.Services.AddDirectoryBrowser();
 
     var fileProvider = new PhysicalFileProvider(Path.Combine(builder.Environment.WebRootPath, "data"));
     var requestPath = "/data";
 
-    // ruleid : open-directory-listing
+    // ruleid: open-directory-listing
     app.UseDirectoryBrowser(new DirectoryBrowserOptions
     {
         FileProvider = fileProvider,
         RequestPath = requestPath
-    });    
+    });
 }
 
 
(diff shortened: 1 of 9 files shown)


p4p3r (approved):
LGTM

mjambon:
What should we do with "rules-test-historical" which is failing? Some to-dos were implemented in semgrep, so the expected findings changed. The older version of semgrep can't find these findings. This is an improvement, not a regression. I don't understand what this check is intended for. Is it ok to merge even though each future PR will get this error?

p4p3r:
> What should we do with "rules-test-historical" which is failing? Some to-dos were implemented in semgrep, so the expected findings changed. The older version of semgrep can't find these findings. This is an improvement, not a regression. I don't understand what this check is intended for. Is it ok to merge even though each future PR will get this error?

The check is mostly intended to catch situations where new rules (because of a new syntactic construct perhaps, etc.) would make "catastrophically" fail a slightly older versions of Semgrep (3 versions in the past). It's there as a safety net to catch situations where merging would require a more "coordinated" approach.

PR #2365, opened 2022-08-31
Misc fixes to make tests pass in semgrep repo

nmote:
I'm updating the semgrep-rules submodule in semgrep, which we use for
testing purposes. I think there must be some differences in behavior
between the test runner that we use in the semgrep repo and the test
runner used here.

Most of these are parse errors. The tree-sitter error recovery appears
to do a pretty good job in the face of these issues, so they don't
prevent Semgrep from running. In particular, there were several files
that had the U+200B (zero width space) character, which led to parse
errors.

A few places used `todoruleid:`. The necessary Semgrep version has been
released so I just changed them to `ruleid:`. One file used `ruleid :`
with a space before the colon. I removed the space.

### DIFF
--- a/csharp/dotnet/security/audit/open-directory-listing.cs
+++ b/csharp/dotnet/security/audit/open-directory-listing.cs
@@ -2,13 +2,13 @@
 public void Configure(IApplicationBuilder app, IHostingEnvironment env)
 {
     var builder = WebApplication.CreateBuilder(args);
-    // ruleid : open-directory-listing
+    // ruleid: open-directory-listing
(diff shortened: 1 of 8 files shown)


nmote:
Interesting, it looks like it runs the rules with old versions of Semgrep, so changing `todoruleid` to `ruleid` doesn't work. Maybe we should just update the test runner used in `semgrep-core` to accept either a finding or no finding when it sees `todoruleid`? That would reduce the signal we get from tests, but it might be the best way forward. Thoughts?

nmote:
I didn't see #2364

PR #2366, opened 2022-08-31
Fix syntax error in httpservlet-path-traversal.java

nmote:
The last two curly braces are missing. I'm not sure why Semgrep
currently parses this without errors, but the changes I'm making in
https://github.com/returntocorp/ocaml-tree-sitter-semgrep/pull/339 lead
to a parse error here. I'm not sure why, but since this is clearly
malformed code, I don't intend to spend time figuring it out.


p4p3r (approved):
LGTM

PR #2370, opened 2022-09-05
add github-script-injection rule

inkz:
The rule is similar to: https://github.com/returntocorp/semgrep-rules/blob/develop/yaml/github-actions/security/run-shell-injection.yaml but targets popular [`actions/github-script`](https://github.com/actions/github-script) action


minusworld (approved):
nice

PR #2376, opened 2022-09-08
Add rule to find classic crypto mode of operations without MAC

luisfontes19:
Rules to find modes of operation (CTR, CBC, CFB, OFB) without MAC (in this case HMAC only) which makes them somewhat insecure. 

I created two rules, one for pycryptodome(x) and another for cryptography package


inkz (approved):

inkz:
Thank you! :pray:

PR #2379, opened 2022-09-08
Rearrange tests for completeness and for sharing with semgrep repo's CI

mjambon:
There were 2 issues:
* `semgrep --test` was not strict enough, ignoring parse errors in target code
* `make test` wasn't testing all the rules and targets

Relevant issues:
https://github.com/returntocorp/semgrep/issues/6068
https://github.com/returntocorp/semgrep/issues/6069

After this, `make test` will take longer. If a quick version of `make test` that only tests a few rules is desired, let me know. The semgrep repo will start relying on this `make test` to validate and test all the rules in semgrep-rules.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -17,18 +17,27 @@ SEMGREP ?= semgrep
 #
 # The semgrep repo also runs this as part of its CI.
 #
+.PHONY: test
 test:
+	$(MAKE) validate
+	$(MAKE) test-only
+
+.PHONY: validate
+validate:
 	for root in $(RULE_FOLDERS); do \
 	  echo "======== validate rule files in $$root ========"; \
 	  time -p $(SEMGREP) --validate \
 	    --strict --disable-version-check \
 	    --metrics=off --verbose \
 	    --config="$$root"; \
-	done
+	done 2>&1
+
+.PHONY: test-only
+test-only:
 	for root in $(RULE_FOLDERS); do \
 	  echo "========= test rules in $$root ========="; \
 	  time -p $(SEMGREP) --test --test-ignore-todo \
 	    --strict --disable-version-check \
 	    --metrics=off --verbose \
 	    "$$root"; \
-	done
+	done 2>&1
--- a/Makefile
+++ b/Makefile
@@ -35,7 +35,7 @@ validate:
 	  time -p $(SEMGREP) --validate \
 	    --strict --disable-version-check \
 	    --metrics=off --verbose \
-	    --config="$$root"; \
+	    --config="./$$root"; \
 	done >&2
 
 .PHONY: test-only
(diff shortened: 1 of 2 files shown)


mjambon:
I don't see why this test is failing:
```
	✖ detect-angular-element-methods                               missed lines: [8, 10], incorrect lines: [] 
1487
	test file path: /src/javascript/angular/security/detect-angular-element-methods.js
```
https://github.com/returntocorp/semgrep-rules/runs/8260952680?check_suite_focus=true

The following works for me. It uses the upcoming https://github.com/returntocorp/semgrep/pull/6070:
```
$ semgrep --test --strict --disable-version-check --metrics=off --verbose ./javascript/angular/security
```

Update: I can reproduce after a docker pull with the following command:
```
$ docker run --rm -v ~/semgrep/semgrep-core/tests/semgrep-rules:/src returntocorp/semgrep:develop semgrep --test /src/javascript/angular/security
...
	✖ detect-angular-element-methods                               missed lines: [8, 10], incorrect lines: [] 
	test file path: /src/javascript/angular/security/detect-angular-element-methods.js


No tests for fixes found.
```
This is version 0.112.1:
```
$ docker run --rm -v ~/semgrep/semgrep-core/tests/semgrep-rules:/src returntocorp/semgrep:develop semgrep --version
Semgrep failed to set the safe.directory Git config option. Git commands might fail: Command failed with exit code: 128
-----
Command failed with output:
fatal: not a git repository: /src/../../../.git/modules/semgrep-core/tests/semgrep-rules


Failed to run 'git config --global --add safe.directory *'. Possible reasons:

- the git binary is not available
- the current working directory is not a git repository
- the current working directory is not marked as safe
    (fix with `git config --global --add safe.directory $(pwd)`)

Try running the command yourself to debug the issue.
0.112.1
```

Update:

The main branch (`develop`) shows the same error. This is a dummy PR that has the problem: https://github.com/returntocorp/semgrep-rules/pull/2381

Makefile
```diff
@@ -1,8 +1,50 @@
+# List of folders containing Semgrep rule files (with a .yml suffix).
+#
+# This is meant to exclude folders that don't contain rule files and
+# may contain .yml files that are not Semgrep rules and would result
+# in errors.
+#
+RULE_FOLDERS = $(shell ls -d */ | grep -v '^stats')
```
inkz: I would also exclude `trusted_python` and `fingerprints` folders
the first one does not contain rules
second one contains CAI rules that do not have tests

inkz: oh `stats` and `fingerprints` are removed before the scan, so no worries about them
https://github.com/returntocorp/semgrep-rules/actions/runs/3019231154/workflow#L22

mjambon: thanks for taking a look, I hadn't noticed that these folders were being removed in the CI job. In semgrep, we used to remove `stats/` but it's unclean. Here's what I'm going to do:
1. filter out `trusted_python` and `fingerprints` in addition to `stats` in the makefile
2. make the CI job use the makefile. This will be https://github.com/returntocorp/semgrep-rules/issues/2380 which right now waiting on semgrep becoming available (and also, git permission issues).

mjambon: (1) is done. (2) will be done later.

inkz (approved):

PR #2391, opened 2022-09-14
fix tainted-sql-string FP from sentiment

kurt-r2c:
- add FP test case
- remove overly permissive regex
- add find `conditions => ...` matcher

### DIFF
--- a/ruby/rails/security/injection/tainted-sql-string.yaml
+++ b/ruby/rails/security/injection/tainted-sql-string.yaml
@@ -41,10 +41,10 @@ rules:
                 - focus-metavariable: $X
             - patterns:
                 - pattern: |
-                    "...#{$EXPR}..."
+                    "$SQLVERB#{$EXPR}..."
                 - pattern-not-inside: |
                     $FUNC("...",  "...#{$EXPR}...",...)
-                - focus-metavariable: $EXPR
+                - focus-metavariable: $SQLVERB
                 - pattern-regex: (?i)(select|delete|insert|create|update|alter|drop)\b
             - patterns:
                 - pattern-either:
--- a/ruby/rails/security/injection/tainted-sql-string.yaml
+++ b/ruby/rails/security/injection/tainted-sql-string.yaml
@@ -1,54 +1,61 @@
 rules:
 - id: tainted-sql-string
   languages:
-  - ruby
+    - ruby
   severity: ERROR
-  message: Detected user input used to manually construct a SQL string. This is usually
-    bad practice because manual construction could accidentally result in a SQL injection.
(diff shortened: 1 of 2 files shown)


ruby/rails/security/injection/tainted-sql-string.yaml
```diff
@@ -1,54 +1,61 @@
...
-        - focus-metavariable: $EXPR
-        - metavariable-regex:
-            metavariable: $SQLSTR
-            regex: (?i)(select|delete|insert|create|update|alter|drop)\b
+            - patterns:
+                - pattern-either:
+                    - pattern: |
+                        $RECORD.where($X,...)
+                    - pattern: |
+                        $RECORD.find(..., :conditions => $X,...)
```
LewisArdern, line 40: should this be => ...,$X,...

ruby/rails/security/injection/tainted-sql-string.yaml
```diff
@@ -1,54 +1,61 @@
...
+                - focus-metavariable: $X
+            - patterns:
+                - pattern: |
+                    "...#{$EXPR}..."
+                - pattern-not-inside: |
+                    $FUNC("...",  "...#{$EXPR}...",...)
```
LewisArdern, line 46: Should this be metavariable-regex with:
```"$SELECT#{$EXPR}...",...)```

File: ruby/rails/security/injection/tainted-sql-string.yaml [hunk 3]
```diff
@@ -1,54 +1,61 @@
...
+            - patterns:
+                - pattern-either:
+                    - pattern: |
+                        $RECORD.where($X,...)
+                    - pattern: |
+                        $RECORD.find(..., :conditions => $X,...)
+                - focus-metavariable: $X
+            - patterns:
+                - pattern: |
+                    "...#{$EXPR}..."
```
LewisArdern: Should this be metavariable-regex with:
```"$SELECT...#{$EXPR}...```

LewisArdern (approved):

ruby/rails/security/injection/tainted-sql-string.yaml
kurt-r2c: the sink is the `conditions` string - it lets you specify a partial raw SQL query. `$Y.find(...)` is SELECT FROM WHERE <conditions> under the hood: `User.find(:first, :conditions => "user_id = '#{params[:user][:user_id]}'")`

ruby/rails/security/injection/tainted-sql-string.yaml
kurt-r2c: Might not matter, but made this change and tests pass so 🤷  might as well

PR #2399, opened 2022-09-16
javascript metadata cleanup

LewisArdern:
This change does the following:

- Adds subcategory for vuln vs audit
- adds new fields based on CWE value
- adds likelihood/severity
- adds confidence
- Adds reference if its missing

How we defined likelihood/severity?

- Impact: in case a vulnerability is exploited, how much damage the attacker could do?
- Likelihood: how likely is for an attacker to find or be able to exploit the vulnerability?

How we define confidence?

- confidence: LOW AUDIT RULE
- confidence: MEDIUM MONITOR MODE
- confidence: HIGH COMMENT MODE


kurt-r2c (approved):

kurt-r2c:
Known failures on legacy stuff with no valid OWASP mapping.

PR #2401, opened 2022-09-19
Updated open redirect in express

LewisArdern:
This update uses unified_mvars to attempt to capture only `req` define in source to reach the 'first' value of `res.redirect` which .... is fairly complicated to do with normal taint mode


0xDC0DE (approved):
lgtm

LewisArdern:
don't merge this yet, working with iago to make it better

LewisArdern:
okay we can merge

PR #2404, opened 2022-09-21
Fix makefile by moving broken shell code to a dedicated script

mjambon:
`make test` was failing silently (error message with exit 0). This is now avoided with a separate bash script. This is a much longer solution than if we could run semgrep simply on `./rules` but at least now it should be more robust than before.

### DIFF
--- a/scripts/run-tests
+++ b/scripts/run-tests
@@ -1,0 +1,95 @@
+#! /usr/bin/env bash
+#
+# Validate rules and run semgrep to check that it's producing the expected
+# results.
+#
+# This was originally part of the makefile but it's gotten too complicated
+# due the need for extra escaping and error codes being ignored.
+#
+set -eu
+
+usage() {
+  cat <<EOF
+Usage: $0 [validate|test]
+
+validate: check only rule validity
+test: run semgrep rules against the target files
+
+Return non-zero if some validation or test doesn't meet the expectations.
+EOF
+}
+
+error() {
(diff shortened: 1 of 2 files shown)


minusworld:
It doesn't look like any of the workflows use Make, so this change shouldn't affect anything in CI.

I am going to approve and merge but if something goes awry, I can revert.

minusworld (approved):

scripts/run-tests
```diff
@@ -0,0 +1,95 @@
...
+  echo "Error: $*" >&2
+  exit 1
+}
+
+# Use the SEMGREP environment variable to specify a non-standard semgrep
+# command. This is useful for calling a development version of semgrep
+# e.g.
+#   PIPENV_PIPFILE=~/semgrep/cli/Pipfile SEMGREP='pipenv run semgrep' make test
+#
+if [[ -z "$SEMGREP" ]]; then
```
aryx, line 32: @mjambon does this script work?
I've just tried make test in semgrep-rules and got:
```
(cli) [pad@thinkstation semgrep-rules (develop)]$ make test
make validate
make[1]: Entering directory '/home/pad/github/semgrep-rules'
./scripts/run-tests validate
./scripts/run-tests: line 32: SEMGREP: unbound variable
make[1]: *** [Makefile:13: validate] Error 1
make[1]: Leaving directory '/home/pad/github/semgrep-rules'
make: *** [Makefile:8: test] Error 2
```

Issue #2380, opened 2022-09-09
plz update Github Actions workflow for me


mjambon:
[I don't have the permission to push changes to `.github/workflows`: 
```
$ git push origin mj-use-makefile-in-ci
...
 ! [remote rejected]   mj-use-makefile-in-ci -> mj-use-makefile-in-ci (refusing to allow a Personal Access Token to create or update workflow `.github/workflows/semgrep-rules-test-develop.yml` without `workflow` scope)
```
]

After both https://github.com/returntocorp/semgrep/pull/6070 and https://github.com/returntocorp/semgrep-rules/pull/2379 are merged, we should make these changes in `.github/workflows/semgrep-rules-test-develop.yml`:

```
    # Validate rules and check that they work against the sample targets.
    # We use the same test commands as semgrep's CI to avoid discrepancies.
    - name: validate rules
      run: |
        export SEMGREP="docker run --rm -w /src -v ${GITHUB_WORKSPACE}/semgrep-rules:/src returntocorp/semgrep:develop semgrep"
        make -C "$GITHUB_WORKSPACE"/semgrep-rules validate
    - name: test with semgrep develop branch
      run: |
        export SEMGREP="docker run --rm -w /src -v ${GITHUB_WORKSPACE}/semgrep-rules:/src returntocorp/semgrep:develop semgrep"
        make -C "$GITHUB_WORKSPACE"/semgrep-rules test-only
```

PR #2405, opened 2022-09-21
Update workflow file to use make

minusworld:


minusworld:
https://github.com/returntocorp/semgrep-rules/issues/2380

mjambon:
Thanks @minusworld, it's failing as expected based on what @aryx reported earlier today.

mjambon:
Actually I was wrong:
* it's my script that's broken. The fix is there: https://github.com/returntocorp/semgrep-rules/pull/2406
* I don't get test failures after that when I run `make test` locally... which is good.

mjambon (approved):

mjambon:
Thank you, @minusworld !

PR #2412, opened 2022-09-23
Update ruby-eval to catch other access to params + cookies

jbergler:
In #2332 params + cookies were added to this rule

params often refers to ActionController::Parameter
cookies often refers to ActionDispatch::Cookie

The user input residing in these classes can be accessed in more ways
than just via [...] such as params.dig(...) or cookies.delete(...)

This updates the rule and test cases to catch this.


0xDC0DE (approved):
That looks good to me, thanks for your contribution!

PR #2413, opened 2022-09-23
Add python metadata 

LewisArdern:
This change does the following:

- Adds subcategory for vuln vs audit
- adds new fields based on CWE value
- adds likelihood/severity
- adds confidence

How we defined likelihood/severity?

Impact: in case a vulnerability is exploited, how much damage the attacker could do?
Likelihood: how likely is for an attacker to find or be able to exploit the vulnerability?

How we define confidence?

- confidence: LOW AUDIT RULE
- confidence: MEDIUM MONITOR MODE
- confidence: HIGH COMMENT MODE


0xDC0DE (approved):
I did a spot check of 10 random rules and they all look good to me!

PR #2414, opened 2022-09-24
add golang metadata

LewisArdern:


go/lang/security/audit/crypto/use_of_weak_crypto.yaml
```diff
@@ -1,73 +1,26 @@
...
-    metadata:
-      owasp: "A9: Using Components with Known Vulnerabilities"
-      cwe: "CWE-327: Use of a Broken or Risky Cryptographic Algorithm"
-      source-rule-url: https://github.com/securego/gosec#available-rules
-      category: security
-      technology:
-        - go
-    pattern-either:
-      - pattern: |
-          des.NewTripleDESCipher(...)
```
colleend, line 56: did we mean to delete these patterns/other rules in this file?

go/lang/security/audit/crypto/use_of_weak_crypto_.yaml
```diff
@@ -0,0 +1,26 @@
...
+  languages: [go]
+  severity: WARNING
+  metadata:
+    owasp:
+    - A03:2017 - Sensitive Data Exposure
+    - A02:2021 - Cryptographic Failures
+    cwe:
+    - 'CWE-327: Use of a Broken or Risky Cryptographic Algorithm'
+    source-rule-url: https://github.com/securego/gosec#available-rules
+    category: security
```
colleend: also I think this rule is a copy of the one above 😓

kurt-r2c (approved):

kurt-r2c (dismissed):
use_of_weak_crypto is an old rule that uses a multi-rule file - the script mangled the output and didn't export the rest of them.

Script probably ought to break these out into their own files if they exist.

Props to @colleend for noticing the issue

go/lang/security/audit/crypto/use_of_weak_crypto_.yaml
LewisArdern: i created the new rules because there were missing checks so i create new rules.. :)

LewisArdern: welp!

LewisArdern:
@kurt-r2c cant split them into their own file because it'll introduce new registry items which people may have been expecting it differently. Even if you delete additional ruleids they will still exist in the registry. I didn't spot the deletion so i added them all as new rules! i will fix this up.

LewisArdern:
I updated the script to handle the multi-line rules, deleting my contributions and adding in the old rule

LewisArdern:
@colleend does this look correct now?

colleend (approved):
yes! looks good to me now :D

PR #2415, opened 2022-09-26
update xss direct-response-write

LewisArdern:
negative feedback said that they had 'safe xss' so included sanitizers which may capture this.


inkz (approved):
lgtm

PR #2419, opened 2022-09-28
Dockerfile metadata

LewisArdern:


go/gorilla/security/audit/handler-assignment-from-multiple-sources.yaml
```diff
@@ -0,0 +1,52 @@
...
+    category: security
+    technology:
+    - gorilla
+    confidence: MEDIUM
+    license: Commons Clause License Condition v1.0[LGPL-2.1-only]
+    references:
+    - https://cwe.mitre.org/data/definitions/289.html
+    subcategory:
+    - vuln
+    impact: MEDIUM
```
minusworld, line 17: I wrote this rule forever ago based on a haphazard conversation. I'm changing likelihood to "LOW" and subcat to "audit" to match my belief about the rule

minusworld (approved):

PR #2425, opened 2022-09-28
Add exception when redirecting to same page in Flask

luisfontes19:
Added some exceptions when redirecting to the same page in python/flask. 

This pattern is common for removing query params for example


LewisArdern:
Awesome! Thank you for your contribution

LewisArdern (approved):

PR #2444, opened 2022-10-03
add generic metadata

LewisArdern:
accidentally 'rewrote' non security rules but has no impact on the correctness rule.


inkz (approved):
lgtm

PR #2451, opened 2022-10-06
Fix wrong use of todook:

aryx:
Those todook: are used in other files to indicate that semgrep
currently reports a match, but at some point should not
and should become an ok:.
In the 2 files in this PR it is used for something else, which
then introduces some regressions in semgrep-core rules regressions
infra.

test plan:
make test in semgrep-core now works

### DIFF
--- a/java/jax-rs/security/insecure-resteasy.java
+++ b/java/jax-rs/security/insecure-resteasy.java
@@ -6,64 +6,63 @@ import javax.ws.rs.core.*;
 
 @Path("/")
 public class PoC_resource {
-    @POST
-    @Path("/concat")
-    @Produces(MediaType.APPLICATION_JSON)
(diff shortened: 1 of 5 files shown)


p4p3r (approved):
LGTM

java/jax-rs/security/insecure-resteasy.java
```diff
@@ -6,64 +6,63 @@
...
-        result.put("Result", elements.size());
+    return result;
+  }
 
-        return result;
-    }
+  @POST
+  @Path("/count")
+  @Produces(MediaType.APPLICATION_JSON)
+  // todook: insecure-resteasy-deserialization
```
aryx, line 35: hmm this one should be removed! semgrep is not currently finding it.

Issue #2457, opened 2022-10-10
False positive when array is passed to preg_replace


Sjord:
[semgrep-rules/preg-replace-eval.yaml at develop · returntocorp/semgrep-rules](https://github.com/returntocorp/semgrep-rules/blob/develop/php/lang/security/preg-replace-eval.yaml)

[`preg_replace`](https://www.php.net/preg_replace) can also take an array, and then this rule triggers, even when static content is used.

preg.php:
```
<?php
echo preg_replace(array("/hello/"), array("world"), "hello world");
```

semgrep output:
```
$ semgrep -c 'r/php.lang.security.preg-replace-eval' preg.php
Semgrep rule registry URL is https://semgrep.dev/registry.
Scanning 1 file.

Blocking Findings:

  preg.php
     php.lang.security.preg-replace-eval.preg-replace-eval
        Calling preg_replace with user input in the pattern can lead to arbitrary code execution.
        The eval modifier (`/e`) evaluates the replacement argument as code.
        Details: https://sg.run/0Qzw

          2┆ echo preg_replace(array("/hello/"), array("world"), "hello world");

Blocking Rules Fired:
   php.lang.security.preg-replace-eval.preg-replace-eval


Ran 1 rule on 1 file: 1 finding.
```

p4p3r:
Hey @Sjord, thanks a lot for reporting the FP!
The rule has been updated to address your feedback.
Please reopen if you're still running into troubles!

PR #2458, opened 2022-10-10
fix false positives in preg-replace-eval rule

0xDC0DE:
Addresses https://github.com/returntocorp/semgrep-rules/issues/2457


p4p3r (approved):
LGTM

PR #2462, opened 2022-10-10
Fix copy/paste error in message of OCaml physical vs structural rule

nmote:
This rule is about the equality operator, not inequality.


kurt-r2c (approved):

kurt-r2c:
@nmote maybe add a link to OCAML structural equality docs so tests pass, otherwise LGTM

aryx:
actually the validate CI check requires a references: now.

PR #2466, opened 2022-10-12
Updating semgrep-scanner trigger to be diff-aware

0xDC0DE:


p4p3r (approved):
LGTM

inkz (approved):
LGTM

but to be paranoic I would change all code patterns like
```yaml
run: |
  foobar ${{ env.SMTH }}
```

to 

```yaml
env:
  VAL1: ${{ value1 }}
run: |
  foobar $VAL1
```
especially when we have most of the vars in`env` already

I think we might have a chance to leak ARGO token with this

```
curl -X POST https://argo.corp.r2c.dev/api/v1/events/security-research/initiate-scan-argo -H "Authorization: ${{ secrets.ARGO_WORKFLOWS_TOKEN }}" -d "{\"branch\" : \"${{github.head_ref}}\", \"repo\" : \"${{github.event.repository.name}}\", \"commit\" : \"${{github.event.pull_request.head.sha}}\", \"changed_files\" : \"${{ env.changed_files }}\" , \"langs\" : \"${{ env.changed_langs }}\"}"
```

by adding a file with some crazy name like `"} --do something else ".txt`

Issue #2465, opened 2022-10-12
False positive report on TLS version for google_compute_ssl_policy by (azure) appservice-use-secure-tls-policy


ldsnia:
**Describe the bug**

google_compute_ssl_policy triggers https://semgrep.dev/r?q=terraform.azure.security.appservice.appservice-use-secure-tls-policy.appservice-use-secure-tls-policy an azure rule.

**To Reproduce**

Add 
```
# not an azure resource

resource "google_compute_ssl_policy" "modern_TLS_policy" {
  min_tls_version = "TLS_1_2"
  name            = "modern-ssl-policy"
  profile         = "MODERN"
}
```
to `terraform/azure/security/appservice/appservice-use-secure-tls-policy.tf`

Run the tests:
`make test .`

```
[...]
        ✖ appservice-use-secure-tls-policy
        missed lines: [], incorrect lines: [24]
        test file path: $GIT_REPO_PATH/semgrep-rules/terraform/azure/security/appservice/appservice-use-secure-tls-policy.tf
[...]
```
 

**Expected behavior**

Test should pass.
A GCP resource should not trigger a rule written for Azure resources.

**Priority**
How important is this to you?
- [ ] P0: blocking me from making progress
- [ ] P1: this will block me in the near future
- [X] P2: annoying but not blocking me

**Screenshots**

**Desktop (please complete the following information):**
 - OS: [e.g. iOS] Linux 5.15.0-1018-gcp #24~20.04.1-Ubuntu SMP Mon Sep 12 06:14:01 UTC 2022 x86_64 x86_64 x86_64 GNU/Linux

**Smartphone (please complete the following information):**

**Additional context**

It seems that the parser and the rules do not take into account the type of resources it's running against. This change could be a larger change. Adding the specific string to the azure pattern is a quick but dirty fix.

0xDC0DE:
Thanks for reporting this report @ldsnia, I have updated the rule. Once a coworker reviews the changes you will no longer run into this false positive.

PR #2468, opened 2022-10-12
Do not detect GCP tls configuration in Azure detection rules.

ldsnia:
GCP configuration should not be impacted by Azure detection rules.

See issue https://github.com/returntocorp/semgrep-rules/issues/2465


0xDC0DE:
I have take a different approach to resolve this False Positive, here: https://github.com/returntocorp/semgrep-rules/pull/2473/files

If this satisfies your requirements, then you can close this PR. If not, let me know and we can work together to improve the rule!

ldsnia:
Your approach is better. Thank you.

PR #2470, opened 2022-10-13
Add rule for redirecting to current request URI in PHP

Sjord:
The request URI can start with two slashes, in which case it will redirect to another domain.

Whether this is actually possible depends for a great deal if a redirecting page can be requested on the target server with a name that is also a valid domain.


0xDC0DE (approved):
That looks good to me! Thanks for contributing to the rules!

Issue #2467, opened 2022-10-12
False positive in javascript.browser.security.raw-html-join.raw-html-join when joining static strings


Sjord:
Joining static strings can't be vulnerable to XSS, but semgrep still warns about it:

```javascript
console.log(['I <3 Semgrep', 'Hello world'].join("\n"));
```

```
$ semgrep -c 'r/javascript.browser.security.raw-html-join.raw-html-join' test.js
Semgrep rule registry URL is https://semgrep.dev/registry.
Scanning 1 file.

Blocking Findings:

  test.js
     javascript.browser.security.raw-html-join.raw-html-join
        User controlled data in a HTML string may result in XSS
        Details: https://sg.run/qqQO

          1┆ console.log(['I <3 Semgrep', 'Hello world'].join("\n"));

Blocking Rules Fired:
   javascript.browser.security.raw-html-join.raw-html-join


Ran 1 rule on 1 file: 1 finding.
```

0xDC0DE:
Hi @Sjord, thanks for another bug report! I'll take a look right away!

PR #2474, opened 2022-10-14
Add FP test case and update raw-html-join rule

0xDC0DE:
Addresses: https://github.com/returntocorp/semgrep-rules/issues/2467


colleend (approved):
thank you!!!

PR #2482, opened 2022-10-19
Add rules for usage of bad crypto schemes in .NET

mtausig:
Add 2 rules detecting the usage of weak or deprecated cryptographic schemes in .NET

### DIFF
--- a/csharp/dotnet/security/use_deprecated_cipher_algorithm.cs
+++ b/csharp/dotnet/security/use_deprecated_cipher_algorithm.cs
@@ -1,0 +1,55 @@
+using System;
+using System.Security.Cryptography;
+					
+public class Encryption
+{
+	public void CreateAes1() {
+		//ok: use_deprecated_cipher_algorithm
+		var key = Aes.Create();
+	}
+	
+	public void CreateAes2() {
+		//ok: use_deprecated_cipher_algorithm
+		var key = Aes.Create("ImplementationName");
+	}
+
+	public void CreateRijndael1() {
+		//ok: use_deprecated_cipher_algorithm
+		var key = Rijndael.Create();
+	}
+	
+	public void CreateRijndael2() {
+		//ok: use_deprecated_cipher_algorithm
+		var key = Rijndael.Create("ImplementationName");
+	}
+
+	public void CreateDES1() {
+		//ruleid: use_deprecated_cipher_algorithm
+		var key = DES.Create();
+	}
+	
+	public void CreateDES2() {
+		//ruleid: use_deprecated_cipher_algorithm
+		var key = DES.Create("ImplementationName");
+	}
+
+	public void CreateTripleDES1() {
+		//ok: use_deprecated_cipher_algorithm
+		var key = TripleDES.Create();
+	}
+	
+	public void CreateTripleDES2() {
+		//ok: use_deprecated_cipher_algorithm
+		var key = TripleDES.Create("ImplementationName");
+	}
+
+	public void CreateRC21() {
+		//ruleid: use_deprecated_cipher_algorithm
+		var key = RC2.Create();
+	}
+	
+	public void CreateRC22() {
+		//ruleid: use_deprecated_cipher_algorithm
+		var key = RC2.Create("ImplementationName");
+	}
+}
--- a/csharp/dotnet/security/use_deprecated_cipher_algorithm.yaml
+++ b/csharp/dotnet/security/use_deprecated_cipher_algorithm.yaml
@@ -7,7 +7,9 @@ rules:
           pattern-either:
             - pattern: DES
             - pattern: RC2
-    message: Usage of deprecated cipher algorithm detected
+    message: |
+      Usage of deprecated cipher algorithm detected.
+      Use Aes or ChaCha20Poly1305 instead.
     languages:
       - csharp
     severity: ERROR
@@ -20,5 +22,7 @@ rules:
         - .net
       category: security
       references:
+        - https://learn.microsoft.com/en-gb/dotnet/api/system.security.cryptography.des?view=net-6.0#remarks
         - https://learn.microsoft.com/en-gb/dotnet/api/system.security.cryptography.rc2?view=net-6.0#remarks
-        - https://learn.microsoft.com/en-gb/dotnet/api/system.security.cryptography.rc2?view=net-6.0#remarks
+        - https://learn.microsoft.com/en-gb/dotnet/api/system.security.cryptography.aes?view=net-6.0
+        - https://learn.microsoft.com/en-gb/dotnet/api/system.security.cryptography.chacha20poly1305?view=net-6.0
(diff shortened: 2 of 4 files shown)


csharp/dotnet/security/use_deprecated_cipher_algorithm.cs
```diff
@@ -0,0 +1,55 @@
...
+	}
+
+	public void CreateTripleDES1() {
+		//ok: use_deprecated_cipher_algorithm
+		var key = TripleDES.Create();
+	}
+	
+	public void CreateTripleDES2() {
+		//ok: use_deprecated_cipher_algorithm
+		var key = TripleDES.Create("ImplementationName");
```
0xDC0DE, line 43: Triple DES is also not a recommended algorithm to my knowledge. 
E.g. the OWASP cheat sheet recommends only the use of AES for symmetric and ECC/RSA for asymmetric encryption.

csharp/dotnet/security/use_deprecated_cipher_algorithm.cs
```diff
@@ -0,0 +1,55 @@
...
+	}
+	
+	public void CreateAes2() {
+		//ok: use_deprecated_cipher_algorithm
+		var key = Aes.Create("ImplementationName");
+	}
+
+	public void CreateRijndael1() {
+		//ok: use_deprecated_cipher_algorithm
+		var key = Rijndael.Create();
```
0xDC0DE, line 18: Rijndael seems to be deprecated based on the docs: https://learn.microsoft.com/en-us/dotnet/api/system.security.cryptography.rijndael?view=net-6.0

csharp/dotnet/security/use_deprecated_cipher_algorithm.cs
0xDC0DE: https://cheatsheetseries.owasp.org/cheatsheets/Cryptographic_Storage_Cheat_Sheet.html#algorithms

csharp/dotnet/security/use_ecb_mode.yaml
```diff
@@ -0,0 +1,28 @@
...
+      - metavariable-pattern:
+          metavariable: $KEYTYPE
+          pattern-either:
+            - pattern: SymmetricAlgorithm
+            - pattern: Aes
+            - pattern: Rijndael
+            - pattern: DES
+            - pattern: TripleDES
+            - pattern: RC2
+    message: Usage of ECB mode detected
```
0xDC0DE: Could you expand the message to also include a recommendation of what to use instead?
You could, for example, reuse or rephrase this section from the OWASP cheatsheet:

> Where available, authenticated modes should always be used. These provide guarantees of the integrity and authenticity of the data, as well as confidentiality. The most commonly used authenticated modes are [GCM](https://en.wikipedia.org/wiki/Galois/Counter_Mode) and [CCM](https://en.wikipedia.org/wiki/CCM_mode), which should be used as a first preference.
> 
> If GCM or CCM are not available, then [CTR](https://en.wikipedia.org/wiki/Block_cipher_mode_of_operation#Counter_%28CTR%29) mode or [CBC](https://en.wikipedia.org/wiki/Block_cipher_mode_of_operation#Cipher_Block_Chaining_%28CBC%29) mode should be used. As these do not provide any guarantees about the authenticity of the data, separate authentication should be implemented,

https://cheatsheetseries.owasp.org/cheatsheets/Cryptographic_Storage_Cheat_Sheet.html#cipher-modes

csharp/dotnet/security/use_deprecated_cipher_algorithm.yaml
```diff
@@ -0,0 +1,24 @@
+rules:
+  - id: use_deprecated_cipher_algorithm
+    patterns:
+      - pattern: $KEYTYPE.Create(...);
+      - metavariable-pattern:
+          metavariable: $KEYTYPE
+          pattern-either:
+            - pattern: DES
+            - pattern: RC2
+    message: Usage of deprecated cipher algorithm detected
```
0xDC0DE: Please include a recommendation for which algorithm to use instead (preferably with mention of the exact API to use). E.g.

> Use the AES algorithm instead. This can be done with the `Aes.create()` function.

0xDC0DE (commented):
I left some comments, if you want help fixing these remarks, give us a shout!

csharp/dotnet/security/use_deprecated_cipher_algorithm.cs
mtausig: I agree, that 3DES is not recommended to be used anymore but it is considered to be still kind of acceptable.
So flagging would be fine, but that would require a separate rule with a lower severity rating (3DES should only be `INFO` imo).

csharp/dotnet/security/use_deprecated_cipher_algorithm.cs
mtausig: Rijndael is basically AES, so this is not a relevant security risk either. The same holds, as for 3DES.

csharp/dotnet/security/use_deprecated_cipher_algorithm.yaml
mtausig: 62088b5a9baa5943821bb58cc7c93c8991680668

csharp/dotnet/security/use_ecb_mode.yaml
mtausig: a76330075690b0e2bf93264a5762b6c3436da5a5

csharp/dotnet/security/use_deprecated_cipher_algorithm.cs
0xDC0DE: Ok, that makes sense to me!

csharp/dotnet/security/use_deprecated_cipher_algorithm.cs
0xDC0DE: 👍

0xDC0DE (approved):

csharp/dotnet/security/use_deprecated_cipher_algorithm.yaml
```diff
@@ -0,0 +1,28 @@
+rules:
+  - id: use_deprecated_cipher_algorithm
+    patterns:
+      - pattern: $KEYTYPE.Create(...);
+      - metavariable-pattern:
+          metavariable: $KEYTYPE
+          pattern-either:
+            - pattern: DES
+            - pattern: RC2
+    message: |
```
kurt-r2c: use chomped folded blocks for message:

```yaml
message: >-
    ...
```

csharp/dotnet/security/use_ecb_mode.yaml
```diff
@@ -0,0 +1,35 @@
...
+      - metavariable-pattern:
+          metavariable: $KEYTYPE
+          pattern-either:
+            - pattern: SymmetricAlgorithm
+            - pattern: Aes
+            - pattern: Rijndael
+            - pattern: DES
+            - pattern: TripleDES
+            - pattern: RC2
+    message: |
```
kurt-r2c: use chomped folded blocks for `message`

kurt-r2c (changes requested):

LewisArdern:
@kurt-r2c were your updates resolved?

kurt-r2c (approved):

PR #2483, opened 2022-10-19
Normalize CWE-185 (bad regex) to CWE-1333 (reDOS)

fproulx-boostsecurity:
Some rules are mislabeled as CWE-185, while they really are looking for CWE-1333.


0xDC0DE:
Thanks for your contribution, this change looks correct to me! Approved 👍

0xDC0DE (approved):

PR #2486, opened 2022-10-20
feat: add yii2 rules

aldok10:


php/yii/auth-bypass-yii-framework-apps3c.php
```diff
@@ -0,0 +1,39 @@
+<?php
+
+use Yii;
+
+class Module extends \yii\base\Module
+{
+
+    # Wrong
```
colleend, line 8: Can use `// ruleid:` above lines you expect to match and `// ok:` above lines you expect to not match (correct lines) -- we usually have at least one `//ruleid` and one `//ok` in every test file!

php/yii/auth-bypass-yii-framework-apps3c.yaml
```diff
@@ -0,0 +1,30 @@
...
+  severity: ERROR
+  message: >-
+      Authentication and authorization checks are often implemented in 
+      "beforeAction" functions in Yii framework. When an authentication
+      or authorization check is failed, the user is usually redirected to 
+      a login page or to an exception one. The "beforeAction must return 
+      "false" also if the user is redirected, otherwise the code protected 
+      by the "beforeAction" function is executed before the redirection, 
+      posing a secrious security risk.
+
```
colleend, line 14: would be great to add OWASP + CWE numbers!

colleend (changes requested):
Thank you for your contribution! We really appreciate it 😄 

One thing we do indeed do is annotate everything we expect to match with `ruleid:` and lines of code that we don't want to match with `ok:`. We also include CWE and OWASP information for all security rules. Feel free to let us know if you want us to do this for you!

Issue #2487, opened 2022-10-20
Usage of ssl._create_unverified_context not detected


Sjord:
Code that disables certificate validation is not detected by a semgrep rule.

As suggested in [PEP 476](https://peps.python.org/pep-0476/#opting-out) and [this article](https://www.howtouselinux.com/post/ssl-certificate_verify_failed-in-python), certificate validation can be disabled by overwriting `ssl._create_default_https_context`:

```
import ssl
ssl._create_default_https_context = ssl._create_unverified_context
urllib2.urlopen(“https://google.com”).read()
```

I would expect this to be detected by [unverified-ssl-context](https://github.com/returntocorp/semgrep-rules/blob/develop/python/lang/security/unverified-ssl-context.yaml), but it isn't. Apparently because in the above code, `ssl._create_unverified_context` is used in an assignment instead of in a function call.

LewisArdern:
Looks like this was fixed @Sjord thanks again for your thoroughness!

PR #2489, opened 2022-10-21
adding case to unverified ssl context

colleend:
For https://github.com/returntocorp/semgrep-rules/issues/2487.

We want to detect cases where default https context is set to unverified.


LewisArdern (approved):
lgtm - did u test the autofix?

colleend:
I did manually yes!!

PR #2494, opened 2022-10-22
Add rule detecting insecure random number generator in csharp

mtausig:


kurt-r2c:
@mtausig an unrelated test is failing - recommend pulling latest into your feature branch to make diff computation more accurate for tests.

mtausig:
@kurt-r2c Squashed & rebased

kurt-r2c (approved):

PR #2497, opened 2022-10-24
Missing integrity

### LINKED ISSUE #2441: [Rule] html.security.audit.missing-integrity.missing-integrity doesn't work as expected
**Describe the bug**

Rule [html.security.audit.missing-integrity.missing-integrity](https://semgrep.dev/r?q=html.security.audit.missing-integrity.missing-integrity) produces false positives with multiple consecutive `<link>` tags.

**To Reproduce**

The following code will trigger the rule even if it shouldn't:

```
<link rel="stylesheet" href="https://someurl/style.css" integrity="sha256-somehashdigest">
<link rel="stylesheet" href="./css/mystyle.css">
```

**Expected behavior**
The previous code snippet shouldn't trigger the rule, beacause the first link tag includes the `integrity` attribute and the second link tag is using a local `href` attribute (the rule is only looking for http(s) `href` attributes). 

**Priority**
How important is this to you?

**Additional Context**

If you try to run the rule against each single link tag, it does not trigger (as it should be).

LewisArdern:
Fixes https://github.com/returntocorp/semgrep-rules/issues/2441

### DIFF
--- a/html/security/audit/missing-integrity.html
+++ b/html/security/audit/missing-integrity.html
@@ -6,7 +6,9 @@
     <meta name="viewport" content="width=device-width, initial-scale=1.0">
     <title>Document</title>
     <!-- ruleid: missing-integrity -->
-    <script src="https://somewhere-external.com/something-external.js"></script>
+    <script src='https://somewhere-external.com/something-external.js'></script>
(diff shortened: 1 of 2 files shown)


colleend (approved):
lgtm 👍

PR #2505, opened 2022-10-27
Enforce new metadata fields 

LewisArdern:
Enforces the new metadata requirements now that the new documents are updated https://semgrep.dev/docs/contributing/contributing-to-semgrep-rules-repository/

1. Adds confidence and confidence validation check
2. Adds likelihood and likelihood validation check
3. Adds impact and impact validation check
4. Adds subcategory and subcategory validation check 

Also fixes the current failing workflows:
* Uses ubuntu-20.04


enncoded (approved):

.github/workflows/semgrep-rules-test.yml
```diff
@@ -11,7 +11,7 @@ on:
 jobs:
   test-latest:
     name: rules-test-latest
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-20.0
```
underyx, line 14: you have a typo here so I think this job cannot run now?

underyx: should probably be `runs-on: ubuntu-20.04`

PR #2529, opened 2022-11-03
move solidity rule, fix file names

kurt-r2c:
fixes for failing job https://github.com/returntocorp/semgrep-rules/actions/runs/3389823123/jobs/5633321750


colleend (approved):
lgtm

PR #2536, opened 2022-11-04
support shebang contexts for finding dangerous execs

kurt-r2c:
Go's `exec.Command` and `exec.CommandContext` support shebang-style command invocation which the existing command injection rule did not correctly handle.


inkz (approved):
lgtm

PR #2537, opened 2022-11-06
Update `command-injection-formatted-runtime-call` rule

inkz:
Small update for `command-injection-formatted-runtime-call` rule


p4p3r (approved):
LGTM

PR #2539, opened 2022-11-07
ignore cookies created by Spring's ResponseCookie builder

objectified:
When someone uses Spring's ResponseCookie builder class to create cookies, the "secure" and "httpOnly" attributes are set by the builder, instead of on the actual cookie, so both the interface and the fix will be different. This PR makes sure that cookies created by this builder are ignored. I will write a Spring specific rule hopefully soon to catch cookie misconfigurations when using the ResponseCookie builder in Spring.


inkz:
thank you :pray:

inkz (approved):

Issue #2426, opened 2022-09-28
[Rule] Go: use-filepath-join is not a very good rule


itaranto:
**Describe the bug**

The [use-filepath-join](https://github.com/returntocorp/semgrep-rules/blob/develop/go/lang/correctness/use-filepath-join.yaml) suggests to *always* use `filepath.Join` which is not entirely correct, `filepath` is meant to be used with filesystem paths and `path` is meant to be used with `/` separated paths (like URLs).

**To Reproduce**

```go
package main

import (
	"fmt"
	"net/url"
	"path"
)

func main() {
	url, err := url.Parse("http://foo:666/bar")
	if err != nil {
		panic(err)
	}

	fmt.Println(path.Join(url.Path, "baz"))
}
```

**Expected behavior**

I'm not sure what would be the expected behavior here, I believe it's very difficult to tell if `path.Join` is being used on a filesystem path or not. I'd rather not have this rule at all since it can be misleading.

**Priority**

How important is this to you?
- [ ] P0: blocking me from making progress
- [ ] P1: this will block me in the near future
- [x] P2: annoying but not blocking me

LewisArdern:
Thanks for your report! Maybe we could do something like this? https://semgrep.dev/playground/s/lewisardern:use-filepath-join-copy?editorMode=advanced in general I agree this rule is not very nice -- but it definitely serves a purpose in file land. 

We could either do it:
1. Try to just remove anything with a HTTP request 
2. Only report when its within another `filepath.$SOMETHING(....,<... file.Join(...)...>,...)` or `file.$SOMETHING` etc.

dgutson:
CC @qequ

itaranto:
> Thanks for your report! Maybe we could do something like this? https://semgrep.dev/playground/s/lewisardern:use-filepath-join-copy?editorMode=advanced in general I agree this rule is not very nice -- but it definitely serves a purpose in file land.
> 
> We could either do it:
> 
>     1. Try to just remove anything with a HTTP request
> 
>     2. Only report when its within another `filepath.$SOMETHING(....,<... file.Join(...)...>,...)` or `file.$SOMETHING` etc.

Yes, something like this could work. My concern is that paths are just strings, I don't see a reliable way to identify paths, i.e. if they are FS paths or `/` separated paths.

LewisArdern:
[...]

PR #2540, opened 2022-11-07
Updated `use-filepath-join`

enncoded:
Updates the rule to be less noisy, w.r.t https://github.com/returntocorp/semgrep-rules/issues/2426

### DIFF
--- a/go/lang/correctness/use-filepath-join.yaml
+++ b/go/lang/correctness/use-filepath-join.yaml
@@ -6,73 +6,33 @@ rules:
     message: "`path.Join(...)` always joins using a forward slash. This may cause
       issues on Windows or other systems using a different delimiter. Use
       `filepath.Join(...)` instead which uses OS-specific path separators."
-    patterns:
+    mode: taint
+    pattern-sources:
+      - patterns:
+          - pattern: |
+              ($STR : string)
+          - pattern-not: |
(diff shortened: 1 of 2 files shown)


go/lang/correctness/use-filepath-join.yaml
```diff
@@ -1,16 +1,78 @@
...
-    - go
+  - id: use-filepath-join
+    languages:
+      - go
+    severity: WARNING
+    message: "`path.Join(...)` always joins using a forward slash. This may cause
+      issues on Windows or other systems using a different delimiter. Use
+      `filepath.Join(...)` instead which uses OS-specific path separators."
+    patterns:
+      - pattern: path.Join(...)
```
kurt-r2c: Is there any reason we can't do taint with `($INPUT : string)` as a source and `path.Join(...)` as a sink?

This many conditions just seems like "code smell" for a taint approach

dgutson: @kurt-r2c I think that semgrep shall support semgrep rules language :)

enncoded: Changed to a taint rule, with functions that return path strings in stdlib as sources.

kurt-r2c (approved):

Issue #2078, opened 2022-05-07
False positive with subprocess.run()


djmattyg007:
**Describe the bug**
The rule `python.lang.security.audit.dangerous-subprocess-use.dangerous-subprocess-use` has a few shortcomings:

- It doesn't understand passing tuples (as opposed to lists)
- It doesn't understand `sys.executable` as the first argument

It's also possible that it doesn't understand passing a non-user-controlled `Path` object (cast to a string) as the second value in the tuple supplied to `subprocess.run()`.

**To Reproduce**

```
import subprocess
from pathlib import Path

_CMD_PATH = Path(__file__).parent.parent / "plugin.py"

def run_script():
    # None of these values (except the environment variables) are user-controlled. This is fine.
    result = subprocess.run(
        (sys.executable, str(_CMD_PATH)),
        stdout=subprocess.PIPE,
        stderr=subprocess.STDOUT,
        text=True,
        timeout=10,
        check=True,
    )
```

**Expected behavior**
There's no reason to warn in any capacity on the above code, because there are no user-controlled values being supplied. There is no reason to manually quote anything either, because python takes care of escaping everything properly for me.

**Priority**
How important is this to you?
- [x] P0: blocking me from making progress
- [ ] P1: this will block me in the near future
- [ ] P2: annoying but not blocking me

While there's no ability to disable individual rules, and the need to turn on `--error` and `--strict` so that CI builds actually fail, any rule with a false positive is blocking.

**Additional Context**

0xDC0DE:
Hi @djmattyg007 ! 
Thanks for your bug report! 

The rule you are referencing here is an `audit` rule. In the README you will find some information about these rules:
> If a security rule is discouraging the use of a bad pattern (such as formatted SQL strings), we recommend appending audit to your namespace. This distinguishes it from a security rule that is specifically aiming to detect a vulnerability.

So, I understand that this is flagging code for you that is not vulnerable, but this is kind of the intention of this type of rule. Generally, there will be a tradeoff for rules. The more accurate they are (and hence the lower the false positive rate), the more likely they are also going to miss some cases (false negatives). The idea behind audit rules is that they can be used by security professionals to audit a repository, and easily find interesting parts in the code to inspect closer.

We are currently working on creating more accurate, vulnerability detecting rules. These rules will make use of our [taint mode](https://semgrep.dev/docs/writing-rules/data-flow/taint-mode/). For Python we have Django and Flask rules on our roadmap for this quarter.

[...]

PR #2558, opened 2022-11-15
Updated `dangerous-subprocess-use-audit` to ignore tuples

### LINKED ISSUE #2554: False positives with subprocess funcs in Python
**Describe the bug**
Python rules concerning dangerous subprocess calls all lack the check for when the arguments to subprocess.$FUNCs are given in tuples and not lists, which results in false positives in rulesets not in taint mode, f.e [this](https://semgrep.dev/playground/r/python.lang.security.audit.dangerous-subprocess-use-audit.dangerous-subprocess-use-audit) audit rule. While i understand that audit rules are meant to be noisy, i dont think this addition will increase the number of false negatives, but will definitely decrease the number of false positives. From what I understand its a matter of adding one pattern-not with a tuple as an argument to a subprocess.$FUNC(). 

**To Reproduce**
`subprocess.call(("echo", "a", ";", "rm", "-rf", "/"))`

**Expected behavior**
checking for subprocess.call((....)) should not yield any different results than subprocess.call([...]).
In the above example 

**Priority**
How important is this to you?
- [x] P0: blocking me from making progress

**Additional Context**
https://github.com/returntocorp/semgrep-rules/issues/2078

enncoded:
Fixes https://github.com/returntocorp/semgrep-rules/issues/2554 and https://github.com/returntocorp/semgrep-rules/issues/2078

### DIFF
--- a/python/lang/security/audit/dangerous-subprocess-use-audit.py
+++ b/python/lang/security/audit/dangerous-subprocess-use-audit.py
@@ -10,6 +10,9 @@ subprocess.call("echo 'hello'")
 subprocess.call(["echo", "a", ";", "rm", "-rf", "/"])
 
 # ok:dangerous-subprocess-use-audit
+subprocess.call(("echo", "a", ";", "rm", "-rf", "/"))
+
+# ok:dangerous-subprocess-use-audit
 raise subprocess.CalledProcessError("{}".format("foo"))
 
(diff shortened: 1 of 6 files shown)


enncoded:
oops pretend that commit message says tuple and not tilde 😅

colleend (approved):
lgtm!

Issue #2560, opened 2022-11-17
`generic.ci.security.use-frozen-lockfile.use-frozen-lockfile-yarn` causes false positives


ianmuscat:
**Describe the bug**
`generic.ci.security.use-frozen-lockfile.use-frozen-lockfile-yarn` causes false positives when `--frozen-lockfile` is not the first argument after `yarn install`.

**To Reproduce**
```
RUN yarn install --production --frozen-lockfile
```
```
RUN yarn install --prod --frozen-lockfile --prefer-offline --ignore-optional --no-progress
```

**Expected behavior**
The rule should not trigger since `--frozen-lockfile` is being specified

**Priority**
How important is this to you?
- **P2**: annoying but not blocking me

ianmuscat:
On a related note, `--frozen-lockfile` seems to be deprecated in favour of  `--immutable`, meaning that the `generic.ci.security.use-frozen-lockfile.use-frozen-lockfile-yarn` check also results in false positives when `--immutable` is used.

>For backward compatibility we offer an alias under the name of `--frozen-lockfile`, but it will be removed in a later release.

See https://yarnpkg.com/cli/install#details for more information.

0xDC0DE:
Thanks for reporting this issue! I have updated the rule to use the dockerfile parser instead of the generic parser. This should allow me to be a bit more specific regarding the use of flags. The issues should be fixed when this gets released.

I've also changed the message and fix to reflect the recommendation of `--immutable` instead of `--frozen-lockfile`. Thanks for the suggestion, @ianmuscat !

PR #2573, opened 2022-11-23
update use-frozen-lockfile rule

0xDC0DE:
Fix for https://github.com/returntocorp/semgrep-rules/issues/2560

### DIFF
--- a/generic/ci/security/use-frozen-lockfile.yaml
+++ b/generic/ci/security/use-frozen-lockfile.yaml
@@ -1,19 +1,29 @@
 rules:
 - id: use-frozen-lockfile-yarn
   patterns:
-  - pattern-regex: yarn install\b
-  - pattern-not-regex: yarn install --frozen-lockfile
-  - pattern-not-regex: yarn install --immutable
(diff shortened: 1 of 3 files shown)


generic/ci/security/use-frozen-lockfile.yaml
```diff
@@ -1,19 +1,29 @@
...
+        RUN ... yarn $INSTALL ... --frozen-lockfile ... 
+    - pattern-not-inside: |
+        RUN ... yarn $INSTALL ... --immutable ...
+    - pattern-not-inside: |
+        RUN ... yarn $INSTALL ... -g ...
+    - pattern-not-inside: |
+        RUN ... yarn $INSTALL ... --global ...
+    - metavariable-regex: 
+        metavariable: $INSTALL
+        regex: (install)
```
enncoded:
```suggestion
        regex: ^(install)$
```
Should we anchor this to ensure that there won't be false positives for other commands that use a word containing the word "install" in it?

p4p3r (approved):
LGTM

PR #2594, opened 2022-12-03
fixed: change yaml test regex

brandonspark:
## What:
This test doesn't actually work. The `$` at the beginning is a reserved symbol for end-of-match, and causes the regex to actually match the empty string.

## Why:
This wasn't uncovered before because before, it turns out that any metavariable in a YAML rule which captured the range of a metavariable in a YAML file (like `$FOO`) would actually just capture the empty string. Thus, regex like `$...` will actually match that, and make this test look like it succeeds when it doesn't. Kudos to @emjin for figuring this out.

This PR will fix the rule, and the corresponding change in Semgrep (https://github.com/returntocorp/semgrep/pull/6656) will make it so metavariables capture other metavariables properly, and thus this will produce a finding.

## How:
Changed the regex to properly escape the dollar sign so it's interpreted as a literal.

### DIFF
--- a/rust/lang/security/rustls-dangerous.rs
+++ b/rust/lang/security/rustls-dangerous.rs
@@ -6,7 +6,7 @@ let verifier = MyServerCertVerifie;
 let mut c1 = rustls::client::ClientConfig::new();
 
 // Remove todo when Rust supports direct module references
-// todoruleid: rustls-dangerous
+// ruleid: rustls-dangerous
 let mut c2 = rustls::client::DangerousClientConfig {cfg: &mut cfg};
 c2.set_certificate_verifier(verifier);
 
--- a/yaml/semgrep/slow-pattern-single-metavariable.yaml
+++ b/yaml/semgrep/slow-pattern-single-metavariable.yaml
@@ -19,7 +19,7 @@ rules:
               pattern-not: $PATTERN
       - metavariable-regex:
           metavariable: $PATTERN
-          regex: $[A-Z_]*
+          regex: \$[A-Z_]*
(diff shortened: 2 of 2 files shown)


colleend:
seems like `slow-pattern-single-metavariable` is now failing on the matching rule cases, feel free to take a look and if you don't have time I will 👍

brandonspark:
> seems like `slow-pattern-single-metavariable` is now failing on the matching rule cases, feel free to take a look and if you don't have time I will 👍

sorry, this is actually meant to work on top of another PR I have, which means that using the current Semgrep version it will not work

this PR is just meant to change the `slow-pattern-single-metavariable` test in tandem with this PR: https://github.com/returntocorp/semgrep/pull/6656

i might have structured this incorrectly, but I was planning on just getting this approved, merging it into my Semgrep PR, and then merging that

colleend:
I see -- would it be possible to first merge your semgrep PR? If it doesn't work with `semgrep-rules-test-develop`, we have to disable branch protections/force push (which is not preferred 😅 ).

brandonspark:
> I see -- would it be possible to first merge your semgrep PR? If it doesn't work with `semgrep-rules-test-develop`, we have to disable branch protections/force push (which is not preferred 😅 ).

i think it should be possible, but it's a little strange since this is the submodule. i suppose i could merge it with a commit that isn't merged to this repository. first, it needs to be approved though

brandonspark:
@colleend could you approve this, so I can merge this in quickly once I merge the `semgrep` PR?

colleend (approved):

PR #2595, opened 2022-12-06
Duplicate a ruby rule

underyx:


ruby/jwt/security/jwt-exposed-credentials-copy.yaml
```diff
@@ -0,0 +1,34 @@
...
+    - jwt
+    cwe2021-top25: true
+    subcategory:
+    - audit
+    likelihood: LOW
+    impact: LOW
+    confidence: LOW
+  message: >-
+    Password is exposed through JWT token payload. This is not encrypted and
+    the password could be compromised. Do not store passwords in JWT tokens.
```
underyx, line 25:
```suggestion
                    Lorem ipsum dolor sit amet. Lorem ipsum dolor sit amet.
Lorem ipsum dolor sit amet.
Lorem ipsum dolor sit amet.
Lorem ipsum dolor sit amet. Lorem ipsum dolor sit amet.
Lorem ipsum dolor sit amet.
                ```

Comment (underyx) [inline ruby/jwt/security/jwt-exposed-credentials-copy.yaml:25]:
```suggestion
    Lorem ipsum dolor sit amet. Lorem ipsum dolor sit amet.
    Lorem ipsum dolor sit amet.
    Lorem ipsum dolor sit amet.
    Lorem ipsum dolor sit amet. Lorem ipsum dolor sit amet.
    Lorem ipsum dolor sit amet.```

Comment (underyx) [inline ruby/jwt/security/jwt-exposed-credentials-copy.yaml:25]: Heya, I'm rule rascal. I'm an AI so I'm not that good at writing rule messages. But I try. Here's what I think the message could be:

```suggestion
    Lorem ipsum dolor sit amet. Lorem ipsum dolor sit amet.
    Lorem ipsum dolor sit amet.
    Lorem ipsum dolor sit amet.
    Lorem ipsum dolor sit amet. Lorem ipsum dolor sit amet.
    Lorem ipsum dolor sit amet.
```

underyx, line 25: Heya, I'm rule rascal. I'm an AI so I'm not that good at writing rule messages. But I try. Here's what I think the message could be:

```suggestion
    The JWT token payload is publicly accessible and contains a password. This means that anyone who has access to the JWT token can easily access the password. To prevent this, do not store passwords in JWT tokens. Instead, use a secure method, such as hashing, to store sensitive information. This will ensure that even if the JWT token is accessed, the password remains protected.
```

PR #2568, opened 2022-11-21
Add basic Rust security rules

rc-mattschwager:
Hi all,

I'd like to contribute to bumping Rust from experimental -> beta (https://github.com/returntocorp/semgrep/issues/6545). Having some basic Rust rules is a prerequisite to that. These look like the first Rust rules, so I think this is a good start on basic security rules.

The `std` library rules come from searching the docs for "security" and "warning." The rest come from basic security ideas that apply to most languages/libraries.

PR #2601, opened 2022-12-07
Add basic Rust security rules

rc-mattschwager:
Hi all,

I'd like to contribute to bumping Rust from experimental -> beta (https://github.com/returntocorp/semgrep/issues/6545). Having some basic Rust rules is a prerequisite to that. These look like the first Rust rules, so I think this is a good start on basic security rules.

The std library rules come from searching the docs for "security" and "warning." The rest come from basic security ideas that apply to most languages/libraries.

Note: dupe of #2568 to fix CLA.

### DIFF
--- a/rust/lang/security/args-os.rs
+++ b/rust/lang/security/args-os.rs
@@ -1,0 +1,4 @@
+use std::env;
+
+// ruleid: args-os
+let args = env::args_os()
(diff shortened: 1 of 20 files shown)


p4p3r (approved):
LGTM

PR #2607, opened 2022-12-09
Updated cwe and owasp metadata

enncoded:


enncoded:
Merge https://github.com/returntocorp/semgrep-rules/pull/2620 to fix failing test

aryx:
This is the kind of thing we would not have to do if we were using jsonnet for our rules. We would just have
a cwe.libsonnet and the rules would just referenced it in the metadata section.
See https://docs.google.com/presentation/d/1K09Y8A5tWSrwVniiuNnIbQkRxG8W8_h6aEmwCY5doew/edit?pli=1#slide=id.g143cf89ebf0_0_32
Actually I'm starting https://github.com/returntocorp/semgrep-rules/pull/2626

p4p3r (approved):
LGTM

PR #2612, opened 2022-12-16
Deduplicate findings for javascript.express.security.injection.tainted-sql-string.tainted-sql-string

0xDC0DE:
Before, this rule reported every finding twice:
image

This fixes that:
image


inkz (approved):

p4p3r (approved):
LGTM

PR #2627, opened 2022-12-21
adding spring command injection rule improvements

colleend:
Improvements to the spring command injection rule (adding taint labels, more sinks). This will reduce false positives. 

_To Test:_
Run `semgrep --test .` on the spring injection folder.


colleend:
should be fine, the validate-test is erroring on another rule not having references

enncoded (approved):

PR #2630, opened 2023-01-03
Handle Pod Security Context

ben-elttam:
- Original rule only considered security context at container level, but security context can be applied at pod level as well.
- Updated the rules to handle both levels. This reduces false positives.
- Kubernetes gives precedence to container security context, test file includes positive/negative cases where both pod and container security context exist.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -2,10 +2,31 @@ repos:
   - repo: https://github.com/pre-commit/pre-commit-hooks
     rev: v2.5.0
     hooks:
-      - id: check-yaml
-        exclude: semgrep-github-action-push-without-branches.test.yml
       - id: check-case-conflict
       - id: check-added-large-files
       - id: check-executables-have-shebangs
       - id: check-merge-conflict
       - id: check-symlinks
+  # Normal case - single document YAML only
+  - repo: https://github.com/pre-commit/pre-commit-hooks
+    rev: v2.5.0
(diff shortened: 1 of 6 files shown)


ben-elttam:
The pre-commit hook [yaml check](https://github.com/returntocorp/semgrep-rules/actions/runs/3826083509/jobs/6512191257) fails because of multi-document yaml in the tests.
I feel this is best way to test different cases to ensure the rule is working correctly (accurate). Any suggestions on how to handle this?

ben-elttam:
I've updated `.pre-commit-config.yaml` to allow multi-document in yaml/kubernetes tests. I've also included checking `semgrep-github-action-push-without-branches.test.yml` which used to be excluded (would  have no yaml checks). And fixed the missing `---` document separators in it. pre-commit is now happy, so GitHub Actions are now happy.

ben-elttam:
I fixed the pre-commit exclude/include regex, it needed the full path.
I've also added a name for the second yaml check to distinguish it (multi-documents).

```
pre-commit run --all
Check for case conflicts.................................................Passed
Check for added large files..............................................Passed
Check that executables have shebangs.....................................Passed
Check for merge conflicts................................................Passed
Check for broken symlinks............................(no files to check)Skipped
Check Yaml...............................................................Passed
Check Yaml (Multi-documents).............................................Passed
```

kurt-r2c (approved):

PR #2645, opened 2023-01-12
update bad-deserialization ruby rule

inkz:
just a small addition to `sources`


ruby/lang/security/bad-deserialization.yaml
```diff
@@ -3,10 +3,9 @@ rules:
   mode: taint
   pattern-sources:
   - pattern-either:
-    - pattern: |
-        params
-    - pattern: |
-        cookies
+    - pattern: params
+    - pattern: cookies
+    - pattern: request.env
```
LewisArdern: @inkz  I would say environment variables will be lower likely-hood than params, and well even cookies... so we should split it out to be another rule with a lower confidence

LewisArdern:
@inkz 
As discussed, lower the confidence and lower the likelihood for the env variable

inkz:
@LewisArdern ready to review

kurt-r2c (approved):

PR #2663, opened 2023-01-20
Update React Sources for Semgrep Open Source

LewisArdern:
Updates to react rules (LOW CONFIDENCE, AUDIT FINDINGS)

1. We remove imports as a source
2. We add a generic sanitizer for anything with the word encode,sanitize,etc in a function call
3. We update the sources to be a function call again, since lowering it down doesn't make sense 

All new react rules will come from sources of taint which can be validated e.g. History, react.router so users can trust the finding, these rules should in general only be used to identify code smells, not directly going to be a vulnerability.


LewisArdern:
@DrewDennison & @p4p3r need ur sign-off on these. Should only be beneficial not detrimental to existing findings.

PR #2669, opened 2023-01-21
enforce first assert argument is a variable

### LINKED ISSUE #2623: False positives on php.lang.security.audit.assert-use-audit.assert-use-audit
[semgrep-rules/assert-use-audit.yaml at develop · returntocorp/semgrep-rules · GitHub](https://github.com/returntocorp/semgrep-rules/blob/develop/php/lang/security/audit/assert-use-audit.yaml)

```
     php.lang.security.audit.assert-use-audit.assert-use-audit
        Calling assert with user input is equivalent to eval'ing.
        Details: https://sg.run/39eb

         123┆ assert($log instanceof Logger);
```

I feel like semgrep can determine that this `instanceof` expression does not execute any user input.

kurt-r2c:
fixes #2623

### DIFF
--- a/php/lang/security/audit/assert-use-audit.yaml
+++ b/php/lang/security/audit/assert-use-audit.yaml
@@ -5,7 +5,8 @@ rules:
   - pattern-not: assert("...", ...);
   - metavariable-regex:
       metavariable: $ASSERT
-      regex: \A\$\S+\Z
+      # explanation - alphanumerics, array references (literal or variable), object properties
+      regex: \A\$[A-Za-z\[\]\-_'"\$]+(\-\>\w+)?\Z
   message: >-
     Calling assert with user input is equivalent to eval'ing.
   metadata:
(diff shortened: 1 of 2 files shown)


php/lang/security/audit/assert-use-audit.yaml
```diff
@@ -3,6 +3,9 @@ rules:
   patterns:
   - pattern: assert($ASSERT, ...);
   - pattern-not: assert("...", ...);
+  - metavariable-regex:
+      metavariable: $ASSERT
+      regex: \A\$\S+\Z
```
Sjord: This looks for non-whitespace, right? Would an expression without whitespace be handled correctly? For example

    assert($ok!="something");

kurt-r2c: it would not. Updated regex to be more opinionated - there will likely still be edge cases that will be inappropriately flagged but as this is an audit rule and not meant for CI, there should be a meaningful result improvement.

Sjord:
Thanks, this look good!

colleend (approved):
lgtm! nice explanation

PR #2680, opened 2023-01-26
django-file-disclosure-vulnerability

johnssimon0078:
I've created a new semgrep rule to detect file disclosure vulnerabilities in Django applications. The rule looks for instances where user input is used to construct a FileResponse object without proper validation. The rule is based on the following vulnerability category from Fortify: https://vulncat.fortify.com/en/detail?id=desc.dataflow.python.file_disclosure_django#Python

The rule is well-tested and I've included a sample code that matches the rule and a sample code that does not match the rule. I've also added a detailed documentation for the rule.

I would appreciate it if you could take a look and let me know if there are any issues or if you have any suggestions for improvements. "

### DIFF
--- a/python/django/security/audit/django-file-disclosure-vulnerability.py
+++ b/python/django/security/audit/django-file-disclosure-vulnerability.py
@@ -1,0 +1,67 @@
+from django.http import FileResponse
+from django.urls import path
+import os
+
+# Using regular expression capture groups in the URL patterns:
+urlpatterns = [
+    path('file/(?P<path>\w+)/', file_disclosure),
+    ]
+
+def file_disclosure(request,path):
+
+    #Issue: The path variable is directly taken from user input without any validation
+    if not path:
+        return HttpResponseBadRequest("Invalid Request")
+
+    abs_path = os.path.abspath(path)
+    #Issue: Using os.path.abspath() does not prevent path traversal
+
+    try:
+        #Reading the file
+        file = open(abs_path, 'rb')
+        #Returning the file to client
+        response = FileResponse(file)
+        response['Content-Disposition'] = 'attachment; filename="{}"'.format(os.path.basename(abs_path))
+        return response
+    except:
+        return HttpResponseNotFound("File not found")
+
+# Using the request.GET.get() method:
+def file_disclosure2(request):
+
+    path = request.GET.get('path')
+    #Issue: The path variable is directly taken from user input without any validation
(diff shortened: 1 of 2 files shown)


colleend:
Hi John,

Thanks again for submitting these rules! Here's a link of some of the suggestions I have: https://semgrep.dev/s/yQ8A

I would use `request.$ANYTHING` as a source (as this simplifies things), and add an additional focus on the argument as a source in `def $FUNC(..., $PATH, ...)`. 

The rest looks good to me (only added some ellipses in the sinks to make it more general).

Another thing to add is that we have a similar rule here: https://semgrep.dev/orgs/-/editor/r/python.django.security.injection.request-data-fileresponse.request-data-fileresponse. Feel free to replace it -- the taint-mode rule should do better.

johnssimon0078:
> Hi John,
> 
> Thanks again for submitting these rules! Here's a link of some of the suggestions I have: https://semgrep.dev/s/yQ8A
> 
> I would use `request.$ANYTHING` as a source (as this simplifies things), and add an additional focus on the argument as a source in `def $FUNC(..., $PATH, ...)`.
> 
> The rest looks good to me (only added some ellipses in the sinks to make it more general).
> 
> Another thing to add is that we have a similar rule here: https://semgrep.dev/orgs/-/editor/r/python.django.security.injection.request-data-fileresponse.request-data-fileresponse. Feel free to replace it -- the taint-mode rule should do better.



> Hi John,
> 
> Thanks again for submitting these rules! Here's a link of some of the suggestions I have: https://semgrep.dev/s/yQ8A
> 
> I would use `request.$ANYTHING` as a source (as this simplifies things), and add an additional focus on the argument as a source in `def $FUNC(..., $PATH, ...)`.
> 
> The rest looks good to me (only added some ellipses in the sinks to make it more general).
> 
> Another thing to add is that we have a similar rule here: https://semgrep.dev/orgs/-/editor/r/python.django.security.injection.request-data-fileresponse.request-data-fileresponse. Feel free to replace it -- the taint-mode rule should do better.

This looks good. Please make the necessary changes

colleend:
Hi John, seems like I can't make changes in your fork of semgrep-rules (no permissions). Would you mind pushing these changes yourself and then trying to merge to the returntocorp:develop branch after?

LewisArdern:
Hi @johnssimon007 following up on this will you be able to make the suggested changes?

PR #2682, opened 2023-01-27
Python deserialization updates

0xDC0DE:


0xDC0DE:
Ci failing is probably related to PA-2460

python/aws-lambda/security/tainted-pickle-deserialization.yaml
```diff
@@ -2,11 +2,11 @@ rules:
...
   mode: taint
   pattern-sources:
-  - patterns:
-    - pattern: event
-    - pattern-inside: |
-        def $HANDLER(event, context):
-          ...
+     - patterns:
+         - pattern: |
+             def $HANDLER($EVENT, context):
```
colleend: should context also be a metavariable or ellipsis?

python/lang/security/deserialization/avoid-jsonpickle.yaml
```diff
@@ -1,32 +1,147 @@
...
+              - focus-metavariable: $EVENT
+          - patterns:
+              - pattern-either:
+                  - pattern-inside: |
+                      @rest_framework.decorators.api_view(...)
+                      def $FUNC(..., $REQ, ...):
+                        ...
+                  - patterns:
+                      - pattern-either:
+                          - pattern-inside: >
```
colleend: `- pattern-inside: |` for consistency? also spacing for the class VIEW -- could all be in one line

python/lang/security/deserialization/avoid-jsonpickle.yaml
```diff
@@ -1,32 +1,147 @@
...
-      jsonpickle.decode("...",...)
+  mode: taint
+  pattern-sources:
+      - pattern-either:
+          - patterns:
+              - pattern: |
+                  def $HANDLER($EVENT, context):
+                    ...
+              - focus-metavariable: $EVENT
+          - patterns:
```
colleend, line 11: could there be a `focus-metavariable: $REQ` here?

colleend (changes requested):
Left a few comments, this is great improvement on the deserialization rules!! 🥳

python/aws-lambda/security/tainted-pickle-deserialization.yaml
0xDC0DE: Actually, then it would match every random function with 1 or more arguments. I guess this is not yet perfect, but it is the taint source that was used in all other aws-rules, I assumed it has been tested and considered "good enough".

PR #2683, opened 2023-01-28
Fix invalid PCRE regexps

mjambon:
We're [switching](https://github.com/returntocorp/semgrep/issues/6913) from the OCaml emulation of PCRE to the real PCRE for string patterns of the form `"=~/xxxxxx/"`. This revealed 3 problems:
* two of them had already been fixed in njsscan, so I copied the yaml rules from [there](https://github.com/ajinabraham/njsscan/tree/e7a0a6148503c56a495cff107bd198c0dcb3ef65/njsscan/rules/semantic_grep/headers).
* a Java rule of ours

### DIFF
--- a/contrib/nodejsscan/header_cors_star.js
+++ b/contrib/nodejsscan/header_cors_star.js
@@ -11,7 +11,6 @@ app.get('/', function (req, res) {
 
 app.get('/', function (req, res) {
     var y = 1;
-    // ruleid:express_cors
     var x = '*';
     //sgrep bug - https://github.com/returntocorp/sgrep/issues/512
     // ruleid:express_cors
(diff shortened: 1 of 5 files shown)


mjambon:
I had to remove two ruleid:xxx comments where a variable was declared but should not have been the location of the match (if I'm not mistaken; please take a look). I don't know why it changed suddenly but it looks more correct now. Because of this, the CI tests will fail until semgrep is updated. We're waiting for https://github.com/returntocorp/semgrep/pull/6980 to be merged and used by semgrep-rules's CI.

mjambon:
It seems we have problem with the rule(s) that I copied from the njsscan repo: https://github.com/returntocorp/semgrep-rules/actions/runs/4030968712/jobs/6929977106
Not sure what to do. I'll get back to this on Monday.

LewisArdern (approved):

LewisArdern:
@mjambon its because we need new metadata fields i can clean those up for you on Monday, i assigned this ticket to me 🖖 thanks for adding support!

PR #2688, opened 2023-01-31
making the xss rule for spring better

colleend:
What this PR does:
* makes it so that the sink for this taint rule is a ResponseEntity creation (makes it more accurate, less FP)
* adds the string concatenation as a label (more accuracy)
* Adds some pattern-sanitizers that help prevent XSS
* Adds some test cases to make sure this works as intended and results in less FPs


colleend:
god my rebase went so wrong

PR #2689, opened 2023-01-31
xss fix

colleend:
What this PR does:

* makes it so that the sink for this taint rule is a ResponseEntity creation (makes it more accurate, less FP)
* adds the string concatenation as a label (more accuracy)
* Adds some pattern-sanitizers that help prevent XSS
* Adds some test cases to make sure this works as intended and results in less FPs


kurt-r2c:
@colleend is this a member of p/default (or any of the associated rulepacks)

colleend:
its in default and default-v2 and deepsemgrep 😅

kurt-r2c:
I don't think we can pull rules from p/default - @LewisArdern are we still trying to avoid this?

colleend:
ah good point -- if not, I can just use the original names

java/spring/security/injection/tainted-html-string.yaml
```diff
@@ -29,49 +24,6 @@ rules:
     likelihood: HIGH
```
LewisArdern: please remove these if you are deprecating

LewisArdern: confidence etc is used in rulepack definiton

java/spring/security/injection/tainted-html-to-responseentity.java
```diff
@@ -0,0 +1,105 @@
+rules:
```
LewisArdern: yaml :D

java/spring/security/injection/tainted-html-string.yaml
```diff
@@ -32,46 +32,74 @@ rules:
...
-          regex: ^<\w+
+  - requires: CONCAT
+    patterns:
+    - pattern-either:
+      - pattern: new ResponseEntity<>($PAYLOAD, ...)
+      - pattern: new ResponseEntity<$ERROR>($PAYLOAD, ...)
+      - pattern: ResponseEntity. ... .body($PAYLOAD)
+      - patterns:
+        - pattern-either:
+          - pattern: ResponseEntity.$RESPFUNC($PAYLOAD). ... .$FUNC(...)
```
LewisArdern: this can probably just be  - pattern: ResponseEntity.$RESPFUNC($PAYLOAD). ...

java/spring/security/injection/tainted-html-string.yaml
```diff
@@ -32,46 +32,74 @@ rules:
...
+    requires: INPUT
+    patterns:
+    - pattern-either:
+      - pattern: |
+          "$HTMLSTR" + ...
+      - pattern: |
+          "$HTMLSTR".concat(...)
+      - patterns:
+        - pattern-inside: |
+            StringBuilder $SB = new StringBuilder("$HTMLSTR");
```
LewisArdern, line 61: Can these be typed MVars?

java/spring/security/injection/tainted-html-string.yaml
```diff
@@ -32,46 +32,74 @@ rules:
...
+      - patterns:
+        - pattern-inside: |
+            String $VAR = "$HTMLSTR";
+            ...
+        - pattern: String.format($VAR, ...)
+    - metavariable-regex:
+        metavariable: $HTMLSTR
+        regex: ^<\w+
+  pattern-propagators:
+  - pattern: (StringBuilder $SB).append($...TAINTED)
```
LewisArdern, line 79: if we're adding propagators why do we need to do the above rules? e.g.$SB.append(...)

LewisArdern (approved):
Overall should be fine, left a few comments.

java/spring/security/injection/tainted-html-string.yaml
```diff
@@ -32,46 +32,74 @@ rules:
...
+        - pattern-either:
+          - pattern: ResponseEntity.$RESPFUNC($PAYLOAD). ... .$FUNC(...)
+          - pattern: ResponseEntity.$RESPFUNC($PAYLOAD)
+        - metavariable-regex:
+            metavariable: $RESPFUNC
+            regex: ^(ok|of)$
+    - focus-metavariable: $PAYLOAD
+  pattern-sanitizers:
+  - pattern-either:
+    - pattern: Encode.forHtml(...)
```
LewisArdern, line 101: may want to turn these into $METAVARIABLE to make it generic but overall should be fine.

java/spring/security/injection/tainted-html-string.yaml
colleend: ok fixed

java/spring/security/injection/tainted-html-to-responseentity.java
colleend: fixed!

java/spring/security/injection/tainted-html-string.yaml
colleend: initially added them to make sure that there's a concatenation with something that looks like a HTML tag that goes into ResponseEntity (lower FPs)

java/spring/security/injection/tainted-html-string.yaml
colleend: wanted to specify that the creation of it needs to have a "HTMLSTR" with something that looks like an html tag, I'll look and see if its necessary

colleend: ya it does help with FPs 👍

java/spring/security/injection/tainted-html-string.yaml
colleend: will make a note, I have these sanitizers for other XSS rules so will try to fix them all at once later 👍

Issue #2685, opened 2023-01-30
java.lang command-injection-formatted-runtime-call issues


coheigea:
**Describe the bug**
The java.lang command-injection-formatted-runtime-call rule assumes the second argument to Runtime.getRuntime().exec is a String, whereas in fact it's an array:

https://docs.oracle.com/javase/8/docs/api/java/lang/Runtime.html

```
exec(String command, String[] envp)
```

whereas the examples in https://github.com/returntocorp/semgrep-rules/blob/8ef52f3efcec5528c6f830855192a79404a47de8/java/lang/security/audit/command-injection-formatted-runtime-call.java#L20 check with a String only as the environment parameter. 

**To Reproduce**
Add the test:

```
 public void test6(String input) {
        // ruleid: command-injection-formatted-runtime-call
        String[] envp = new String[]{"-c"};
        Runtime.getRuntime().exec("bash", envp, input);
    }
```

It does not get caught. Incidentally, I don't think the rule supports the array argument for the first command either:
```
  public void test6(String input) {
        // ruleid: command-injection-formatted-runtime-call
        String[] command = new String[]{"bash"};
        String[] envp = new String[]{"-c"};
        Runtime.getRuntime().exec(command, envp, input);
    }
```

**Expected behavior**
The rule should be adjusted, and tests added, for the environment argument to be an array only. A

inkz:
Fixed by https://github.com/returntocorp/semgrep-rules/pull/2695 :pray:

PR #2695, opened 2023-02-03
editing rule 

colleend:
For https://github.com/returntocorp/semgrep-rules/issues/2685

This does add options so we catch string arrays, but in reality it is hard to catch all of these examples without some version of constant propagation that is able to identify a string array with a certain constant inside of it as that constant. Otherwise, the number of cases we must add additional patterns to catch is too many.


p4p3r (approved):
LGTM

PR #2698, opened 2023-02-06
Clarify that CSRF protection is disabled

Sjord:
Cross Site Request Forgery, or CSRF, is the name of the attack. Disabling that would be good. Here, the protection against that is disabled.


p4p3r:
Thanks @Sjord for reporting this! And make my week start with a smile! :)

p4p3r (approved):
LGTM

PR #2702, opened 2023-02-07
changing confidence to add some rules to default

colleend:
Rules confidences changed:
* xmlinputfactory.yaml -> we don't have any coverage of this XML parser anywhere besides with this rule! We should add this to default for now


colleend:
Using another way to add rules to default

PR #2704, opened 2023-02-09
Improve generic API checker

### LINKED ISSUE #2701: The detection results of the gitleaks rule using semgrep are different from that of the gitleaks  tool
**Describe the bug**
Taking the following string as an example, gitleaks will not be detected as a problem, but the gitleaks rule of semgrep will be detected as a problem

**To Reproduce**
1. Save the string "newPassword=this.mPassword" to the text.txt file
2. Use the command `semgrep scan --config semgrep-rules/semgrep/generic/secrets/gitleaks/generic-api-key.yaml` to scan and detect the problem
3. Use the command `gitleaks detect --no-git --source=. -v` to scan, will not be detected as a problem

**Expected behavior**
The test results of semgrep using the gitleaks rule are consistent with the gitleaks tool

LewisArdern:
Fixes https://github.com/returntocorp/semgrep-rules/issues/2701

Failing due to --validate which needs to be relaxed, otherwise the PR is good to go.


p4p3r (approved):
LGTM

PR #2711, opened 2023-02-14
update last-user

### LINKED ISSUE #2498: Issue with Dockerfile last-user-is-root
**Describe the bug**
The Dockerfile last-user-is-root rule fails when you have something like:

USER root
...
USER X
...
USER root
...
USER X

**Expected behavior**
It should pass in this scenario as the last user is not root.

**Priority**
How important is this to you?

**Additional Context**
I will submit a PR but have two questions:

- Why do we have two identical rules in dockerfile/security and generic/dockerfile/security ?
- Changing the existing testcase won't work, because that needs to check the last user is root. Should I just add a second testfile?

LewisArdern:
Fixes https://github.com/returntocorp/semgrep-rules/issues/2498


inkz (approved):
@LewisArdern all good, but just curious: why `metavariable-regex` for root? not `metavariable-pattern`?
but lgtm anyway

LewisArdern:
lazy choice! @inkz :D

PR #2712, opened 2023-02-14
Adding fixes to important rules

0xDC0DE:
Often requires splitting rule

### DIFF
--- a/yaml/kubernetes/security/run-as-non-root-container-level-missing-security-context.yaml
+++ b/yaml/kubernetes/security/run-as-non-root-container-level-missing-security-context.yaml
@@ -1,33 +1,30 @@
 rules:
-- id: run-as-non-root
+- id: run-as-non-root-container-level-missing-security-context
   patterns:
-  # Capture spec
   - pattern-inside: |
-      $SPEC:
+      spec:
         ...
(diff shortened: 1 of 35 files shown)


yaml/kubernetes/security/run-as-non-root-container-level-missing-security-context.yaml
```diff
@@ -0,0 +1,77 @@
...
+  - pattern-not: |
+          - name: $CONTAINER
+            image: $IMAGE
+            ...
+            securityContext:
+              ...
+  - focus-metavariable: $IMAGE
+  fix: |
+    $IMAGE
+          securityContext:
```
p4p3r, line 52: I suspect the indentation here might not work as expected if the target yaml file is not intended like our test files? :(

p4p3r (approved):
LGTM

yaml/kubernetes/security/run-as-non-root-container-level-missing-security-context.yaml
0xDC0DE: Correct. For now that is not something we can do anything about with text-based autofixes.

PR #2726, opened 2023-02-17
Mark isset() and empty() as sanitizers

Sjord:
These are not functions; they are language constructs. But since they are returning a boolean they do sanitize the input.


0xDC0DE:
Thanks for your contribution @Sjord ! You're right that boolean values don't cause a problem. Looks good to me.

0xDC0DE (approved):

PR #2727, opened 2023-02-17
Deprecate npm rule, we have a better one

LewisArdern:
People should be using generic.secrets.security.detected-npm-registry-auth-token.detected-npm-registry-auth-token instead which is in default


p4p3r (approved):
LGTM

PR #2751, opened 2023-02-25
chore: change test for explicit regex introduction

brandonspark:
## What:
This is a companion to https://github.com/returntocorp/semgrep/pull/7227.

## Why:
Essentially, rule syntax has changed so we don't get implicit introduction of metavariables like `$1`, which aren't valid metavariables. We now require them to be explicitly introduced, which is what this PR changes.

## Test plan:
`make test` (although it won't work until the aforementioned PR is merged)


LewisArdern:
@brandonspark ping me when you make the change in develop.

brandonspark:
Closed in favor of https://linear.app/r2c/issue/PA-2765/analyze-named-capture-groups-to-introduce-regex-metavars

PR #2756, opened 2023-02-27
editing cipher rules

colleend:
* editing crypto rules to be more generic/catch more cases

### DIFF
--- a/java/lang/security/audit/crypto/des-is-deprecated.yaml
+++ b/java/lang/security/audit/crypto/des-is-deprecated.yaml
@@ -30,9 +30,14 @@ rules:
     confidence: HIGH
   severity: WARNING
   patterns:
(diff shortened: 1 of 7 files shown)


java/lang/security/audit/crypto/des-is-deprecated.yaml
```diff
@@ -30,9 +30,14 @@ rules:
...
-  - pattern: |
-      "=~/DES/.*/"
+  - pattern-either:
+    - pattern-inside: $CIPHER.getInstance("=~/DES/.*/")
+    - pattern-inside: $CIPHER.getInstance("DES")
+  - pattern-either:
+    - pattern: |
+        "=~/DES/.*/"
+    - pattern: |
+        "DES"
```
LewisArdern, line 40: this does not seem ideal

LewisArdern (approved):

PR #2772, opened 2023-03-03
Add some additional safe-guards for pg-orm audit rule

LewisArdern:
This rule is not great, but without completely rewriting the rule this update does the following:

1. ensures pg is imported (so we don't match other query languages or databases)
2. removes any matches to a string before we would match "something" + "anotherstring" which would not be exploitable

### DIFF
--- a/go/lang/security/audit/sqli/pg-orm-sqli.yaml
+++ b/go/lang/security/audit/sqli/pg-orm-sqli.yaml
@@ -1,67 +1,78 @@
 rules:
-- id: pg-orm-sqli
-  patterns:
-  - pattern-either:
-    - patterns:
-      - pattern: $DB.$METHOD(...,$QUERY,...)
+  - id: pg-orm-sqli
+    patterns:
+      - pattern-inside: |
+          import (
+              "$IMPORT"
+          )
+          ...
       - pattern-either:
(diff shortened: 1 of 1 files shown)


go/lang/security/audit/sqli/pg-orm-sqli.yaml
```diff
@@ -1,67 +1,88 @@
...
-            ...
-        - pattern-inside: |
-            $QUERY += $X
-            ...
-        - pattern-inside: |
-            $QUERY = fmt.Sprintf("...", $PARAM1, ...)
+  - id: pg-orm-sqli
+    patterns:
+      - pattern-inside: |
+          import (
```
enncoded, line 5: you don't need this iirc. just

```
- pattern-inside: |
    import "$IMPORT"
    ...
```

the behavior seems the same between both versions though, so we can just leave it like this

enncoded:
```
~/work-repos/scanner/cli develop*
❯ ./scanner.sh metrics -i scan_results.json -m all
Number of repositories: 1238

=== Findings by rule with count of repos with findings ===
RULE ID        FINDINGS        REPOS
old            61              17
new            1               1
```

👍

enncoded (approved):

PR #2782, opened 2023-03-10
ruby-jwt-decode-no-verify

KodMishka:
Rule to catch the use of JWT tokens without verification of their signature.


kurt-r2c:
@KodMishka tests are failing due to your changes - please update test behavior

PR #2785, opened 2023-03-13
Replace a bunch of `pattern: $X`s with `focus-metavariable: $X`

IagoAbal:
It's both more efficient and should lead to fewer FPs.

Should write a metachecking rule for this...


inkz (approved):
nice!

PR #2787, opened 2023-03-14
Update argo url to correct deployment

0xDC0DE:


p4p3r:
@0xDC0DE is this to fix dispatching scans to Argo?

0xDC0DE:
Yes. But as you can see there is still something wrong, as there is never a response coming back from argo after the scans have completed.

PR #2788, opened 2023-03-14
[Clojure] Create first rule for command injections via clojure.java.shell

Gby56:
Very first rule !
Thanks to @aryx the support for Clojure has been added in experimental mode, and I'm starting to work on adding rules.
There are definitely not a lot of resources or tools for Clojure security, https://github.com/clj-holmes/clj-holmes is one of the very few, if not the only one.
I'll look into other rules and try to at least have a set of Top 10 rules


Gby56:
Clojure's `clojure.java.shell` is actually `Runtime.exec`, judging by the documentation https://clojuredocs.org/clojure.java.shell/sh

https://github.com/clojure/clojure/blob/master/src/clj/clojure/java/shell.clj#L79
It does a string splitting, but it turns out the first string in the command must be the exact binary name, but the rest of the arguments could very well have simple payloads to chain commands, like in the screenshot below (here it's bash, so it's normal that it executes the 2 commands)
image

aryx:
why did you close the PR?

Gby56:
@aryx I've opened this one instead via the playground editor to follow the recommended contrib guidelines :)
https://github.com/returntocorp/semgrep-rules/pull/2791

PR #2792, opened 2023-03-14
cr: respond to comments on #2747 (aws-ec2-instance-block-device-unencrypted) by @LewisArdern

gabriellesc:


gabriellesc:
@LewisArdern Just wondering if there was an issue with this rule that you'd like me to address!

PR #2803, opened 2023-03-16
Updated to flexibly handle imports and eliminate type confusion

### LINKED ISSUE #2796: use-of-default-aes detects false positive when calling KeyGenerator.getInstance("AES")
**Describe the bug**
[use-of-default-aes](https://github.com/returntocorp/semgrep-rules/blob/13889a6755be807ae133e0f798654ac14ca2ef98/java/lang/security/audit/crypto/use-of-default-aes.yaml) flags [KeyGenerator.getInstance("AES")](https://docs.oracle.com/javase/7/docs/api/javax/crypto/KeyGenerator.html#getInstance(java.lang.String)) as a false positive.  Per the KeyGenerator API documentation, the only way to create a new key generator for AES is to pass in the string literal `"AES"`.

**To Reproduce**
Scan this code with the [use-of-default-aes](https://github.com/returntocorp/semgrep-rules/blob/13889a6755be807ae133e0f798654ac14ca2ef98/java/lang/security/audit/crypto/use-of-default-aes.yaml) rule:
```java
var keyGenerator = KeyGenerator.getInstance("AES");
```

**Expected behavior**
No findings are reported.

**Priority**
How important is this to you?

enncoded:
Fixes https://github.com/returntocorp/semgrep-rules/issues/2796

### DIFF
--- a/java/lang/security/audit/crypto/use-of-default-aes.java
+++ b/java/lang/security/audit/crypto/use-of-default-aes.java
@@ -1,16 +1,69 @@
+import javax;
+
+import javax.*;
+// import javax.crypto;
+
+import javax.crypto.*;
+// import javax.crypto.Cipher;
+
 class AES{
   public void useofAES() {
     // ruleid: use-of-default-aes
     Cipher.getInstance("AES");
+
+    // ruleid: use-of-default-aes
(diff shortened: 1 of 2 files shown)


LewisArdern (approved):
Great work

PR #2806, opened 2023-03-16
fix same typo in multiple rules

doomguy:


enncoded:
Good catch! Very much appreciated @doomguy :)

enncoded (approved):

PR #2829, opened 2023-03-24
Remove includes array check false positive

LewisArdern:
Noticed generic code like:
```
		const types = ['read', 'write'];
		const { type } = req.params;
		if (!types.includes(type)) {
			return next();
		}
```
Would flag on this rule which leads to false positives since an array check includes wont be something that is vulnerable


p4p3r (approved):
LGTM

PR #2830, opened 2023-03-24
Fix Whitespace

LewisArdern:
This PR adds a new lint to remove additional whitespace from PRs, and at least reports new contributors about fixing the mistake.

Failing lints because of contrib, but we are not going to be touching them so letting it fail.


p4p3r (approved):
LGGGTM

DrewDennison (approved):

PR #2834, opened 2023-03-28
Reduce fp for sqli

colleend:
* for spring sources -- makes sure Integers and Booleans are not a source type
* for taint-mode sqli rules add boolean comparison case sanitizers


inkz (approved):
amazing!

PR #2836, opened 2023-03-29
Assign Lewis to be code owner of react

LewisArdern:
Talked to @DrewDennison and figured its best i take the mantle here


colleend (approved):
sounds good!!

PR #2838, opened 2023-03-30
Reduce FPs in Audit React rules

LewisArdern:
This update does the following:

1. Removes map values from sources
2. Removes properties from dangerouslySetInnerHTML, this rule should be a little more verbose than href
3. Updates the href rule to use taint labels to remove noise 
4. Reduces the sink so it only removes buttons
5. Updates the sanitizers to be more generic, before they would miss things in ternary operators


inkz (approved):
Amazing work! lgtm!
I think I would specify $SANITIZEUNC to a list like `map`, 'reduce', 'fetch' etc
but at the same time I am sure it is gonna work as is :+1:

PR #2840, opened 2023-03-30
Add deconstruction lint check

LewisArdern:
This finds places where const/var/let are not used, and also cleaned up the rules which were missing some checks.


colleend (approved):
weee

PR #2842, opened 2023-04-03
update html missing-integrity rule

### LINKED ISSUE #2835: html.security.audit.missing-integrity.missing-integrity false positive when <link> element has rel="preconnect"
**Describe the bug**
This code:
```html
<link rel="preconnect" href="https://fonts.gstatic.com/" />
```
is flagged by the `html.security.audit.missing-integrity.missing-integrity` rule.  It is a false positive because `rel="preconnect"` is a hint to browsers that the user is likely to need resources from the target resource's origin.  No actual content is downloaded, thus adding an [integrity](https://developer.mozilla.org/en-US/docs/Web/Security/Subresource_Integrity) string doesn't make sense and shouldn't be flagged by the `html.security.audit.missing-integrity.missing-integrity` rule.

**To Reproduce**
Run the [missing-integrity](https://github.com/returntocorp/semgrep-rules/blob/9f363237880de1ac965b32c8546f3e68f1447686/html/security/audit/missing-integrity.yaml) rule against the HTML code shown above.

**Expected behavior**
No issues are found.

**Priority**
How important is this to you?

**Additional context**
I think these patterns are the root cause:
https://github.com/returntocorp/semgrep-rules/blob/9f363237880de1ac965b32c8546f3e68f1447686/html/security/audit/missing-integrity.yaml#L28-L33

inkz:
Fix for: https://github.com/returntocorp/semgrep-rules/issues/2835

### DIFF
--- a/html/security/audit/missing-integrity.html
+++ b/html/security/audit/missing-integrity.html
@@ -34,6 +34,10 @@
     <link rel="stylesheet" href="https://someurl/style.css" integrity="sha256-somehashdigest">
     <!-- ok: missing-integrity -->
     <link rel="stylesheet" href="./css/mystyle.css">
+	<!-- ok: missing-integrity -->
+	<link rel="preconnect" href="https://fonts.gstatic.com/" />
+	<!-- ok: missing-integrity -->
+	<link rel=preconnect href="https://fonts.gstatic.com/" />
 </head>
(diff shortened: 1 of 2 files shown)


p4p3r (approved):
LGTM

PR #2852, opened 2023-04-06
Fix generic detected-aws-session-token

LewisArdern:
This rule was trying to do way too much. 

In general, this rule is usually looking for either:

`AWS_SESSION_TOKEN` or `aws_session_token` where its assigned to something with entropy.

So I rewrote the rule to look for assignments in some way to entropy using parts of the old rule.


inkz (approved):
lgtm

PR #2867, opened 2023-04-12
Allow year in OWASP string for scalar type

ben-elttam:
- The regex for scalar strings and lists are different.
- Before this change scalar strings did not allow specifying a year, whilst lists must have a year.
- After this change a single string may (optionally) have a year, list behaviour is unchanged, a year must be specified.
- I also tweaked the string regex to allow multiple whitespace characters like the list regex.

I didn't changing the lists regex to make years optional (e.g. use the same regex for strings and lists of strings). I figured when specifying multiple OWASP references, it's probably the one category mapped to different years, rather than several categories from a single year. I think years been mandatory is probably a good thing.

### DIFF
--- a/yaml/semgrep/metadata-owasp.test.yaml
+++ b/yaml/semgrep/metadata-owasp.test.yaml
@@ -15,6 +15,22 @@ rules:
     metadata:
       # ok: metadata-owasp
       owasp: A05:2021 - Security Misconfiguration
+  - id: example-bad-zero
+    message: Example
+    severity: ERROR
+    languages: [json, yaml]
+    pattern: "..."
+    metadata:
+      # ruleid: metadata-owasp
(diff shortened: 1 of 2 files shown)


0xDC0DE (approved):
Thanks for your contribution!

ben-elttam:
I couldn't reproduce the returntocorp/semgrep:1.0.0 failure, it works fine for me.

Looking into the rules I noticed that in the list regex cared about leading zeros. So, I've updated the tests cases and regexs to have matching consistent behaviour of caring about leading zeros. I also noticed the old regex allowed double zero as in `A00`, so I've tightened that up.

I've added more test cases for these.

0xDC0DE:
Don't worry about the historical tests. They are not required, and I believe not cause by your PR at the moment.

PR #2868, opened 2023-04-12
Fix metadata-technology to handle join mode

ben-elttam:
- Before this fix it would match the inner rules of a join mode rule, but these are not allowed to have the metadata field, and if added to satisify this rule, then it would fail validation.
- Now, for join mode, it will match only the outer rule and ignore the inner rules.


0xDC0DE (approved):

0xDC0DE:
Hi @ben-elttam, Thanks for your contribution! I saw your message on the community slack about the matching range. Would you mind adding `focus-metavariable` to your rule where appropriate?

ben-elttam:
> Hi @ben-elttam, Thanks for your contribution! I saw your message on the community slack about the matching range. Would you mind adding `focus-metavariable` to your rule where appropriate?

Hi @0xDC0DE, I raised https://github.com/returntocorp/semgrep/issues/7532 for that.

This PR is to fix the metadata-technology rule from falsely reporting that the sub-rules of a join rule should have a metadata.technology field, which isn't valid according to the schema. If added, the rule will then fail validation.

I could add focus-metavariable to the name of the rules, or to the metadata field, which might be missing.

PR #2869, opened 2023-04-12
Fix languages to long names

### LINKED ISSUE #2754: RFC: Should the language be javascript (long) or js (short)?
**Describe the bug**

Semgrep rules specify the language(s), and this can be either long form (e.g. `javascript`), or short form (e.g. `js`).

Which should be preferred?
(Considering: Linting rules / checks in CI could be advising one over the other).

**To Reproduce**

Write a rule, similar to one of the many rules like `javascript/lang/best-practice/assigned-undefined.yaml`

```yaml
rules:
  - id: my-new-rule
    languages:
      - javascript
      - typescript
```

Run semgrep (meta) against the rule for linting.

`yaml/semgrep/consistency/lang-consistency-js.yaml` complains the language should be `js` not `javascript`.

> Found 'javascript' in language config which diverges from semgrep.dev normalization.
    Please use 'js' instead.

**Expected behavior**

Consistency between the registry rules (returntocorp/semgrep-rules) and the `yaml/semgrep/consistency/lang-consistency-*` rules.

The consistency rules prefer short form, but the majority of the registry uses the long form.

Looking at returntocorp/semgrep-rules there are 17 `js` vs 314 `javascript` rules and 18 `ts` vs `207` typescript rules.

This was calculated using the snippet:

```
for x in $(fd '\.ya?ml$' --exclude  '*.test.*' ); do yq '.rules | map(.languages) | flatten' "$x"; done | jq -s 'flatten | group_by(.) | map({"\(first)": length}) | add'
```

**Priority**
How important is this to you?
- [X] P3: What does return2corp and the community think?

**Additional context**

As a community poll, please vote for long form by clicking the 🚀 and the short form by clicking the 🎉

ben-elttam:
- Fixes #2754

### DIFF
--- a/javascript/ajv/security/audit/ajv-allerrors-true.yaml
+++ b/javascript/ajv/security/audit/ajv-allerrors-true.yaml
@@ -19,8 +19,8 @@ rules:
     impact: LOW
     confidence: LOW
   languages:
-  - js
-  - ts
+  - javascript
+  - typescript
   severity: WARNING
   pattern-either:
   - pattern: |
--- a/javascript/apollo/security/apollo-axios-ssrf.yaml
+++ b/javascript/apollo/security/apollo-axios-ssrf.yaml
@@ -27,7 +27,7 @@ rules:
     impact: MEDIUM
     confidence: LOW
   languages:
-  - js
+  - javascript
   severity: WARNING
   patterns:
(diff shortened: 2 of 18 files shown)


0xDC0DE (approved):
Thank you, Ben!

PR #2881, opened 2023-04-19
edit ssrf rule

colleend:
Editing this SSRF rule to have less false positives.


colleend:
@emjin @IagoAbal is the failure in semgrep-develop due to a recent benchmark test? The go rule is running fine locally, might be because my test file is large

IagoAbal:
I can reproduce it locally. If you invoke `semgrep-core` directly it works fine and runs fast. The problem happens if you use the CLI.

IagoAbal:
Oh, I see, the "impossible" happened:
> Fatal error: exception Common.Impossible

IagoAbal:
It happens if you add `-json` to `semgrep-core`, plain-text ouput works fine.

IagoAbal:
Found the issue, working on a PR.

go/lang/security/injection/tainted-url-host.yaml
```diff
@@ -53,4 +40,58 @@ rules:
...
+        "https://" + $X
+    - pattern-not: |
+        "http://" + $X
+    - pattern-not: |
+        "//" + $X
+    - pattern-not: |
+        $X + $F
+    - focus-metavariable: $X
+  pattern-sinks:
+  - requires: FORMAT or (not ADD)
```
IagoAbal: Should this be `FORMAT and (not ADD)` ?

colleend: good question, it should be "or" because we are looking for something that has string concat via format strings with something looking like `http/https` OR something that is NOT `http://www.example.com/` + `<user input>` (doesn't matter bc its only path that is being controlled). 

It's essentially two different options of concat that we're looking at: via format string and via `+` operator, and we want either option to work.

colleend:
Thank you for looking into this! I was very confused 😄

go/lang/security/injection/tainted-url-host.yaml
kurt-r2c: additional clarification - does this negation construct for labels behave intuitively? 

consider:

```go
  err_redirect := "https://app.foo.com/error"
  // stuff happens
  if err != nil {
    http.Redirect(w, r, err_redirect, http.StatusSeeOther)
  } 
```

I'm assuming the redirect string doesn't get tainted because it's not a source (doesn't match `INPUT` requirements). So, things with the `INPUT` label (and nothing else) can make it into the sink as long as they don't get the `ADD` label?

This *feels* like a sanitizer construction but I understand the use of the labeling here, at least intuitively.  Is there an equivalent sanitizer construction? Is it more performant to use sanitizers vs this sort of label conditional?

kurt-r2c (approved):

go/lang/security/injection/tainted-url-host.yaml
IagoAbal: You would need to add something positive to have `INPUT or (SOMETHING and not ADD)` because otherwise the sink will never be evaluated, taint-mode doesn't explicitly track the absence of a label. If there is no taint arriving at the sink, the sink won't match. There needs to be _some_ label arriving at the sink, and then you can use `not X` as a sanitizer to remove cases.

BTW the issue with semgrep-core was fixed here returntocorp/semgrep#7636!

colleend: These are very good points that ya'll have brought up -- @kurt-r2c I will look into trying this with sanitizers today, and @IagoAbal thank you for clarification with the taint labels. I'll try messing around with `FORMAT or (INPUT and not ADD)`. Thanks for fixing the core issue!

haha I'm vaguely glad I did something in an unexpected way because it's good to have discovered the crash 😛

colleend:
@kurt-r2c changed this to use sanitizers and a positive label -- does this feel more intuitive/readable to you?

go/lang/security/injection/tainted-url-host.yaml
```diff
@@ -53,4 +40,53 @@ rules:
...
+    - pattern-either:
+      - pattern: $X + $F
+      - pattern: $X
+    - focus-metavariable: $X
+  pattern-sanitizers:
+  - patterns:
+    - pattern: $F + $X
+    - metavariable-regex: 
+        metavariable: $F
+        regex: ".*//[a-zA-Z0-10]+..*"
```
kurt-r2c, line 68: this regex make a screm so loud

kurt-r2c:
```suggestion
        regex: ".*//[a-zA-Z0-10]+\\..*"
```
I'm assuming the `.` after the greedy quantifier is meant to be an explicit `.` which means you're missing an escape here

kurt-r2c (changes requested):
one regex correctness thing, otherwise this looks good.

The other way was also reasonable, now that I've thought about it a minute, but I think we should prefer sanitizers explicitly when building what is effectively "sanitizer" behavior.

go/lang/security/injection/tainted-url-host.yaml
colleend: ah yes it definitely is meant to be explicit `.` thank you!

colleend: added some not-insides to exclude cases where the direct `$X` matches  a fmt-string line and proceeds to override the cases where `fmt.Println("example.com%q)` are not matched. I tried other things like sanitizers but this worked best... not sure if there's a better solution 🤔

kurt-r2c (approved):

PR #2883, opened 2023-04-21
Add rule jackson-unsafe-deserialization (Java)

lfama:
This rule checks if default typing (which can lead to RCE in conjunction with unsafe deserialization) is enabled (globally or via Jackson annotations).
References:
- https://swapneildash.medium.com/understanding-insecure-implementation-of-jackson-deserialization-7b3d409d2038
- https://cowtowncoder.medium.com/on-jackson-cves-dont-panic-here-is-what-you-need-to-know-54cd0d6e8062
- https://snyk.io/blog/java-json-deserialization-problems-jackson-objectmapper/

### DIFF
--- a/java/lang/security/jackson-unsafe-deserialization.yaml
+++ b/java/lang/security/jackson-unsafe-deserialization.yaml
@@ -23,13 +23,13 @@ rules:
               - pattern: $OM.readValue($JSON, $CLASS.class);
     message: >-
       When using Jackson to marshall/unmarshall JSON to Java objects,
-       enabling default typing is dangerous and can lead to RCE. If an attacker
-       can control `$JSON` it might be possible to provide a malicious JSON which
-       can be used to exploit unsecure deserialization. In order to prevent this
-       issue, avoid to enable default typing (globally or by using "Per-class"
-       annotations) and avoid using `Object` and other dangerous types for member
-       variable declaration which creating classes for Jackson based
-       deserialization.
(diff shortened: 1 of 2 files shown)


java/lang/security/jackson-unsafe-deserialization.yaml
```diff
@@ -0,0 +1,53 @@
+rules:
+  - id: jackson-unsafe-deserialization
+    patterns:
+      - pattern-either:
+          - patterns:
+              - pattern-inside: |
+                  ObjectMapper $OM = new ObjectMapper(...);
```
kurt-r2c, line 7: Recommend multiple `pattern-inside` here - I think you want to ensure that the ObjectMapper is instantiated *and* default typing is enabled, e.g.

```yaml
- patterns:
  - pattern-inside: |
      ObjectMapper $OM = new ObjectMapper(...);
      ...
  - pattern-inside: |
      $OM.enableDefaultTyping();
      ...
  - pattern: $OM.readValue($JSON,...);
```

see https://semgrep.dev/s/PzPv for an example w/test case where object typing is set after an ObjectMapper read

kurt-r2c (changes requested):

java/lang/security/jackson-unsafe-deserialization.yaml
lfama: That's correct. Thank you for the review!

kurt-r2c (approved):

PR #2893, opened 2023-04-27
update formatted-sql-string to support concatenated / mutable String objects and a few additional sinks 

infractal:
- Missing $ in metavariable SOURCE.
- Added addBatch / nativeSQL / create(.*) / prepare as potential sink methods for sqli   (execute(), executeQuery(), executeUpdate() already covered by regex Statement sink methods:   execute, executeQuery, executeUpdate, ... EntityManager sink methods:   createNativeQuery, createQuery, createStoredProcedureQuery Connection sink methods:   nativeSQL, prepareCall
- Add prepared statements (can be used incorrectly) as sinks.
- Handle string concats + mutable StringBuffer/StringBuilder as propagators. (does not handle sbf.append().append() ... )

### DIFF
--- a/java/lang/security/audit/formatted-sql-string.yaml
+++ b/java/lang/security/audit/formatted-sql-string.yaml
@@ -55,7 +55,7 @@ rules:
       - pattern: (EntityManager $EM).$SQLFUNC(...)
     - metavariable-regex:
         metavariable: $SQLFUNC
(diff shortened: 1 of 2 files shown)


kurt-r2c:
resolved by https://github.com/returntocorp/semgrep-rules/pull/2892

PR #2906, opened 2023-05-05
Improved jackson-unsafe-deserialization rule to match more cases

lfama:
Now the rule matches also when the `ObjectMapper` is created inside a different method (i.e. some initialization method) and then used in another method.
I've also modified the Java test file to match the new case.


java/lang/security/jackson-unsafe-deserialization.yaml
```diff
@@ -22,6 +22,22 @@ rules:
                   metavariable: $TYPE
                   regex: (Object|Serializable|Comparable)
               - pattern: $OM.readValue($JSON, $CLASS.class);
+          - patterns:
+              - pattern-inside: |
+                  class $CLASS {
+                    ObjectMapper $OM;
+                    $INITMETHODTYPE $INITMETHOD(...) {
+                      ...
+                      $OM = new ObjectMapper();
```
colleend, line 33: include `...` between these two lines, otherwise if there's a variable defined between these two lines, it would not match

colleend (changes requested):
thanks for your contribution! I left one small comment, otherwise this looks good to me. I can also help push the change if that's easier.

java/lang/security/jackson-unsafe-deserialization.yaml
lfama: Yes, I also agree that's better. Please go ahead and push the change! Thanks!

colleend (approved):
lgtm

PR #2907, opened 2023-05-08
Adding XXE rules for SAXParserFactory

coheigea:
See https://cheatsheetseries.owasp.org/cheatsheets/XML_External_Entity_Prevention_Cheat_Sheet.html#jaxp-documentbuilderfactory-saxparserfactory-and-dom4j
It duplicates the existing DocumentBuilderFactory rules.

### DIFF
--- a/java/lang/security/audit/xxe/documentbuilderfactory-disallow-doctype-decl-false.java
+++ b/java/lang/security/audit/xxe/documentbuilderfactory-disallow-doctype-decl-false.java
@@ -44,6 +44,12 @@ class GoodDocumentBuilderFactory {
         dbf.setAttribute(XMLConstants.ACCESS_EXTERNAL_SCHEMA, "");
     }
 
+    public void GoodSAXParserFactory() throws  ParserConfigurationException {
+        SAXParserFactory spf = SAXParserFactory.newInstance();
+        //ok:documentbuilderfactory-disallow-doctype-decl-false
+        spf.setFeature("http://apache.org/xml/features/disallow-doctype-decl", true);
+    }
+
 }
 
 class BadDocumentBuilderFactory{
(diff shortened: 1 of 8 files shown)


java/lang/security/audit/xxe/saxparserfactory-external-parameter-entities-true.yaml
```diff
@@ -0,0 +1,37 @@
...
+    references:
+    - https://semgrep.dev/blog/2022/xml-security-in-java
+    - https://semgrep.dev/docs/cheat-sheets/java-xxe/
+    - https://blog.sonarsource.com/secure-xml-processor
+    category: security
+    technology:
+    - java
+    - xml
+    cwe2022-top25: true
+    cwe2021-top25: true
```
colleend: this rule will actually match the same things as https://github.com/returntocorp/semgrep-rules/blob/develop/java/lang/security/audit/xxe/documentbuilderfactory-external-parameter-entities-true.yaml: (https://semgrep.dev/s/55e3) because the metavariable can match anything 👍

java/lang/security/audit/xxe/saxparserfactory-disallow-doctype-decl-false.yaml
```diff
@@ -0,0 +1,73 @@
+rules:
```
colleend: same with this rule!: https://semgrep.dev/s/Rd4K

java/lang/security/audit/xxe/saxparserfactory-external-general-entities-true.yaml
```diff
@@ -0,0 +1,37 @@
...
+    cwe2022-top25: true
+    cwe2021-top25: true
+    subcategory:
+    - vuln
+    likelihood: LOW
+    impact: HIGH
+    confidence: HIGH
+  message: >-
+    External entities are allowed for this SAXParserFactory.
+    This is vulnerable to XML external entity attacks. Disable this by setting the feature "http://xml.org/sax/features/external-general-entities"
```
colleend: same with this rule 👍

colleend (changes requested):
Thank you for your contribution! As some of these rules are the `documentbuilderfactory` rules with just the metavariable names changed, they will match the same things. I'd recommend just leaving the `saxparserfactory-disallow-doctype-decl-missing` -- thanks for the rules contribution, this is great!

coheigea:
Thanks @colleend for the review. What do you think about changing the existing DocumentBuilderFactory message from:
"DOCTYPE declarations are enabled for this DocumentBuilderFactory."
to something like:
"DOCTYPE declarations are enabled for $DBFACTORY"
Otherwise running the rules on SAXParserFactory instances will report incorrectly that a DocumentBuilderFactory is present. Is there a way to output the class of $DBFACTORY instead?

colleend:
Great suggestion! That can definitely be done -- metavariables can be used in the messages as well as in the yaml rule pattern, so changing the message to reference the metavariable itself should work 👍 Would you mind removing the duplicate rules and then changing the relevant rule messages? Thank you so much for your contributions! 🙏

 Here's a possible message edit: https://semgrep.dev/s/Xdg4

coheigea:
@colleend Please review again, I also added some tests into the existing java classes for SAXParserFactory for the existing rules.

colleend (approved):
Thank you for your contribution! We appreciate it! It looks good to me.

PR #2908, opened 2023-05-09
Add swift insecure random rule

LewisArdern:
This introduces insecure random in swift, this rule finds all the places `SecCopyRandomBytes` is not used.


colleend (approved):
lgtm

PR #2910, opened 2023-05-10
Remove interfile for label and update rule for speed improvement

LewisArdern:
taint labels are still not yet interfile ready so removing the flag. 

Iago saw a 3x speed increase in this rule with this change


colleend (approved):
FINALLYYYYY

PR #2916, opened 2023-05-12
Pattern parse error in rule clojure.lang.security.use-of-md5.use-of-md5

jonny64:
https://github.com/returntocorp/semgrep/issues/7654 HOTFIX


ievans:
See comment https://github.com/returntocorp/semgrep/issues/7654#issuecomment-1546976898

PR #2922, opened 2023-05-19
XInclude processing should be disabled?

infractal:
dbf.setXIncludeAware(false) is best practice.

Also the dbf.setXXX(...) lines appearing in a different order could trigger a false positive.
A feature to match a number of statements (without side-effects) against a set (disregarding order) would be useful.


0xDC0DE:
Hi @infractal !

In [our research](https://semgrep.dev/docs/cheat-sheets/java-xxe/#:~:text=%E2%9C%85-,setXIncludeAware,-(false)) we found no additional security being provided by `setXIncludeAware`. Do you have any example code and payload where this feature provides additional protection? I would be very interested in that!

Regarding the updates to these rules, you are right that they should not be order dependent like they are here! And the rules should probably just ignore the `setXIncludeAware` and its value.

It's on our to do list to clean up these XXE rules, but it's not high priority right now.

We have a set of accurate Java XXE rules in the pro rules, and we have to decide how to continue with these community contributed rules in parallel.

The OWASP rules in particular are linked to from an OWASP cheat sheet: https://cheatsheetseries.owasp.org/cheatsheets/XML_External_Entity_Prevention_Cheat_Sheet.html#semgrep-rules

Removing these outright would not be the most community-friendly move. I think a potential solution would be to recreate the whole set of linked XXE rules and replace the links in the OWASP cheat sheet with the more accurate versions. 

The scala rule however, should probably be converted to use `taint-labels`, I'll talk to the team to see how we can get started on this.

infractal:
Yes, given the results from your research it makes most sense to ignore setXIncludeAware. Or at least, since the default value is false, not have the rules rely on it being set to true. 

Given its previous inclusion in research papers, perhaps it only makes a difference in older versions of libraries, but as of now I have no examples of specific versions where that applies. Will update if I find anything.

0xDC0DE:
@infractal , just to keep you in the loop. A discussion is happening about moving this particular set of pro rules to open source. If that happens I will also be updating the OWASP cheat sheets. 🤞

Issue #2914, opened 2023-05-11
ActiveRecord SQL injection false positive when using `slice`


nightpool:
**Describe the bug**

Currently, semgrep-rules flags ActiveRecord code like `User.where(params[:query])` (which allows for a partial SQL string to be inserted) but allows code like `User.where(user_id: params[:user_id])` (where ActiveRecord will properly parameterize the query). However, it also falsely flags cases where a `where` hash is used (and the query is parameterized), but that hash is constructed using `slice` instead of using a hash literal. See below:

**To Reproduce**

```ruby
  def ok_test6
    user = User.where(user_id: params[:user_id])[0]
    user = User.where(params.slice(:user_id))[0]
  end
```

Although the two lines are identical, the second line is flagged by semgrep and the first line is not.

**Expected behavior**

Neither line is flagged.

**Priority**
How important is this to you?
- [x] P2: annoying but not blocking me

**Additional Context**

This resulted in only a small number of false-positives in our codebase, and although any false-positive is annoying, and this one was particularly scary since the code *looks* dangerous on first glance, it was relatively easy to resolve.

0xDC0DE:
Thanks @nightpool for reporting this false positive! The update for the rule will be merged soon.

LewisArdern:
Looks like this should be fixed with the recent merge

PR #2925, opened 2023-05-23
Fix false positive in ruby rails tainted-sql-string

0xDC0DE:
Fix for https://github.com/returntocorp/semgrep-rules/issues/2914

### DIFF
--- a/ruby/rails/security/injection/tainted-sql-string.yaml
+++ b/ruby/rails/security/injection/tainted-sql-string.yaml
@@ -38,8 +38,11 @@ rules:
     - pattern-either:
       - patterns:
         - pattern-either:
-          - pattern: |
-              $RECORD.where($X,...)
+          - patterns:
+            - pattern: |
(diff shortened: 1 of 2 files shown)


ruby/rails/security/injection/tainted-sql-string.yaml
```diff
@@ -38,8 +38,11 @@ rules:
     - pattern-either:
       - patterns:
         - pattern-either:
-          - pattern: |
-              $RECORD.where($X,...)
+          - patterns:
+            - pattern: |
+                $RECORD.where($X,...)
+            - pattern-not: |
+                $RECORD.where($PARAMS.slice(...), ...)
```
inkz, line 45: why not setting `$PARAMS.slice` as sanitizer?

kurt-r2c: maybe file as future enhancement?

kurt-r2c (approved):

PR #2927, opened 2023-05-24
Fix test error for returntocorp/semgrep#6780

ihji:
returntocorp/semgrep#6780


ihji:
https://github.com/returntocorp/semgrep/pull/7869

0xDC0DE (approved):

IagoAbal:
I assume we're fine merging despite the "semgrep-rules-test-historical" failure. In the future I think we should put the `# todoruleid` annotations in the right place, and the test runner should have some flag so that a finding where there is a `todoruleid` is not reported as an error.

PR #2934, opened 2023-05-29
Fix FPs in python empty hardcoded password

nightshiba:
Here's a fix for the FPs that I've encountered. An example of such a FP:
```python
def __init__(self, password=""):
        self.domain = domain
        if not password:
            password = getpass.getpass()
```


inkz (approved):

inkz:
thank you!

PR #2935, opened 2023-05-30
using type inference in sanitizers

colleend:
CC: @nmote 

* replacing sanitizers that look like `$X == $Y` and `$X != $Y` with (boolean $X) in Java
* adding tests to make sure this works


nmote:
Nice! Makes sense that the historical test failed, since type inference for binary operators is recent.

colleend:
I did take a look at this, and the reason why tests are failing is bc of another change (not bc of these rules 💯 )

inkz (approved):
amazing

PR #2937, opened 2023-05-31
update sources and add int as sanitizer

colleend:
* updates all spring sources to filter out primitive types
* updates sanitizers with bool to also sanitize ints


inkz (approved):
lgtm

PR #2952, opened 2023-06-12
Add pattern for a popular log4j package import

ilyaglow:
Hey,

I noticed that semgrep doesn't detect the code example from [Lunasec's original article](https://www.lunasec.io/docs/blog/log4j-zero-day/#example-vulnerable-code) and I decided to add another pattern for the `org.apache.logging.log4j` package. This import name seems to be [pretty popular on github](https://github.com/search?q=lang%3AJava+%22import+org.apache.logging.log4j%22&type=code).

I haven't found a best practice in the docs for the case when two test files are needed for a single rule, so I'm not sure if I named the test file properly. Let me know if I did it in the wrong way.

While I understand that the current rule is not perfect and prone to false-positives, I hope this PR is still going to be useful for someone.

Cheers.

### DIFF
--- a/java/log4j/security/log4j-message-lookup-injection-lunasec.java
+++ b/java/log4j/security/log4j-message-lookup-injection-lunasec.java
@@ -1,0 +1,33 @@
+// Source: https://www.lunasec.io/docs/blog/log4j-zero-day/#example-vulnerable-code
+// Copyright © LunaSec
+import org.apache.logging.log4j.LogManager;
+import org.apache.logging.log4j.Logger;
+
+import java.io.*;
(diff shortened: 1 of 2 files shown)


LewisArdern:
Appreciate the PR @ilyaglow we plan to deprecate this rule since our supply chain product already finds it, which is where this rule should live in the semgrep eco-system. Without verifying the version of the package this rule is useless to most of the consumers of semgrep and therefore we have decided to remove it.

You can happily use a fork of this rule for your own testing purposes outside of the semgrep registry similar to https://github.com/elttam/semgrep-rules

PR #2955, opened 2023-06-13
Match app-specific class/module when configuring Rails session-store

philipturnbull:
Rails apps can derive their own custom class from `Rails::Application`, then use that when configuring the app. An example from `OWASP/railsgoat`:

```
module Railsgoat
  class Application < Rails::Application
    ...
  end
end

...

Railsgoat::Application.config.session_store :cookie_store, key: "_railsgoat_session", httponly: false
```

---

Also, `httponly` doesn't have to be the last argument so support that too.


colleend (approved):
awesome work!!

kurt-r2c (approved):

Issue #2959, opened 2023-06-14
Artifactory password falsely triggering on yarn lock file


ChristianAlexander:
**Describe the bug**
After it was recently updated (#2956), the `detected-artifactory-password` rule started falsely flagging the following line within a yarn.lock file of a project I work on:

```
integrity sha512-AP1AyUTbi2szylgr+O0OB7gkIxEGzySLITZ2GpsaoX72YMCGI2jYAc+WUhPfvUnZYiauF4zTnN4V4TGuvFjJlw==
```

**To Reproduce**
Run semgrep with the [rule](https://semgrep.dev/r/generic.secrets.security.detected-artifactory-password.detected-artifactory-password) active, and observe that the string above is detected as an artifactory password, despite the [attempt](https://github.com/returntocorp/semgrep-rules/pull/2956/files#diff-97df62bc78881ac1e067dc17b2de5965d04d727de949e35ecf7c47a73eda23dbR10-R15) to exclude `sha(128|256|512).*` in the rule.

**Expected behavior**
This should not have triggered the artifactory password rule.

**Priority**
How important is this to you?
- [ ] P0: blocking me from making progress
- [ ] P1: this will block me in the near future
- [x] P2: annoying but not blocking me

**Screenshots**

**Desktop (please complete the following information):**

**Smartphone (please complete the following information):**

**Additional context**

enncoded:
Closing because a fix was merged. Reopen if it's still causing you issues! Thanks :)

PR #2961, opened 2023-06-14
fix FP from #2959

kurt-r2c:
addresses #2959


colleend (approved):
niceeee

kurt-r2c:
Historical test failing due to missing features from aliengrep in historical version. Given that this is a fix for a customer-reported issue (and folks *should* be on latest) I'm willing to take 1 additional FP until aliengrep ages into historical.

PR #2962, opened 2023-06-14
adding type inf for sql rules

colleend:
* changing `sanitizers: - pattern: (boolean $X)` -> taint options
* adding taint options to sql rules
* added a test that should work in pro but doesn't in regular semgrep


kurt-r2c (approved):

kurt-r2c:
Historical tests are failing due to taint options. These will age out as we release.

PR #2965, opened 2023-06-16
Deprecate some Ruby rules

philipturnbull:
These rules do not check the gem versions so they are false-positive prone.


colleend (approved):
lgtm, thanks phil!

PR #2966, opened 2023-06-16
Deprecate CVE-2021-44228 rule

philipturnbull:
This rule does not check the version of log4j so it is prone to false positives.


colleend:
Do we currently have an alternative to this rule that checks for log4j?

kurt-r2c:
@colleend yes

kurt-r2c (approved):

eggdkk:
> @colleend是的

I didn't find a rule that can detect log4j instead, can you remind me of it, thanks!

PR #2973, opened 2023-06-21
feat: metacheck for poorly indented rules

brandonspark:
## What:
This PR adds in a metacheck which detects poorly indented YAML, such as
```
patterns:
  -pattern: |
    foo
```
Notably, this is an invalid Semgrep rule, because in YAML, this is actually interpreted as a map from the string `"-pattern"` to the string `"foo"`.

## Test plan:
Automated tests

### DIFF
--- a/yaml/semgrep/key-indentation.yaml
+++ b/yaml/semgrep/key-indentation.yaml
@@ -23,4 +23,4 @@ rules:
         - focus-metavariable: $KEY
         - metavariable-regex:
             metavariable: $KEY
-            regex: "^-(\w)$"
+            regex: "^-(\w*)$"
--- a/yaml/semgrep/key-indentation.yaml
+++ b/yaml/semgrep/key-indentation.yaml
@@ -23,4 +23,4 @@ rules:
         - focus-metavariable: $KEY
         - metavariable-regex:
             metavariable: $KEY
-            regex: "^-(.*)$"
+            regex: "^-(\\w)$"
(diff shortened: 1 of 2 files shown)


yaml/semgrep/key-indentation.yaml
```diff
@@ -0,0 +1,26 @@
...
+      - patterns:
+        - pattern-inside: |
+            rules: ...
+        - pattern: |
+            $KEY: >-
+              $VALUE
+        - focus-metavariable: $KEY
+        - metavariable-regex:
+            metavariable: $KEY
+            regex: "^-(.*)$"
```
enncoded: Wouldn't this detect on anything after the dash, including spaces and stuff? Did you mean to do `"^-(\w)*$"`?

enncoded (changes requested):

yaml/semgrep/key-indentation.yaml
brandonspark: @enncoded ooh, i think you're right! i was thinking it wouldn't matter, since the range of `$KEY` should be just that of the entire key to the pair, but it doesn't hurt to change -- fixed!

enncoded: The regex as you changed it to is invalid, but doing `regex: ^-(\w*)$` works. Also update the test to include an `#ok: ruleid` check. Feel free to copy/paste from https://semgrep.dev/playground/s/50N0

enncoded (approved):

PR #2975, opened 2023-06-22
Update r2c/Return to Corp to Semgrep

armchairlinguist:
Administrative update to align with company rename.

Generally (with a few exceptions):
https://r2c.dev/slack -> https://go.semgrep.dev/slack
https://r2c.dev -> https://semgrep.dev
@r2c.dev -> @semgrep.com
Return To Corporation -> Semgrep, Inc.
r2c -> Semgrep **or** Semgrep, Inc. as appropriate

### DIFF
--- a/CODE_OF_CONDUCT.md
+++ b/CODE_OF_CONDUCT.md
@@ -55,7 +55,7 @@ further defined and clarified by project maintainers.
 ## Enforcement
 
 Instances of abusive, harassing, or otherwise unacceptable behavior may be
-reported by contacting the project team at semgrep@semgrep.dev. All
+reported by contacting the project team at support@semgrep.com. All
 complaints will be reviewed and investigated and will result in a response that
 is deemed necessary and appropriate to the circumstances. The project team is
 obligated to maintain confidentiality with regard to the reporter of an incident.
--- a/CODE_OF_CONDUCT.md
+++ b/CODE_OF_CONDUCT.md
@@ -55,7 +55,7 @@ further defined and clarified by project maintainers.
 ## Enforcement
 
 Instances of abusive, harassing, or otherwise unacceptable behavior may be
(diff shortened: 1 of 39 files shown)


CODE_OF_CONDUCT.md
```diff
@@ -55,7 +55,7 @@ further defined and clarified by project maintainers.
 ## Enforcement
 
 Instances of abusive, harassing, or otherwise unacceptable behavior may be
-reported by contacting the project team at semgrep@r2c.dev. All
+reported by contacting the project team at semgrep@semgrep.dev. All
```
armchairlinguist: I'm not sure if this is the correct address in this case; should it be semgrep@semgrep.com? Or something else?

werrett: Err. Neither of these addresses exists! 

I would suggest we make it `support@semgrep.com`

armchairlinguist: Thanks for checking this, @werrett ! I don't know if that address would be the most efficient, but if there's not a dedicated one currently, we could use it as the best option for now, and consider revising later.

p4p3r: There is still an instance of email address to be changed, am I right?

raj-r2c: @armchairlinguist @werrett  support@semgrep.com already exists and is used by customers to open support cases today. That should be the address we should include for all customer facing error messages.

armchairlinguist: I updated this, but I did want to briefly note it's not an error message - it's guiding text that's part of this repository's Code of Conduct file. I think this is good for now, but it's possible we'd want a dedicated address sometime in the future.

p4p3r (approved):
LGTM

PR #2976, opened 2023-06-22
fix: rule after matching fields with stmts

brandonspark:
After https://github.com/returntocorp/semgrep/pull/8052, this has the effect of making statements patterns work on sequences of fields (such as class bodies, etc). This has the consequence of adding this finding to one of our rules.

CI fails, but that's OK, it will succeed after it's merged.


enncoded (approved):

brandonspark:
@enncoded is it fine to be failing `semgrep-rules-test-historical`, since past versions of Semgrep would not produce this finding?

enncoded:
ya

PR #2984, opened 2023-06-26
fix: update regex capture groups

brandonspark:
This PR updates some rules to use the named regex capture groups, instead of the to-be-deprecated numerical offsets.

See https://github.com/returntocorp/semgrep/pull/8115


p4p3r (approved):
LGTM

PR #2985, opened 2023-06-26
chore: use deepok to mark lines as ok when run with the pro engine

emjin:
We have deepruleid to mark lines as matching only with the pro engine, but we didn't have anything to mark lines as ok only with the pro engine. deepok is intended for that.

This PR updates annotations to use deepok.

### DIFF
--- a/java/spring/security/injection/tainted-sql-string.java
+++ b/java/spring/security/injection/tainted-sql-string.java
@@ -177,8 +177,7 @@ class Test {
   public ResultSet ok7(@RequestBody String name, Foo foo) {
         var v = foo.getBars(name).get(0).getX();
         String sql = "SELECT * FROM table WHERE name = ";
-        // ok in pro engine
-        // ruleid: tainted-sql-string
+        // ruleid: deepok: tainted-sql-string
         sql += v + ";";
         Connection conn = DriverManager.getConnection("jdbc:mysql://localhost:8080", "guest", "password");
         Statement stmt = conn.createStatement();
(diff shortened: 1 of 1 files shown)


emjin:
Failing tests are validation errors, appear unrelated (they're probably due to Brandon's changes)

colleend:
yep, seems like the validation errors are unrelated and historical isn't passing bc of some type inference checks. Lgtm -- thanks emma!

colleend (approved):

p4p3r:
I merged all the latest changes in develop (including Brandon's changes to regex metavars capture), but tests are still reported as failing. Is that expected?

colleend:
nope, that's not expected -- I suspect that this is because `ruleid: deepok: ` isn't recognized by the `semgrep --test` test script. 

I can look into it for a bit 👍 

---
edit: yep: 
```
Could not parse         // ruleid: deepok: tainted-sql-string as a test annotation in file /Users/chai/Desktop/semgrep_repos/semgrep-rules/java/spring/security/injection/tainted-sql-string.java. Skipping this line
Could not parse         // ruleid: deepok: tainted-sql-string as a test annotation in file /Users/chai/Desktop/semgrep_repos/semgrep-rules/java/spring/security/injection/tainted-sql-string.java. Skipping this line
```

emjin:
Oh no, --test must check it differently. I could change the format of this annotation if that would be the simplest way

colleend:
passes develop now! I will merge :D

Issue #2990, opened 2023-06-30
[owasp.java.ssrf.java.net.url] False Negative When Detecting SSRF in the java.net.URL Sink


SaeedHashem:
**Describe the bug**
The rule `owasp.java.ssrf.java.net.url` doesn't detect the potential SSRF when the variable to hold the URL object is declared earlier in the code. This allows this rule to miss trivial cases where the code is indeed vulnerable.

**To Reproduce**
While this code will get flagged (Copied from `contrib/owasp/java/ssrf/ssrf.java`)
```java
public static void imageIO(String url) {
        try {
            URL u = new URL(url);
            ImageIO.read(u); // send request
        } catch (IOException e) {
            logger.error(e.getMessage());
        }

    }
```

This doesn't get flagged
```java
public static void imageIO(String url) {
        URL u = null ;
        try {
            u = new URL(url);
            ImageIO.read(u); // send request
        } catch (IOException e) {
            logger.error(e.getMessage());
        }

    }
```
**Expected behavior**
Both of examples above should get flagged by this rule.

**Priority**
How important is this to you?
- [ ] P0: blocking me from making progress
- [ ] P1: this will block me in the near future
- [X] P2: annoying but not blocking me

**Additional Context**
This was noticed when Semgrep missed an actual SSRF bug which was discovered in a manual source code review exercise. I plan to send a pull request for this, but just thought to publish this issue first in case I never get the chance to follow through.

PR #2991, opened 2023-06-30
Added a 2nd pattern to owasp.java.ssrf.java.net.url for wider coverage

SaeedHashem:
This is the pull request to address the issue reported here #2990. I have added an additional pattern to account for the case that was missed by this rule in the `ssrf.yaml` file as well as the respective test case code in the `ssrf.java` file.


p4p3r:
Hey @SaeedHashem thanks for you contribution!
I suggested just a tiny change (it will fix the failing `rules-test-develop` check).
Otherwise it looks good!

0xDC0DE (approved):

PR #2993, opened 2023-07-01
Add missing syntactic elements reported by new parsers

mjambon:
This allows https://github.com/returntocorp/semgrep/pull/8190 to pass its tests.

Tree-sitter-based parsers insert missing tokens such as a semicolon or a closing parenthesis when it thinks it's missing. They've always done this but until now, it was not reported. Now, each time a fake "MISSING" node is found in the parse tree, we produce an error and report the input as partially parsed. On semgrep-rules and in the semgrep repo tests, we found these kinds of errors:
* missing semicolon or similar in a pattern that otherwise works: I fixed these in semgrep-rules. This can be a problem for rules outside of the default rule registry. The plan is to not fail but produce a warning when encountering such a broken pattern so as to play well with legacy rules.
* missing semicolon or similar in a test target: I corrected these and they should not occur in user code since they're invalid programs.
* parser bug: these are old bugs that were invisible and will now be reported. However, this still results in the same well-formed  AST as before and the same Semgrep findings (regardless of whether they're correct findings).


ruby/lang/security/hardcoded-secret-rsa-passphrase.rb
```diff
@@ -1,63 +1,64 @@
...
+      end
+    end
+
+    def ok
+      key_pem = File.read @keypem
+      #ok: hardcoded-secret-rsa-passphrase
+      key = OpenSSL::PKey::RSA.new key_pem, ENV['SECRET']
+    end
+  end
+end
```
mjambon, line 64: This is a full reindentation + the addition of a missing `end`.

terraform/gcp/security/gcp-dataproc-cluster-public-ip.yaml
```diff
@@ -4,18 +4,19 @@ rules:
...
+        cluster_config {
+          gce_cluster_config {
+            ...
+            internal_ip_only = true
+            ...
+          }
+        ...
         }
-      ...
       }
```
mjambon, line 20: There was a missing closing brace `}` and bad indentation.

dockerfile/correctness/multiple-entrypoint-instructions.yaml
```diff
@@ -6,15 +6,10 @@ rules:
...
-          metavariable: $ENTRYPOINT
-          patterns:
-            - pattern: |
-                ENTRYPOINT ...
-      - focus-metavariable: $ENTRYPOINT
+          ENTRYPOINT ...
     message: >-
-      Multiple ENTRYPOINT instructions were found. Only the last one will take effect.
+      Multiple ENTRYPOINT instructions were found. Only the last one
+      will take effect.
```
mjambon, line 20: I don't know why the previous version was working because `$ENTRYPOINT ...` would parse as a shell command, not a dockerfile instruction. The new version is parsed and works as expected but the result is the whole region starting and ending with `ENTRYPOINT` and no longer just one `ENTRYPOINT` line.

brandonspark (approved):
looks good to me!

dockerfile/correctness/multiple-entrypoint-instructions.yaml
p4p3r: I remember changing this rule a while ago specifically to restrict the matched range, because we saw that some users were getting very long ranges (which tented to produce "unstable" findings when anything changed in the range).
What about doing this here to restrict the range? It seems to work.
```
    patterns:
      - pattern: |
          ENTRYPOINT ...
          ...
          $ENTRYPOINT ...
      - metavariable-regex:
          metavariable: $ENTRYPOINT
          regex: (?i)(ENTRYPOINT)
      - focus-metavariable: $ENTRYPOINT
```

p4p3r (dismissed):
See my comment on the dockerfile rule.

dockerfile/correctness/multiple-entrypoint-instructions.yaml
mjambon: I had to look at the grammar definition to figure out what's going on :-|

The line `$ENTRYPOINT ...` confused me. It should probably be rejected by the parser. What's expected to work is either a [metavariable representing a whole dockerfile instruction or an ellipsis representing a sequence of instructions](https://github.com/returntocorp/ocaml-tree-sitter-semgrep/blob/4374124a9fce73c9c6a8fc23d52edde7c985f259/lang/semgrep-grammars/src/semgrep-dockerfile/grammar.js#L17-L18). The non-misleading way to write it is:
```
$ENTRYPOINT
...
```
i.e. `$ENTRYPOINT` captures the whole instruction `ENTRYPOINT /bin/true`, not just `ENTRYPOINT` like I thought. The ellipsis `...` is superfluous.

I pushed a revised version that restores the original behavior and clarifies what's going on.

mjambon: I pushed another change to make the solution closer to the original one (using a semgrep pattern instead of a regexp).

DrewDennison (approved):

p4p3r (approved):
LGTM

PR #2994, opened 2023-07-03
Deprecate Ruby YAML and MySQL rules

philipturnbull:
Deprecate the `yaml-parsing` and `dynamic-finders` rules because they rely on a specific version of Rails.


p4p3r (approved):
LGTM

PR #2997, opened 2023-07-06
Enhance Python eval() detection

### LINKED ISSUE #2999: Enhance Python eval() detection rule
**Describe the bug**
The current rule needs some improvements in detecting `eval()` especially if it accepts _f-string_ (`f" some string "`) as an argument. 

**To Reproduce**
Rule can't properly detect `eval()` in the following code snippets:
```python3
def eval_something(something):
    # ruleid:eval-detected
    eval(f"some_func({{something}})")
```
```python3
user_input = "command injection payload"
# ruleid:eval-detected
eval(f"some_func({user_input})")
```
```python3
# ok:eval-detected
eval("")

# ok:eval-detected
eval(f"")
```
**Expected behavior**
The rule should detect all `eval()` calls that don't have an empty string, static string, and _f-string_ without `{}` as an argument. The detection can be done by updating the pattern to use metavariable and RegEx. In this way, we can reduce the number of false positives by ignoring all `eval()` calls that don't rely on external sources.

```yaml
- pattern: eval($ARG)
    - metavariable-pattern:
        metavariable: $ARG
        patterns:
          - pattern-not-regex: (^f")([^{}]*)("$)
          - pattern-not-regex: (^")(.*)("$)
```
 Please review my pull request #2997. Thanks

**Priority**
- [ ] P1: this will block shortlyure

user1-v0:


inkz (approved):
lgtm

inkz:
Thanks @bluemarco! 
I slightly updated both test cases and the rule to be more relevant, plus I removed ` f-string without {}` test case because in that situation code fails to run and do not represent a vulnerability since it is just not working.

Also keep in mind that in this test:
```python
def eval_something(something):
    # ruleid:eval-detected
    eval(f"some_func({{something}})")
```

curly braces are escaped `{{}}` and `f"some_func({{something}})"` represents literal string , so it should not be highlighted
refer to: https://docs.python.org/3/library/string.html#format-string-syntax

PR #3001, opened 2023-07-17
update yaml/github-actions/security/run-shell-injection to remove commit_hash from being reported as injectable.

DocEmmetBrown:
`github.event.inputs.commit_hash` might not be treated as injectable, since it's not user-controlled.
This PR removes this specific event from the list being reported as injectable, and adds a test case.


DocEmmetBrown:
actually, I mis-read the action. closing this PR.

PR #3011, opened 2023-07-21
Add curl on subshell to curl-pipe-bash rule

JacobBrackett:
This is another common pattern untrusted URLs are downloaded and exec-d without verification


kurt-r2c (approved):

bash/curl/security/curl-pipe-bash.yaml
```diff
@@ -25,9 +25,11 @@ rules:
     - audit
     likelihood: LOW
     impact: LOW
-  patterns:
```
kurt-r2c, line 28: @JacobBrackett you have an indentation error here - the `patterns` top level key is now a child key of `metadata` and semgrep cannot find a valid pattern.

kurt-r2c (changes requested):

kurt-r2c (approved):

PR #3045, opened 2023-08-17
Semgrep ruby no jwt verify

outlinesecurity:


ruby/jwt/security/jwt-decode-no-verify.rb
```diff
@@ -0,0 +1,23 @@
...
+payload[:iat] = Time.now.to_i
+payload[:email] = "sample@mail.com"
+payload[:name] = "john doe"
+
+token = JWT.encode payload, secret, 'HS256'
+
+print token
+
+def bad(token, secret)
+    decoded_token = JWT.decode token, secret, false
```
0xDC0DE, line 15:
```suggestion
    # ruleid: ruby-jwt-decode-without-verify
    decoded_token = JWT.decode token, secret, false
```

ruby/jwt/security/jwt-decode-no-verify.rb
```diff
@@ -0,0 +1,23 @@
...
+    print decoded_token    
+end
+
+
+def ok(token, secret)
+    decoded_token = JWT.decode token, secret, true
```
0xDC0DE, line 21:
```suggestion
def ok(token, secret)
    # ok: ruby-jwt-decode-without-verify
    decoded_token = JWT.decode token, secret, true
```

0xDC0DE (commented):
I have added test syntax to the test code, this way we can test if this rule behaves as expected now and in the future.

PR #3046, opened 2023-08-17
Only warn if string-concatenation is used in `avoid-sqlalchemy-text`

philipturnbull:
This is rule is false-positive heavy. Use taint-mode and only alert if the argument to `sqlalchemy.text(...)` is built using string concatentation, f-strings, etc.


0xDC0DE:
The final tests are failing after a PR to Semgrep. This PR in semgrep-rules has updated the rule tests: https://github.com/returntocorp/semgrep-rules/pull/3047

If you merge develop into your branch all tests should pass and we can merge your PR.

0xDC0DE (approved):

PR #3047, opened 2023-08-17
Fix test expectations after Semgrep PR 8440

IagoAbal:
See returntocorp/semgrep#8440


javascript/lang/security/audit/unknown-value-with-script-tag.js
```diff
@@ -52,7 +52,7 @@ exports.promotionVideo = () => {
       // ok:unknown-value-with-script-tag
       var w = "<script>";
 
-      // ok:unknown-value-with-script-tag
```
IagoAbal, line 55: I think this should have been a `todoruleid` rather than an `ok`.

0xDC0DE (approved):

PR #3055, opened 2023-08-20
elixir: add atom exhausting rule

wingyplus:
Detect all Elixir `(String|List).to_atom/1` function calls and fix it with `(String|List).to_existing_atom/1`.

### DIFF
--- a/elixir/lang/correctness/atom-exhaustion.yaml
+++ b/elixir/lang/correctness/atom-exhaustion.yaml
@@ -1,0 +1,19 @@
+rules:
+  - id: atom_exhaustion
+    message: >-
+      Atom doesn't garbage by the Garbage Collector. Creating it dynamically with `String.to_atom/1` or `List.to_atom/1` can possible memory leaks if the input cannot control.
+    severity: ERROR
+    languages:
+      - elixir
+    patterns:
+      - pattern: $MODULE.to_atom($STRING)
+      - metavariable-regex:
+          metavariable: $MODULE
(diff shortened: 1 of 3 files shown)


wingyplus:
Move to correctness category from suggestion in Slack thread https://semgrep.slack.com/archives/C01LNF69F9B/p1692552800141539

elixir/lang/correctness/atom-exhaustion.yaml
```diff
@@ -0,0 +1,19 @@
+rules:
+  - id: atom_exhaustion
+    message: >-
+      Atom doesn't garbage by the Garbage Collector. Creating it dynamically with `String.to_atom/1` or `List.to_atom/1` can possible memory leaks if the input cannot control.
```
colleend:
```suggestion
      Atom values are appended to a global table but never removed. If input is user-controlled, dynamic instantiations such as `String.to_atom` or `List.to_atom` can lead to possible memory leaks. Instead, use `String.to_existing_atom` or `List.to_existing_atom`.
```

colleend (changes requested):
Thank you for submitting this! Changed the message a bit, but otherwise it looks good to me 👍

wingyplus:
> Thank you for submitting this! Changed the message a bit, but otherwise it looks good to me 👍

Thank you. The message looks better!

wingyplus:
@colleend done. :)

colleend (approved):
lgtm!

PR #3060, opened 2023-08-22
Add redos check

LewisArdern:
JS/TS didn't have a check where it was using the redos feature, so this rule includes that


colleend (approved):
lgtm!! thanks lewisss

PR #3068, opened 2023-08-24
Update XXE rule for DocumentBuilderFactory

hocnc:
Currently, The XXE rules for DocumentBuilderFactory that miss the following case.

```java
File inputFile = new File("poc.xml");
DocumentBuilder dbBuilder = DocumentBuilderFactory.newInstance().newDocumentBuilder();
Document doc = dbBuilder.parse(inputFile);
```

And I add the **documentbuilderfactory-no-unused-disallow-doctype-decl-true.yaml** rule to detect the above

### DIFF
--- a/javascript/lang/security/audit/detect-redos.js
+++ b/javascript/lang/security/audit/detect-redos.js
@@ -1,0 +1,19 @@
+// ruleid: detect-redos
+const re = new RegExp("([a-z]+)+$", "i"); 
+// ruleid: detect-redos
+const re = new RegExp(/([a-z]+)+$/, "i"); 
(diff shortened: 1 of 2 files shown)


colleend:
Thank you for your submission! It seems like the reason why this doesn't match is because there's a bug in the `owasp.java.xxe.javax.xml.parsers.DocumentBuilderFactory` rule. Would you mind submitting a fix such that it looks like this: https://semgrep.dev/playground/s/peeR and matches your test case on line 446?

hocnc:
ok let me submit a fix
Thank you!

PR #3075, opened 2023-08-28
Fix alias path traversal rule

hocnc:
If running the current `alias-path-traversal` rule with [nginx.conf](https://github.com/hestiacp/hestiacp/blob/main/src/deb/nginx/nginx.conf)

The rule will detect the snippet below is vulnerable 

```
		location /fm/ {
			alias /usr/local/hestia/web/fm/dist/;
```
The current rule: https://semgrep.dev/playground/s/QPlP
My rule: https://semgrep.dev/playground/s/g9Zn


enncoded:
Hi @hocnc!

I'm not very familiar with Nginx, what does ensuring that there's no trailing slash on the path do? I see you switched the `pattern-not-inside` with a `metavariable-regex` that enforces that.

I see
> If a location is defined by a prefix string that ends with the slash character, and requests are processed by one of [proxy_pass](http://nginx.org/en/docs/http/ngx_http_proxy_module.html#proxy_pass), [fastcgi_pass](http://nginx.org/en/docs/http/ngx_http_fastcgi_module.html#fastcgi_pass), [uwsgi_pass](http://nginx.org/en/docs/http/ngx_http_uwsgi_module.html#uwsgi_pass), [scgi_pass](http://nginx.org/en/docs/http/ngx_http_scgi_module.html#scgi_pass), [memcached_pass](http://nginx.org/en/docs/http/ngx_http_memcached_module.html#memcached_pass), or [grpc_pass](http://nginx.org/en/docs/http/ngx_http_grpc_module.html#grpc_pass), then the special processing is performed. In response to a request with URI equal to this string, but without the trailing slash, a permanent redirect with the code 301 will be returned to the requested URI with the slash appended. If this is not desired, an exact match of the URI and location could be defined like this: 
location /user/ {
    proxy_pass http://user.example.com;
}
location = /user {
    proxy_pass http://login.example.com;
}

in the [documentation](http://nginx.org/en/docs/http/ngx_http_core_module.html#location), but I don't quite understand how that's relevant. Perhaps I'm looking at the wrong thing. Do you think you could fill me in?

Thanks for the contribution :D

hocnc:
Hi @enncoded 
You can read this [article](https://www.acunetix.com/vulnerabilities/web/path-traversal-via-misconfigured-nginx-alias/)
 if the location doesn't ends with directory separator as below that allow an attacker to read file outsite the target folder.(`/data/w3/images`)
```
location /i {
    alias /data/w3/images/;
}
```
But in this case [nginx.conf](https://github.com/hestiacp/hestiacp/blob/main/src/deb/nginx/nginx.conf), `alias-path-traversal` rule detects the code snippet that is vulnerable. Even though, the location `/fm/` ends with `/`.
```
location /fm/ {
	alias /usr/local/hestia/web/fm/dist/;
```
I think maybe root cause is [indentation](https://semgrep.dev/docs/writing-rules/generic-pattern-matching/#common-pitfall-2-not-enough-indentation).

And my solution is to use regex to detect the location that does not end with '/'

enncoded:
Thanks for linking that article. Your fix looks good to me :)

Could you add that article to the references? I'll merge it after. Thanks!

enncoded (approved):

enncoded:
Perfect!

PR #3077, opened 2023-08-28
Fixed false positive with `usedforsecurity` flag in `hashlib.md5`

### LINKED ISSUE #3052: Python. Ignore md5 when used with usedforsecurity
**Is your feature request related to a problem? Please describe.**
Python [allows](https://docs.python.org/3/library/hashlib.html#hashlib.md5) to pass in `usedforsecurity` keyword when you call `hashlib.md5()` function. This is a keyword-only parameter as specified in docs by an asterisk  (`*`) in the  function's signature `hashlib.md5([data, ]\*, usedforsecurity=True)` with a default value of `True`. If a developer manually passes in a value `usedforsecurity=False` it means that he knows what he is doing and that md5 is **not** used for security, which means semgrep shouldn't be triggered by md5's usage.

**Describe the solution you'd like**
`semgrep` doesn't trigger in CI when it encounters:
```
foo = hashlib.md5(data, usedforsecurity=False)
# or
bar = md5(data, usedforsecurity=False)
```

**Describe alternatives you've considered**
I've used `# nosemgrep` but in the code it looks more like a workaround than an actual solution. MD5 is used in a legacy codebase so changing it to sha-secure codes would mean serious refactoring, which is not desirable.

**Use case**
MD5 can be used as an integrity check when passing data between services & encoding file names. Thanks to its' output's small size (even smaller than uuid), it doesn't take a lot of computation time and memory to generate and process with your eyes.

image

enncoded:
Fixes https://github.com/returntocorp/semgrep-rules/issues/3052


philipturnbull (approved):

enncoded:
Reopening to try to make license check pass

PR #3087, opened 2023-08-31
Move `interfile: true` under `options`

0xDC0DE:


enncoded (commented):
Should we also remove `interfile: true` from metadata? Or are we keeping that for compatibility with old versions?

0xDC0DE:
I figured that didn't hurt to keep, and I couldn't do both with one autofix rule 🤷 . But I can always write a second rule to delete the one under metadata if you feel that is the correct thing to do.

IagoAbal:
@colleend what is the `pro-scan-perf-test` exactly running? Just adding the `options / interfile: true` shouldn't make anything slower.

enncoded (approved):

PR #3092, opened 2023-09-05
Add rules helpful for auditing ocaml code bases

EricSesterhennX41:
These rules help to identify various coding patterns that might lead to security issues. They do not highlight places that are always vulnerable, but which should be audited more closely.


ocaml/lang/security/hashtable-dos.yaml
```diff
@@ -0,0 +1,16 @@
+rules:
+  - id: ocamllint-hashtable-dos
+    pattern: Hashtbl.create $Y
+    message: Creating a Hashtbl without the optional random number parameter makes it prone to DoS attacks when attackers are able to fill the table with malicious content. Hashtbl.randomize or the R flag in the OCAMLRUNPARAM are other ways to randomize it.
```
0xDC0DE, line 6: Based on this description, there could be some `pattern-not` or `pattern-not-inside` clauses added to reduce false positives for this rule.

ocaml/lang/security/digest.yaml
```diff
@@ -0,0 +1,23 @@
...
+      - pattern: Digest.file
+    message: Digest uses MD5 and should not be used for security purposes.
+    languages: [ocaml]
+    severity: WARNING
+    metadata:
+      category: security
+      technology:
+        - ocaml
+      cwe: "CWE-328: Use of Weak Hash (4.12)"
+      confidence: MEDIUM
```
0xDC0DE: I don't think confidence Medium is appropriate here, since this rule does not verify if it is used for security purposes. I'm not sure we can say with medium confidence that findings by this rule are actual security problems.

ocaml/lang/security/digest.yaml
```diff
@@ -0,0 +1,23 @@
+rules:
+  - id: ocamllint-digest
+    pattern-either: 
+      - pattern: Digest.string
+      - pattern: Digest.bytes
+      - pattern: Digest.substring
+      - pattern: Digest.subbytes
+      - pattern: Digest.channel
+      - pattern: Digest.file
+    message: Digest uses MD5 and should not be used for security purposes.
```
0xDC0DE: It would be nice to extend the message with remediation guidance, and offer an appropriate alternative

ocaml/lang/security/exec.yaml
```diff
@@ -0,0 +1,21 @@
+rules:
+  - id: ocamllint-exec
+    pattern-either:
+      - pattern: Unix.execve
+      - pattern: Unix.execvp
+      - pattern: Unix.execvpe
+      - pattern: Unix.system
+      - pattern: Sys.command
```
0xDC0DE: I would definitely add a `pattern-not` clause to this rule that excludes explicitly hardcoded strings.

ocaml/lang/security/exec.ml
```diff
@@ -0,0 +1,11 @@
...
+(* ruleid:ocamllint-exec *)
+let a = Unix.execve "ls"
+(* ruleid:ocamllint-exec *)
+let b = Unix.execvp "ls"
+(* ruleid:ocamllint-exec *)
+let c = Unix.execvpe "ls"
+(* ruleid:ocamllint-exec *)
+let d = Unix.system "ls"
+(* ruleid:ocamllint-exec *)
+let e = Sys.command "ls"
```
0xDC0DE: I don't think any of these test cases are actual examples of possible command injection since they all have hardcoded commands and no interaction at all with potential user input.

ocaml/lang/security/filenameconcat.yaml
```diff
@@ -0,0 +1,16 @@
+rules:
+  - id: ocamllint-filenameconcat
+    pattern: Filename.concat
+    message: When attacker supplied data is passed to Filename.concat directory traversal attacks might be possible.
```
0xDC0DE, line 4: I would at least add a `pattern-not` to exclude hardcoded strings.

ocaml/lang/security/marshal.yaml
```diff
@@ -0,0 +1,21 @@
...
+    metadata:
+      category: security
+      technology:
+        - ocaml
+      cwe: "CWE-502: Deserialization of Untrusted Data"
+      confidence: HIGH
+      likelihood: MEDIUM
+      impact: HIGH
+      subcategory:
+        - vuln
```
0xDC0DE: I don't think in its current state, this is a vuln finding rule with confidence high. For that we would need to use taint mode and add a list of sources of potential user input. Currently this rule only verifies whether data marshalled. Any checks for whether that data is 'untrusted' are missing.

Alternatively, you could add a pattern-not to exclude hardcoded  strings, and change the confidence to medium or low.

ocaml/lang/security/tempfile.yaml
```diff
@@ -0,0 +1,16 @@
...
+    metadata:
+      category: security
+      technology:
+        - ocaml
+      cwe: "CWE-367: Time-of-check Time-of-use (TOCTOU) Race Condition"
+      confidence: LOW
+      likelihood: MEDIUM
+      impact: MEDIUM
+      subcategory:
+        - vuln
```
0xDC0DE: I think this is an audit type rule.

ocaml/lang/security/unsafe.yaml
```diff
@@ -0,0 +1,46 @@
...
+    message: Unsafe functions do not perform boundary checks or have other side effects, use with care.
+    languages: [ocaml]
+    severity: WARNING
+    metadata:
+      category: security
+      technology:
+        - ocaml
+      cwe: "CWE-242: Use of Inherently Dangerous Function (4.12)"
+      confidence: HIGH
+      likelihood: MEDIUM
```
0xDC0DE: I would reduce the confidence of this rule. While these are unsafe functions, there are likely also findings where they are used safely. I would not say with high confidence that anything marked by this rule is an actual vulnerability.

0xDC0DE (changes requested):
Overall these look like good audit type rules! I have made some suggestions to improve them or lower the confidence level where appropriate. 

I would also like to see some negative test cases (so `ok` test cases on top of `ruleid` test cases).

ocaml/lang/security/digest.yaml
EricSesterhennX41:
```suggestion
    message: Digest uses MD5 and should not be used for security purposes. Consider using SHA256 instead.
```

ocaml/lang/security/digest.yaml
EricSesterhennX41:
```suggestion
      confidence: LOW
```

ocaml/lang/security/exec.ml
EricSesterhennX41:
```suggestion
#load "unix.cma";;
let p = String.concat "ls " [" "; Sys.argv.(1)]
(* ruleid:ocamllint-exec *)
let a = Unix.execve p
(* ruleid:ocamllint-exec *)
let b = Unix.execvp p
(* ruleid:ocamllint-exec *)
let c = Unix.execvpe p
(* ruleid:ocamllint-exec *)
let d = Unix.system p
(* ruleid:ocamllint-exec *)
let e = Sys.command p
```

ocaml/lang/security/hashtable-dos.yaml
```diff
@@ -0,0 +1,16 @@
+rules:
+  - id: ocamllint-hashtable-dos
+    pattern: Hashtbl.create $Y
```
EricSesterhennX41:
```suggestion
    patterns:
    - pattern: Hashtbl.create $Y
    - pattern-not: Hashtbl.create $Y ~random:true
```

ocaml/lang/security/marshal.yaml
```diff
@@ -0,0 +1,21 @@
...
+    # https://eternal.red/2021/secure-ocaml-sandbox/
+    message: Marshaling is currently not type-safe and can lead to insecure behaviour when untrusted data is marshalled. Marshalling can lead to out-of-bound reads as well.
+    languages: [ocaml]
+    severity: WARNING
+    metadata:
+      category: security
+      technology:
+        - ocaml
+      cwe: "CWE-502: Deserialization of Untrusted Data"
+      confidence: HIGH
```
EricSesterhennX41:
```suggestion
      confidence: LOW
```

ocaml/lang/security/tempfile.yaml
EricSesterhennX41:
```suggestion
        - audit
```

ocaml/lang/security/unsafe.yaml
```diff
@@ -0,0 +1,46 @@
...
+      - pattern: $X.unsafe_write 
+    message: Unsafe functions do not perform boundary checks or have other side effects, use with care.
+    languages: [ocaml]
+    severity: WARNING
+    metadata:
+      category: security
+      technology:
+        - ocaml
+      cwe: "CWE-242: Use of Inherently Dangerous Function (4.12)"
+      confidence: HIGH
```
EricSesterhennX41:
```suggestion
      confidence: MEDIUM
```

ocaml/lang/security/marshal.yaml
EricSesterhennX41: I tried to address your inputs where I was able to. Not sure about how to exclude the hardcoded strings properly. Do you have any pointers on that? I am travelling during the next two weeks and might not be able to respond.

0xDC0DE: Yes ofcourse! I am not too familiar with OCaml, and it seems like your testcode is missing a few tests for the different patterns you have defined. But I imagine a pattern like this will do the trick:

```
patterns:
  - pattern: Marshal.from_string $STR $INT
  - pattern-not: Marshal.from_string "..." $INT
```

This will allow you to look for calls to the from_string and exclude the ones where the string is hardcoded.

ocaml/lang/security/exec.yaml
EricSesterhennX41:
```suggestion
    patterns:
      - pattern-either:
        - pattern: Unix.execve $STR
        - pattern: Unix.execvp $STR
        - pattern: Unix.execvpe $STR
        - pattern: Unix.system $STR
        - pattern: Sys.command $STR
      - pattern-not: Unix.execve "..."
      - pattern-not: Unix.execvp "..."
      - pattern-not: Unix.execvpe "..."
      - pattern-not: Unix.system "..."
      - pattern-not: Sys.command "..."
```

ocaml/lang/security/filenameconcat.yaml
EricSesterhennX41:
```suggestion
    patterns:
      - pattern-either:
        - pattern: Filename.concat $STR $STR
        - pattern: Filename.concat $STR "..."
        - pattern: Filename.concat "..." $STR
      - pattern-not: Filename.concat "..." "..."
```

EricSesterhennX41:
```suggestion
    patterns:
      - pattern-either:
        - pattern: Filename.concat $STR $STR
        - pattern: Filename.concat $STR "..."
        - pattern: Filename.concat "..." $STR
      - pattern-not: Filename.concat "..." "..."
    message: When attacker supplied data is passed to Filename.concat directory traversal attacks might be possible.
```

ocaml/lang/security/marshal.yaml
EricSesterhennX41: I think for this one the pattern-not does not really make sense, since deserializing static strings is not really a useful thing to do.

ocaml/lang/security/filenameconcat.yaml
0xDC0DE:
```
    patterns:
      - pattern: Filename.concat $STR $STR
      - pattern-not: Filename.concat "..." "..."
```
I think this should be sufficient

0xDC0DE (approved):
Looks good to me if the CI checks pass.

aryx:
Quick warning @EricSesterhennX41, because those rules have LOW confidence and are a bit noisy, we will remove them
from the p/ocaml ruleset. They are still in semgrep-rules and accessible from the registry, just not anymore in the default ruleset associated with p/ocaml.

EricSesterhennX41:
Thanks for the heads up! I was mostly using them to support manual 
audits, not for constant monitoring.

On 12/13/23 11:06, Yoann Padioleau wrote:
> Quick warning @EricSesterhennX41, because those rules have LOW confidence and are a bit noisy, we will remove them
> from the p/ocaml ruleset. They are still in semgrep-rules and accessible from the registry, just not anymore in the default ruleset associated with p/ocaml.
> 
> 

-- 
Eric Sesterhenn (Principal Security Consultant)
X41 D-SEC GmbH, Krefelder Str. 123, 52070 Aachen
T: +49 241 9809418-0, Fax: -9
Unternehmenssitz: Aachen, Amtsgericht Aachen: HRB19989
Geschäftsführer: Markus Vervier

PR #3097, opened 2023-09-06
Remove FPs from secrets rule

LewisArdern:
This rule will eventually be phased out, but this fixes a FP reported from a customer


p4p3r (approved):
LGTM

PR #3123, opened 2023-09-21
Java runtime rule tests bug fix

0xDC0DE:


kurt-r2c (approved):
per discussion in slack, this is fine - @0xDC0DE to work on tests for the alternative command injection issue here.

PR #3125, opened 2023-09-22
Used typed mvar for c# javascript deserializer

p4p3r:
Use typed metavariable instead of simple hardcoded pattern.


neuroo:
Closes RULES-5655

neuroo (approved):
LGTM

PR #3131, opened 2023-09-27
Align test target extension with the 'languages' listed in the rule

mjambon:
This facilitates tests (in semgrep-proprietary) that infer the desired analyzer from the file extension. With a `.py` extension, the semgrep-pro test suite assumes the rule can only analyze the file as Python and will ignore it if the analyzer specified in the rule is `generic` or `regex`.

In one instance, the rule listed only JavaScript in a rule that came with TypeScript (TSX) sample target. I don't know if there are others. Rules that work for both JavaScript and TypeScript must list both languages in the `languages` section of the rule.


p4p3r:
@mjambon renaming the files break `semgrep --test` logic though, as I believe it uses the rule/test file names to pair them up (see failure in the CI checks: https://github.com/returntocorp/semgrep-rules/actions/runs/6321073401/job/17164522155?pr=3131)

mjambon:
@p4p3r oops, we can't add these `*txt` extensions because of the include/excludes filters. I have to fix the issue in semgrep-pro's tests instead. Sorry about that. I'll let you know when I have a PR ready.

mjambon:
One of these changes is still valid. I made a separate PR for it: https://github.com/returntocorp/semgrep-rules/pull/3142

PR #3135, opened 2023-09-29
Fix slow pattern and regex

p4p3r:


LewisArdern:
I don't think the regex is now accurate now that we 'look inwards' may want to check it

LewisArdern (approved):
lgtm, the rule is confusing tho a wildcard CORS generally means * in the cors policy, what this means is that the domain can be something else than intended e.g. 
```
https://quxabizabazafoobar.com
```
will match because of the regex since its not escaping the \.

p4p3r:
Time on `https://github.com/returntocorp/advisory-frontend/blob/main/app/api/inngest/popular-packages-2.ts`:
`0.97s user 0.13s system 94% cpu 1.156s total`

Time on `https://github.com/returntocorp/advisory-frontend/blob/main/app/api/inngest/popular-packages.ts`:
`0.76s user 0.13s system 93% cpu 0.950s total`

PR #3139, opened 2023-10-02
Fix perf of ruby.lang.security

IagoAbal:
In principle, for these rules, this:

```
...
def $METHOD
  ...
end
...
```

should be equivalent to just `...`, but the latter is much faster.

(Yet I'm not sure why the current form is so slow...)

test plan:
% semgrep -c hardcoded-secret-rsa-passphrase.yaml path/to/gitlabhq/app/models/ci/build.rb
Before the change, on a MacBook Pro (Apple M1), this took ~3s, after the change, ~0.1s.
% semgrep -c insufficient-rsa-key-size.yaml path/to/gitlabhq/app/models/ci/build.rb
Before the change, on a MacBook Pro (Apple M1), this took ~20s, after the change, ~0.1s.


p4p3r:
I suppose that, if Semgrep understood constant to class and instance fields, we could reduce the rule to just the following?
```yaml
    patterns:
      - pattern-either:
          - pattern: OpenSSL::PKey::RSA.new(..., '$SECRET')
          - pattern: OpenSSL::PKey::RSA.new(...).to_pem(..., '$SECRET')
          - pattern: OpenSSL::PKey::RSA.new(...).export(..., '$SECRET')
      - focus-metavariable: $SECRET
```

IagoAbal:
> I suppose that, if Semgrep understood constant to class and instance fields, we could reduce the rule to just the following?

It does understand it, but only when it is "guaranteed". In the target example I think we use global variables like  `$secret` and `$bad2` (or similar). So even if they're set to tainted values somewhere, nothing guarantees you that they are not sanitized somewhere else. We could have an option to be more aggressive with this and assume that anything that could happen will happen.

p4p3r (approved):
LGTM

PR #3140, opened 2023-10-02
Update `server-dangerous-object-deserialization` rule

nmote:
I'm making a change to the Semgrep Pro Engine which will correctly model
the fact that `java.lang.Object` is at the top of any inheritance
hierarchy in Java. As such, the pattern `Object $X` will match any
object type, including the `String` types used in the tests here.

In general, we do want the Pro Engine, when presented with a pattern
`(Foo $X)`, to also match any subtypes of `Foo`.

Based on a discussion with Pieter, this should actually match any object
type except for `String`s and boxed types. As such, I have updated this
rule and the test cases accordingly. Now, it functions the same both on
the Pro Engine and OSS, and is more accurate than before on both.

### DIFF
--- a/java/rmi/security/server-dangerous-object-deserialization.java
+++ b/java/rmi/security/server-dangerous-object-deserialization.java
@@ -11,15 +11,29 @@ import java.rmi.RemoteException;
 // ruleid:server-dangerous-object-deserialization
 public interface IBSidesService extends Remote {
    boolean registerTicket(String ticketID) throws RemoteException;
-   void vistTalk(String talkname) throws RemoteException;
+   void vistTalk(String talkID) throws RemoteException;
    void poke(Object attende) throws RemoteException;
+}
+
+// ruleid:server-dangerous-object-deserialization
+public interface IBSidesService extends Remote {
+   boolean registerTicket(String ticketID) throws RemoteException;
+   void vistTalk(String talkID) throws RemoteException;
+   void poke(StringBuilder attende) throws RemoteException;
 }
 
 // ok:server-dangerous-object-deserialization
 public interface IBSidesServiceOK extends Remote {
    boolean registerTicket(String ticketID) throws RemoteException;
-   void vistTalk(String talkname) throws RemoteException;
+   void vistTalk(String talkID) throws RemoteException;
    void poke(int attende) throws RemoteException;
+}
+
+// ok:server-dangerous-object-deserialization
+public interface IBSidesServiceOK extends Remote {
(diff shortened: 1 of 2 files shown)


p4p3r:
The original test case is correct though. The vulnerability lies in the combination of:
- writing an interface that extends `java.rmi.Remote`
- adding to that interface methods that takes arguments of type `java.lang.Object`

The test should stay unchanged (apart from test annotations changing line) and the rule pattern be rewritten to be:
```yaml
    patterns:
      - pattern: |
          interface $INTERFACE extends java.rmi.Remote {
            $RETURNTYPE $METHOD($TYPE $PARAM) throws RemoteException;
          }
      - metavariable-regex:
          metavariable: $TYPE
          regex: (Object|java.lang.Object)
      - focus-metavariable: $TYPE
```

0xDC0DE:
My understanding is that usually in these type of things deserialisation happens before it can be checked what the actual object type is. This is definitely the case for the ObjectInputStream in Java, based on OWASP: https://cheatsheetseries.owasp.org/cheatsheets/Deserialization_Cheat_Sheet.html#java

Java RMI uses the default ObjectInputStream as far as I can tell:
https://medium.com/geekculture/detecting-dangerous-rmi-objects-with-codeql-33e03686921f

With that reasoning, remotely invoking a methodcall that takes a String object but invoking it with some other malicious data, this data will still be deserialised and hence this can be exploited.

 I would add these references to the current rule, because the current references are pretty useless! One throws a 404.

0xDC0DE:
final verdict: It looks like Oct 30, 2019 this is updated to also distinguish the String type and parse it with the readString method instead of the readObject method and so it is not vulnerable to use Strings or primitive types. Anything else is vulnerable.
https://github.com/openjdk/jdk/commit/796f3ba8bea5276165cd9ed8ba2ae6e415a9d57b#diff-899c2f41933b5dc20511bfc450a3[…]36686bf39c8b04e46457347eR324-R325

The following sources helped me track down this info and should probably be added to the rule:
```
- https://book.hacktricks.xyz/network-services-pentesting/1099-pentesting-java-rmi
- https://youtu.be/t_aw1mDNhzI
- https://github.com/qtc-de/remote-method-guesser
- https://github.com/openjdk/jdk/blob/master/src/java.rmi/share/classes/sun/rmi/server/UnicastRef.java#L303C4-L331
```

nmote:
@0xDC0DE and @p4p3r, I have updated this based on review feedback. Could you take another look?

p4p3r (approved):
LGTM

PR #3142, opened 2023-10-03
Add TypeScript to the supported languages for a rule that listed only JS

mjambon:
supersedes https://github.com/returntocorp/semgrep-rules/pull/3131


inkz (approved):
makes sense

PR #3151, opened 2023-10-06
Rust Semgrep Rules

Matthewbialek:
Hello Semgrep team
We would like to contribute the rules we are using internally for Rust! We would love to see the Rust support for the tool to grow.  These rules do include some todorule id because I believe the pattern matching is having some issues and this could be illustrated. I understand the state of the language support of course also! 

I will be on vacation for the next 3 weeks! But i will return and handle all the feedback We look forward to a growing relationship with the semgrep community :) 

This pull request was a group effort with these great developers:
Rohit Joshi
Chris Jacoby 
[Josher](https://github.com/thejosher)


0xDC0DE:
Looks like we are still missing a CLA and some failing lints for our rules. If you have any questions or require any help, let us know! Here is a link to the docs about this as well:

https://semgrep.dev/docs/contributing/contributing-to-semgrep-rules-repository

0xDC0DE:
I'll also do a review of the actual rules tomorrow!

rust/lang/security/audit/Insufficient-protected-credentials.yml
```diff
@@ -0,0 +1,20 @@
+rules:
+  - id: insufficient-protected-credentials
+    message: |
+      This is a Insufficiently Protected Credentials weakness: https://cwe.mitre.org/data/definitions/522.html
+      Consider using an appropriate security mechanism to protect the credentials (e.g. keeping secrets in environment variables)
```
0xDC0DE, line 5:
```suggestion
    message: >-
      This is an Insufficiently Protected Credentials weakness: https://cwe.mitre.org/data/definitions/522.html
      Consider using an appropriate security mechanism to protect the credentials (e.g. keeping secrets in environment variables).
```

rust/lang/security/audit/Insufficient-protected-credentials.yml
```diff
@@ -0,0 +1,20 @@
...
+    metadata:
+      cwe: 'CWE-522: Insufficiently Protected Credentials'
+      owasp: 'A2: Broken Authentication'
+      source-rule-url: https://r2c.dev/blog/2020/hardcoded-secrets-unverified-tokens-and-other-common-jwt-mistakes/
+    pattern-either:
+      - patterns:
```
0xDC0DE, line 11: some unnecessary operators here

rust/lang/security/audit/Insufficient-protected-credentials.yml
0xDC0DE: Please use the `>-` operator for messages. This together with some other guidelines regarding metadata are available in the docs.

rust/lang/security/audit/command-injection-process-builder.rs
```diff
@@ -0,0 +1,5 @@
+//ruleid: dangerous-spawn-process
```
0xDC0DE, line 1: Could you also add True Negative tests, so tests indicated with `//ok: <ruleid>`

rust/lang/security/audit/command-injection-process-builder.yaml
```diff
@@ -0,0 +1,26 @@
...
+      asvs:
+        section: 'V5: Validation, Sanitization and Encoding Verification Requirements'
+        control_id: 5.3.8 OS Command Injection
+        control_url: https://github.com/OWASP/ASVS/blob/master/4.0/en/0x13-V5-Validation-Sanitization-Encoding.md#v53-output-encoding-and-injection-prevention-requirements
+        version: '4'
+    languages: [rust]
+    severity: WARNING
+    patterns:
+      - pattern-either:
+          - patterns:
```
0xDC0DE, line 19: unnecessary operators

rust/lang/security/audit/command-injection-process-builder.yaml
```diff
@@ -0,0 +1,26 @@
...
+              - pattern: |
+                  fn $FUNC(...) {
+                    ...
+                    Command::new(...);
+                    ...
+                  }
+
```
0xDC0DE, line 26:
```suggestion
              - pattern: |
                    Command::new(...)
              - pattern-not: |
                    Command::new("...")

```

Some suggestions:
1. By removing the function declaration around the actual vulnerable line of code we can reduce the amount of lines being reported or marked. Instead of the entire function declaration, Semgrep will now report the exact line of the `Command::new`. You will need to update the test syntax accordingly! Additionally, this would also capture calls that do not end in a semicolon. 

For example, a construction like this might be missed by your pattern (I didn't actually run it):
```
let output = if cfg!(target_os = "windows") {
    Command::new(cmd)
            .args(["/C", "echo hello"])
            .output()
            .expect("failed to execute process")
} else {
    Command::new("sh")
            .arg("-c")
            .arg("echo hello")
            .output()
            .expect("failed to execute process")
};
```

You should look up some uses of the functions and add them to the testcode with appropriate test syntax to verify this!

2. I have added a pattern-not to exclude very obvious False Positives, like a hardcoded String. This will reduce the noise for this rule.

rust/lang/security/audit/crypto/deprecated-ssl.yaml
```diff
@@ -0,0 +1,50 @@
...
+      cwe: 'CWE-327: Use of a Broken or Risky Cryptographic Algorithm'
+      owasp: 'A9: Using Components with Known Vulnerabilities'
+      references:
+        - https://golang.org/doc/go1.14#crypto/tls
+        - https://www.us-cert.gov/ncas/alerts/TA14-290A
+    languages: [rust]
+    severity: ERROR
+    fix-regex:
+      regex: SslVersion::SSL3
+      replacement: SslVersion::TLS1_3
```
0xDC0DE, line 16: If there are fixes, it is possible to add fix tests.

Create a new file called `deprecated-ssl.fixed.rs` with the expected outcome of running the autofix on the `deprecated-ssl.rs` file.

More info here: https://semgrep.dev/docs/writing-rules/testing-rules/#testing-autofix

`fix-regex` works but using regular `fix` will be more powerful, especially when we develop AST-based autofix for Rust. (https://semgrep.dev/blog/2022/autofixing-code-with-semgrep/)

If you use `fix` your rules will likely be more future-proof. Your rule would like something like this:

```
patterns:
  - pattern: |
        $VAR.set_min_proto_version(Some($TLS))
  - metavariable-regex:
        metavariable: $TLS
        regex: SslVersion::(SSL3|TLSv1|TLSv1_1)
  - focus-metavariable: $TLS
fix: |
  SslVersion::TLS1_3
```

This would allow you to combine all three rules into just one rule! You can even use the metavariable in the message if you want to still keep three separate messages:

```
message: >-
    $TLS is insecure ...
```

rust/lang/security/audit/crypto/math-random.rs
```diff
@@ -0,0 +1,15 @@
...
+fn thread_rng() {
+    //ruleid: math-random-used
+    let x = rand::thread_rng();
+}
+
+fn thread_rng_1() {
+    //todoruleid: math-random-used
+    use rand::thread_rng;
+    //todoruleid: math-random-used
+    let _ = thread_rng();
```
0xDC0DE, line 14: Semgrep does some symbolic propagation and should be aware enough for this case to be captured with your current version of the rule:

https://semgrep.dev/playground/s/N0oy

rust/lang/security/audit/crypto/math-random.yaml
```diff
@@ -0,0 +1,30 @@
...
+
+#    - pattern-not-inside: |
+#        use rand::CryptoRng;
+#        ...
+#    - pattern-not-inside: |
+#        use rand_core::StdRng;
+#        ...
+#    - pattern-not-inside: |
+#        use rand_core::OsRng;
+#        ...
```
0xDC0DE, line 30: The problem here is that you have nested them under the `pattern-either`, this operator works like an `OR` operator. 

Your `pattern-inside` is written correctly and works as expected.
https://semgrep.dev/playground/s/kJPP

Please add testcases for these `pattern-not-inside` cases to your testcode.

0xDC0DE (changes requested):
I have reviewed some of the rules, but not yet all! I will come back to this. 
It seems like you are missing some metadata for these to be accepted as well:

Here is a link to the docs: https://semgrep.dev/docs/contributing/contributing-to-semgrep-rules-repository/#semgrep-registry-rule-requirements

rust/lang/security/audit/crypto/reversible-hash.rs
```diff
@@ -0,0 +1,33 @@
...
+    let keypair = PKey::from_rsa(Rsa::generate(2048).unwrap()).unwrap();
+    let signer = Signer::new(msg_digest, &keypair).unwrap();
+    let signature = signer.sign_to_vec().unwrap();
+
+    println!("signature = {:?}", signature);
+}
+
+// Use of unsecure sha1 as MessageDigest
+fn reversible_hash_2() {
+    //ruleid: reversible-hash
```
0xDC0DE, line 20: Please add negative test cases for approved hash algortihms not flagged by this rule.

rust/lang/security/audit/crypto/reversible-hash.yaml
```diff
@@ -0,0 +1,21 @@
...
+  - id: reversible-hash
+    message: |
+      The product uses a hashing algorithm that produces a hash value that can be used
+      to determine the original input, or to find an input that can produce the same hash,
+      more efficiently than brute force techniques. This weakness is especially dangerous
+      when the hash is used in security algorithms that require the one-way property to
+      hold. For example, if an authentication system takes an incoming password and
+      generates a hash, then compares the hash to another hash that it has stored in its
+      authentication database, then the ability to create a collision could allow an attacker
+      to provide an alternate password that produces the same target hash, bypassing authentication.
```
0xDC0DE, line 11: What would be a recommended solution?

rust/lang/security/audit/crypto/use-of-des.yaml
```diff
@@ -0,0 +1,15 @@
...
+    languages: [rust]
+    severity: WARNING
+    metadata:
+      owasp: 'A9: Using Components with Known Vulnerabilities'
+      cwe: 'CWE-327: Use of a Broken or Risky Cryptographic Algorithm'
+      references:
+        - https://docs.rs/des/latest/des/
+    pattern-either:
+      - pattern: use des;
+      - pattern: use des::Des;
```
0xDC0DE, line 15:
```suggestion
    pattern-either:
      - pattern: use des;
      - pattern: use des::Des;
    fix: |
      use aes::Aes128; 
```

Would probably be easy to add a fix here. Can you add a fix test as well?

rust/lang/security/audit/crypto/tls-insecure-cipher.yaml
```diff
@@ -0,0 +1,14 @@
...
+      Detected an insecure CipherSuite via the 'tls' module. This suite is considered weak.
+      Use the function 'acceptor.set_ciphersuites()' to get a list of good cipher suites.
+    metadata:
+      cwe: 'CWE-327: Use of a Broken or Risky Cryptographic Algorithm'
+      owasp: 'A9: Using Components with Known Vulnerabilities'
+    languages: [rust]
+    severity: WARNING
+    pattern-either:
+      - pattern: |
+          $VAR.set_ciphersuites("=~/(TLS_RSA_WITH_RC4_128_SHA|TLS_RSA_WITH_AES_128_CBC_SHA256|TLS_RSA_WITH_AES_128_CBC_SHA256|TLS_ECDHE_ECDSA_WITH_RC4_128_SHA|TLS_ECDHE_RSA_WITH_AES_128_CBC_SHA256|TLS_ECDHE_ECDSA_WITH_AES_128_CBC_SHA256|TLS_AES_256_GCM_SHA384)/")
```
0xDC0DE, line 13:
```suggestion
    patterns:
      - pattern: |
        $VAR.set_ciphersuites($CIPHER)
      - metavariable-regex:
          metavariable: $CIPHER
          regex: (TLS_RSA_WITH_RC4_128_SHA|TLS_RSA_WITH_AES_128_CBC_SHA256|TLS_RSA_WITH_AES_128_CBC_SHA256|TLS_ECDHE_ECDSA_WITH_RC4_128_SHA|TLS_ECDHE_RSA_WITH_AES_128_CBC_SHA256|TLS_ECDHE_ECDSA_WITH_AES_128_CBC_SHA256|TLS_AES_256_GCM_SHA384)
    fix: |
        TLS_AES_128_GCM_SHA256
```

Feels like a rule that could easily have a fix. Mind adding one, as well as a fixtest? I've made a suggestion that might help you in the right direction.

rust/lang/security/audit/crypto/use-of-md2.yaml
```diff
@@ -0,0 +1,33 @@
+rules:
+  - id: use-of-md2
+    languages: [rust]
+    message: avoid cryptographically insecure hashing function(s)
```
0xDC0DE, line 4: Please expand the message with "why" and "what to do instead".

rust/lang/security/audit/crypto/use-of-md4.yaml
```diff
@@ -0,0 +1,33 @@
+rules:
+  - id: use-of-md4
+    languages: [rust]
+    message: avoid cryptographically insecure hashing function(s)
```
0xDC0DE, line 4: Please expand the message with "why" and "what to do instead"

rust/lang/security/audit/crypto/use-of-md5.yaml
```diff
@@ -0,0 +1,45 @@
+rules:
+  - id: use-of-md5
+    message: |
```
0xDC0DE, line 3:
```suggestion
    message: >-
```

rust/lang/security/audit/crypto/use-of-weak-rsa-key.yaml
```diff
@@ -0,0 +1,21 @@
...
+      - pattern-either:
+          - pattern: |
+              Rsa::generate($BITS)
+          - pattern: |
+              rsa::Rsa::generate($BITS)
+          - pattern: |
+              openssl::rsa::Rsa::generate($BITS)
+      - metavariable-comparison:
+          metavariable: $BITS
+          comparison: $BITS < 2048
```
0xDC0DE, line 21:
```suggestion
    patterns:
      - pattern-either:
          - pattern: |
              Rsa::generate($BITS)
          - pattern: |
              rsa::Rsa::generate($BITS)
          - pattern: |
              openssl::rsa::Rsa::generate($BITS)
      - metavariable-comparison:
          metavariable: $BITS
          comparison: $BITS < 2048
      - focus-metavariable: $BITS
    fix: |
      2048
```

rust/lang/security/audit/dangerous-exec-command.yaml
```diff
@@ -0,0 +1,32 @@
+rules:
+  - id: dangerous-exec-command
+    patterns:
+      - pattern-either:
+          - patterns:
+              - pattern-either:
+                  - pattern: |
+                      Command::new($CMD)
```
0xDC0DE, line 8:
```suggestion
          - pattern: |
                      Command::new($CMD)
```
A few unnecesary operators

rust/lang/security/audit/dangerous-exec-command.yaml
```diff
@@ -0,0 +1,32 @@
...
+          - patterns:
+              - pattern-either:
+                  - pattern: |
+                      Command::new("=~/(sh|bash|ksh|csh|tcsh|zsh|ls)/").arg("-c").arg($CMD)
+                  - pattern: |
+                      Command::new("=~/(sh|bash|ksh|csh|tcsh|zsh)/").arg("-c").arg($CMD).output()
+                  - pattern: |
+                      Command::new("=~/(sh|bash|ksh|csh|tcsh|zsh)/").arg("-c").arg($CMD).output().expect(...)
```
0xDC0DE, line 16:
```suggestion
          - pattern-either:
              - pattern: |
                  Command::new("=~/(sh|bash|ksh|csh|tcsh|zsh|ls)/").arg("-c").arg($CMD)
              - pattern: |
                  Command::new("=~/(sh|bash|ksh|csh|tcsh|zsh)/").arg("-c").arg($CMD).output()
              - pattern: |
                  Command::new("=~/(sh|bash|ksh|csh|tcsh|zsh)/").arg("-c").arg($CMD).output().expect(...)
```
unnecessary operators

rust/lang/security/audit/dangerous-exec-command.yaml
```diff
@@ -0,0 +1,32 @@
...
+      - pattern-inside: |
+          use std::process::Command;
+          ...
+      - pattern-not-inside: |
+          $CMD = "...";
+          ...
```
0xDC0DE, line 22: Can you add testcases for this? It doesn't seem to be working correctly. It seems like with this you want to exclude harcoded strings, but the testcases in the testcode all have hardcoded strings.

I want to see some `ok: dangerous-exec-command` testcases. This rule is likely to be too noisy in its current form.

rust/lang/security/audit/dangerous-system-call.yaml
```diff
@@ -0,0 +1,18 @@
...
+  - id: argument-injection
+    patterns:
+      - pattern-either:
+          - pattern: |
+              fn $FUNC(...) {
+               requests::$METHOD(...);
+               ...
+               Command::new(...);
+               ...
+                       }
```
0xDC0DE, line 11: This feels like it should become a taint mode rule with proper data flow analysis. Is the intent to monitor user input (through a request method) flowing into a command?

This should give you a start:

```
mode: taint
pattern-sources: 
  - pattern: |
       requests::$METHOD(...)
pattern-sinks:
  - pattern-either: 
     - pattern: Command::new(...). ... .args(...)
     - pattern: Command::new(...)
```

Currently this rule is assuming:
- every method after `requests::` is bad. In reality am assuming some will be harmless and not actually return dangerous user input, you will want to put constraints on the $METHOD metavariable
- it is dangerous for this input to end up in `Command::new(...)` and this can be called with or without additional `args()` calls. 

It's likely to not be completed yet.

Additionally, I wonder what the difference is between this rule and `dangerous-exec-command`. It feels like these two rules are looking for the same patterns.

rust/lang/security/audit/deserialize-untrusted.yaml
```diff
@@ -0,0 +1,16 @@
+rules:
+  - id: deserialize-untrusted-data
+    message: |
+      Data that is untrusted can not be trusted to be well-formed.
+      The application deserializes untrusted data without sufficiently verifying that the resulting data will be valid.
```
0xDC0DE, line 5: I don't know the details with Rust. But in most other languages, verifying if the deserialized data is valid, is too late. Often remote code execution is possible **during** deserialization. Are you sure this is not also the case for Rust?

I don't see any resources listed in the references of this rule, so I cannot easily verify whether this is the case.

rust/lang/security/audit/ldap-query.yml
```diff
@@ -0,0 +1,18 @@
...
+  - id: ldap-injection
+    patterns:
+      - pattern-either:
+          - pattern: |
+              let $DATA = requests::$METHOD(...).unwrap();
+              ...
+              let mut $LDAP = LdapConn::new(...)?;
+              ...
+              let $RESULT = $LDAP.search(...,$DATA,...)?.success()?;
+
```
0xDC0DE, line 11: This looks like it should be a taint mode rule! https://semgrep.dev/docs/writing-rules/data-flow/taint-mode/

rust/lang/security/audit/path/path-injection-cli-arg.yaml
```diff
@@ -0,0 +1,35 @@
...
+      - pattern-either:
+          - patterns:
+              - pattern: |
+                  fn $FUNC(...) {
+                    let $VAR = $F_ENV(...);
+                    ...
+                    $X.push(...);
+                    ...
+                    $F_FS($X);
+                  }
```
0xDC0DE, line 13: Should be a taint mode rule! https://semgrep.dev/docs/writing-rules/data-flow/taint-mode/

rust/lang/security/audit/path/path-injection-env-var.yaml
```diff
@@ -0,0 +1,46 @@
...
+                    - pattern: "std::env::var_os"
+                    - pattern: "env::var_os"
+                    - pattern: "var_os"
+              - metavariable-pattern:
+                  metavariable: $F_FS
+                  pattern-either:
+                    - pattern: "std::fs::File::open"
+                    - pattern: "fs::File::open"
+                    - pattern: "File::open"
+                    - pattern: "open"
```
0xDC0DE, line 36: Taint mode !
https://semgrep.dev/docs/writing-rules/data-flow/taint-mode/

rust/lang/security/audit/sensitive-data-exposure.yml
```diff
@@ -0,0 +1,22 @@
...
+    severity: WARNING
+    pattern-either:
+      - pattern: |
+          let $VAL = requests::$METHOD(...).unwrap();
+          ...
+          headers.set($HEADER($VAL));
+      - pattern: |
+          let $VAL = requests::$METHOD(...);
+          ...
+          headers.set($HEADER($VAL));
```
0xDC0DE, line 21: Taint mode !
https://semgrep.dev/docs/writing-rules/data-flow/taint-mode/

rust/lang/security/audit/temporary-insecure-file-permissions.yaml
```diff
@@ -0,0 +1,19 @@
...
+      - pattern-either:
+          - pattern: |
+              let $TMPDIR = tempdir()?;
+               ...
+              let $FILENAME = dir.path().join(...);
+               ...
+               File::create($FILENAME);
+               ...
+    message: |
+      Opening temporary files without appropriate measures or controls can leave the file, its contents and any function that it impacts vulnerable to attack.
```
0xDC0DE, line 13: Could you add remediation guidance and testcases for the appropriate solution?

rust/lang/security/audit/xss-templates.yaml
```diff
@@ -0,0 +1,24 @@
+rules:
+  - id: cross-site-scripting-templates-detected
+    patterns:
+      - pattern-either:
+          - pattern: |
+              let $VAR = requests::$METHOD(...).unwrap();
+              ...
+              return [...,$VAR,...].join("\n");
+
```
0xDC0DE, line 9: Taint mode!

Also think more about how to send HTML back. Commonly we use direct response writers as taint sinks for XSS. I haven't used Rust beyond a few small tests, but what I came up with in the Rocket framework for example was this:

```
#[get("/xss?<param>")]
fn xss(param: String) -> Html<String> {
    Html(format!("Html {}", param))
}
```

In this case the `Html(...)` function could be a taint sink, for example.

0xDC0DE (changes requested):

rust/lang/security/audit/crypto/tls-insecure-cipher.yaml
Matthewbialek: I was reading through the docs and i couldnt find anything about fix tests? i was also looking for examples in the other rules.

Matthewbialek: I saw the higher comment now

rust/lang/security/audit/sensitive-data-exposure.yml
Matthewbialek: @0xDC0DE could you help me understand how to do this properly. It keeps failing to parse

rust/lang/security/audit/deserialize-untrusted.yaml
Matthewbialek: https://www.reddit.com/r/rust/comments/8mjait/is_serialization_considered_a_security_risk/
https://github.com/serde-rs/serde/issues/1087

Matthewbialek: https://serde.rs/

0xDC0DE:
@Matthewbialek looks like it's been a while since this PR is touched. Would you still like some help getting this merged?

Issue #3104, opened 2023-09-11
detected-twitter-oauth rule triggers on minimized CSS files


mtausig:
**Describe the bug**

The regex of the `detected-twitter-oauth.yaml` rule is quite permissive. It just looks for the first `twitter` followed by a certain amount of quoted alphanumeric characters anywhere else in the same line following.

If you have a minimized CSS file (which contains only a single line), this is very likely to produce a false positive

**To Reproduce**

The following part of a CSS file triggers the rule, but really should not:

```CSS
.twitter{foo:bar}.bla-dialog label{background-image:url(data:image/png;base64,HofkIm9g8pPKVJUWaTsw2zfUTcxeUnlLDbfcsq9qGubnM2s3lJuUe7ZLKxjhblntXMudiBk1FrV8RwAIKoAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAACdusctOBkpkmk8wpMNGNjuLWXZuWwx7nkotmkS2e5MXf1i8pPKUmu95YW9uG+L/wCpj/R/Xn+Up8pZq/qDLCnah6H5B+Q8/wApPKV/8kzHZhvm/wDrmb2LylHlLHb9QZZV7UNs3OZuZPKTylhtvWWTtw0zaryzVblG0TLuF8nldFIh1OTkEC1ptPMrwBaAAAAAAAAAAAA)}
```

**Expected behavior**

Either the rule should be written with a stricter regex, or at least CSS files should be ignored as a workaround.

**Priority**

How important is this to you?
- [ ] P0: blocking me from making progress
- [ ] P1: this will block me in the near future
- [x] P2: annoying but not blocking me

**Additional Context**

-

LewisArdern:
This was fixed, just closing out old issues

PR #3165, opened 2023-10-16
Fix twitter oauth rule triggering on minimized css files

0xDC0DE:
Fix for https://github.com/returntocorp/semgrep-rules/issues/3104


p4p3r (approved):
LGTM

PR #3166, opened 2023-10-16
Update ReDoS message

### LINKED ISSUE #3106: regular-expression-dos message
[semgrep-rules/csharp/lang/security/regular-expression-dos/regular-expression-dos.yaml at develop · returntocorp/semgrep-rules](https://github.com/returntocorp/semgrep-rules/blob/develop/csharp/lang/security/regular-expression-dos/regular-expression-dos.yaml)

> An attacker can then cause a program using a regular expression to enter these extreme situations
    and then hang for a
    very long time.

This message seem to be copied from somewhere and taken out of context. It refers to "these extreme situations" without specifying what those are, and has "then" twice. I think it does not describe the problem very well.

0xDC0DE:
Fixes: https://github.com/returntocorp/semgrep-rules/issues/3106


p4p3r (approved):
LGTM

Issue #3109, opened 2023-09-13
Python unverified-jwt-decode rule deprecated


spehill:
https://github.com/returntocorp/semgrep-rules/blob/7d92e4efaffab8d5151dbb1c707dbe9af82312d5/python/jwt/security/unverified-jwt-decode.yaml#L4

The verify parameter in `jwt.decode` was deprecated in [v1.1.0](https://pyjwt.readthedocs.io/en/latest/changelog.html#id62)
and dropped altogether in [v2.0.0](https://pyjwt.readthedocs.io/en/latest/changelog.html#dropped-deprecated-verify-param-in-jwt-decode)

A new rule should be added, or this one modified to include the new way of specifying whether signature verification is used

i.e.
```
rules:
  - id: unverified-jwt-decode
    pattern: |
      jwt.decode(..., options={..., "verify_signature": False, ...}, ...)
    message: Semgrep found a match
    fix-regex:
      regex: (options={.*["']verify_signature["']\s*:\s*)False.*}
      replacement: \1True}
    languages: [python]
    severity: ERROR
```

PR #3170, opened 2023-10-17
Updated unverified-jwt-token according to new APIs. Added fixtest

0xDC0DE:
Fix for https://github.com/returntocorp/semgrep-rules/issues/3109


p4p3r (approved):
LGTM

PR #3171, opened 2023-10-17
Add additional testcases to flask cookies.

0xDC0DE:
Testcases from this PR https://github.com/returntocorp/semgrep-rules/pull/3116

Initially they were not flagged by the rule, but this was because of the missing `make_response()` function in the original testcode.

The rule did not need changing, but I might as well commit these new testcases.


python/flask/security/audit/testcode.py
```diff
@@ -0,0 +1,23 @@
+        import flask
```
philipturnbull, line 1: It looks like this file is not needed?

philipturnbull (approved):
`python/flask/security/audit/secure-set-cookie.py` lgtm but I don't think `testcode.py` is needed?

PR #3172, opened 2023-10-17
test-is-missing-assert

Sjord:
Check for `a == b` in tests, which should probably be `assert a == b`.


0xDC0DE:
Hey @Sjord, thanks for the contribution. This rule looks good, I think since it's a correctness rule, that is also all the metadata we need. Let's see if all the CI check pass!

0xDC0DE (approved):

PR #3174, opened 2023-10-17
Update test targets for tests relying on include:

aryx:
Thanks to https://github.com/returntocorp/semgrep/pull/8993
the include: directive in the rule is now ignored in a test context,
so you can use back the same name than the rule for the test target
file

test plan:
see related PR in semgrep


aryx:
cI is failing here, but all tests are passing in https://github.com/returntocorp/semgrep/pull/8993
so we're good to merge here.

p4p3r:
Why [generic/ci/audit/changed-semgrepignore.generic](https://github.com/returntocorp/semgrep-rules/pull/3174/files#diff-e7148bacf5efbb81e9cc3c847250c17f466b931ae9843c8d09cf137e8ea342ac) has been deleted? It's a test file for a rule.

p4p3r:
Why the rule files have been moved? Moving means changing rule id in the registry, which means findings deleted and appearing as new findings for users. We can't move rules around in the filesystem.

aryx:
Answering on Slack for live discussion

p4p3r (approved):
LGTM

p4p3r (approved):
LGTM

PR #3182, opened 2023-10-19
Github-owned actions starting with  are not third party actions

0xDC0DE:


philipturnbull (approved):

yaml/github-actions/security/third-party-action-not-pinned-to-commit-sha.test.yml
```diff
@@ -54,6 +54,13 @@ jobs:
       # ruleid: third-party-action-not-pinned-to-commit-sha
       - uses: docker://alpine:3.8
 
+      # ok: third-party-action-not-pinned-to-commit-sha
+      - name: Upload SARIF file for GitHub Advanced Security Dashboard
+        uses: github/codeql-action/upload-sarif@e4262713b504983e61c7728f5452be240d9385a7
```
armchairlinguist: Should this test line be @v2 instead? Otherwise I think the hash would keep it from flagging even without the change. (But please ignore this if I'm wrong!)

0xDC0DE: No, you are totally right, that was an oversight on my part. I updated it!

jayaddison:
Hi @0xDC0DE @philipturnbull @armchairlinguist - apologies for the long-post-merge comment here.  A couple of months ago I learned that GitHub Dependabot Alerts -- note: a different feature than the Dependabot Upgrade feature that opens PRs to increment dependency versions -- does not yet support alerting on GitHub Actions actions that are referenced by commit SHA.

I do think it's often sensible to pin by commit SHA - and Dependabot will offer updates for those, so I don't think there's a problem with this pull request.  But I do want to try to raise some awareness that there are currently some downsides to pinning third-party actions to commit versions, too.

Ref: https://github.com/orgs/community/discussions/154189

PR #3184, opened 2023-10-19
Fix metadata for Apex rules

p4p3r:


0xDC0DE (approved):

p4p3r:
The output of "validate-r2c-registry-metadata" is wrong: it says a file has been changed while that file doesn't even exist! To be investigated...

p4p3r:
And now the failure in validate-r2c-registry-metadata is going by itself...

PR #3190, opened 2023-10-23
2 new django rules: 1) Requiring {% csrf_token  %} in manually-defined forms in template files 2) Recommending using form.cleaned_data[] after form.is_valid() has run successfully

justinhopkins:
Hi there,

Here are 2 rules I've found useful in my own Django projects. First time writing semgrep rules, so I'm a little concerned about how well they generalize. 

1) django-no-csrf-token. In manually-defined django template's (templates/*.html files), one should always insert {% csrf_token %} inside the form tag. {% csrf_token %} are now considered standard practice to get prevention against Cross-Site Request Forgeries.
See: https://docs.djangoproject.com/en/4.2/howto/csrf/

2) django-using-request-post-after-is-valid. Django best practices for accessing clean data is to first run $FORM.is_valid() inside a 'POST' handler, and then access the resulting fields via $FORM.cleaned_data[], as opposed to request.POST[]. $FORM.cleaned_data[] has been safely sanitized. The -pattern-inside clause I've come up with doesn't cover everything, but in my experience this is a very common django pattern for form validation.
See:  https://docs.djangoproject.com/en/4.2/ref/forms/api/#accessing-clean-data

Anyways, appreciate y'all having a look and I'm happy to make changes based on your feedback.


python/django/security/django-no-csrf-token.yaml
```diff
@@ -0,0 +1,21 @@
+rules:
+  - id: django-no-csrf-token
+    patterns:
+      - pattern: "<form...>...</form>"
+      - pattern-not-inside: "<form...>...{% csrf_token %}...</form>"
+    message: Manually-created forms in django templates should specify a csrf_token to prevent CSRF attacks
+    languages: [generic]
+    severity: WARNING
+    metadata:
```
kurt-r2c, line 9: I would include https://docs.djangoproject.com/en/4.2/howto/csrf/ from your PR description under `references` in metadata as well

python/django/security/django-using-request-post-after-is-valid.yaml
```diff
@@ -0,0 +1,25 @@
...
+      - pattern-inside: |
+          if $FORM.is_valid():
+            ...
+      - pattern-either:
+        - pattern: request.POST[...]
+        - pattern: request.POST.get(...)
+    message: Use $FORM.cleaned_data[] instead of request.POST[] after form.is_valid() has been executed to only access sanitized data
+    languages: [python]
+    severity: WARNING
+    metadata:
```
kurt-r2c, line 16: same here - add a `references` key with https://docs.djangoproject.com/en/4.2/ref/forms/api/#accessing-clean-data

kurt-r2c (changes requested):
Couple small nits, then this is good to go

python/django/security/django-using-request-post-after-is-valid.yaml
justinhopkins: References added. Good ideas; thanks!

python/django/security/django-no-csrf-token.yaml
justinhopkins: Added

philipturnbull (approved):

kurt-r2c (approved):

PR #3201, opened 2023-10-27
Update ELB security policies

allanbreyes:
The current rule produces findings for ELB security policies that are TLS 1.2 and TLS 1.3, including the AWS-recommended one (`ELBSecurityPolicy-TLS13-1-2-2021-06`), which is TLS 1.3 with TLS 1.2 compatibility.

This pull request updates the patterns to regexes that should be a little more future proof.

Reference: https://docs.aws.amazon.com/elasticloadbalancing/latest/application/create-https-listener.html#describe-ssl-policies


kurt-r2c (approved):
we like less brittleness

allanbreyes:
Looks like a perf test failed: https://github.com/returntocorp/semgrep-rules/actions/runs/6668912440/job/18230637467

Seems to be from a flake or networking issue?

```
curl: (6) Could not resolve host: argoworkflows-dev.corp.r2c.dev
```

Please advise on how to proceed. Thank you!

kurt-r2c:
@allanbreyes you were missing changes from upstream related to the decommissioning of some old infrastructure. Test should pass now.

allanbreyes:
Thanks for the review and merge!

PR #3218, opened 2023-11-09
Add reference to event-binding-payload-with-hyphen rule

philipturnbull:
This rule was added in #3208 but was missing a reference. CI didn't catch the error because the `Validate metadata` workflow was not running correctly when the PR was merged.


p4p3r (approved):
LGTM, thanks!

PR #3219, opened 2023-11-13
Fix jsonwebtoken rule for fps

LewisArdern:
Fixes https://semgrep.slack.com/archives/C0242656NQ2/p1699891553731939

the source is 'questionable' but without completely rewriting the rule this will fix the issue.


p4p3r (approved):
LGTM

PR #3230, opened 2023-11-21
Skip elixir which is now in Pro

aryx:
test plan:
see related PR in semgrep

### DIFF
--- a/.codemapignore
+++ b/.codemapignore
@@ -1,10 +1,16 @@
-# showing just the rules, for pad's codemap
+# -*- sh -*-
+# gitignore-like file for Codemap (see https://github.com/aryx/codemap)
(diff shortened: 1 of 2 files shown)


aryx:
see https://github.com/semgrep/semgrep/pull/9308

.codemapignore
```diff
@@ -0,0 +1,13 @@
+# showing just the rules, for pad's codemap
```
p4p3r: Did you actually mean to add this file?

aryx: yes, I've added similar files in other repos to visualize and generate code statistics (like the fact that
we reached recently 1 M LOC, of real code (not generated stuff).

p4p3r (approved):
LGTM

PR #3231, opened 2023-11-22
Update django-no-csrf-token rule

inkz:
Mitigate FPs for cases when GET method is used (GET is also a default method for forms)


p4p3r (approved):
LGTM

PR #3236, opened 2023-11-30
Remove incorrect fix for cookie-issecure-false.yaml

0xDC0DE:
This fix is seemingly copied from another rule. This rule checks for missing `setSecure(true)`, but the fix replaces a `setSecure(false)` with a `setSecure(true)`.

For a good autofix, this rule needs to be split into two rules:
- One for a missing `setSecure` call, that adds one with the right arguments
- One for a `setSecure` call with `false` that needs to be replaced with `true`.

Neither will need `fix-regex`, but can be fixed with a regular `fix`.

### DIFF
--- a/java/servlets/security/cookie-issecure-false.yaml
+++ b/java/servlets/security/cookie-issecure-false.yaml
@@ -10,9 +10,6 @@ rules:
   message: >-
     Default session middleware settings: `setSecure` not set to true.
     This ensures that the cookie is sent only over HTTPS to prevent cross-site scripting attacks.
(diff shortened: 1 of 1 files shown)


inkz (approved):

inkz:
@0xDC0DE do you think this update is relevant? https://github.com/semgrep/semgrep-rules/blob/develop/java/servlets/security/cookie-issecure-false.yaml#L13
or it is better to remove `fix` anyway?

0xDC0DE:
No that fix seems fine.

PR #3241, opened 2023-12-12
fix: Exclude access to request.user from view

tartieret:
The django test for access to foreign keys exclude `request.user` but fails if we do this within a view and use `self.request.user`. This PR fixes the issue by excluding this common use case in django & django rest framework.


tartieret:
@0xDC0DE sorry to ping you, but I saw that you recently approved another django-related rule and that you work at Semgrep. I am wondering if we could get this merged as it leads to false positive and would be a super quick review?

p4p3r (approved):
@tartieret, it looks good to me!

p4p3r:
I'm going to merge in `develop` and it will make its way into `release` later this week!

tartieret:
> I'm going to merge in `develop` and it will make its way into `release` later this week!

Thanks a lot!

PR #3244, opened 2023-12-13
Add new rule for flask-route-decorator

khanhldt:
Wahoo! New published rules with khanhldt.pnbj from @khanhldt.

See [semgrep.dev/s/2ZJ3Y](https://semgrep.dev/s/2ZJ3Y) for more details.

Thanks for your contribution! ❤️


khanhldt:
I signed the CLA but that message doesn't get resolved for some reason.

0xDC0DE:
It looks like Semgrep does not take the order of decorators into account currently: https://semgrep.dev/playground/s/pKLAZ

While your regex is a nice workaround, I am reporting this as a bug/feature request so that we can avoid complex regexes for this.

There are other route decorators in Flask besides the `route` decorator, e.g. `delete`, `patch`, `put`, `post`, `get`. These are basically aliases for the `route` decorator that use a different default value for the optional `methods` argument. I think the rule should probably include those as well.

I've annotated a few more things you should change for the CI checks to pass.

python/flask/correctness/flask-route-decorator.py
```diff
@@ -0,0 +1,50 @@
+from flask import Blueprint
+
+blueprint = Blueprint()
+
+
+# ok: Only 1 decorator
```
0xDC0DE: Semgrep reports this test syntax as a valid rule id. The `ok:` test syntax expects to be followed by a ruleid. If you want to indicate why this is ok, you can do that on a separate line

```suggestion
# Only 1 decorator
# ok: flask-route-decorator
```

python/flask/correctness/flask-route-decorator.py
```diff
@@ -0,0 +1,50 @@
...
+@blueprint.route("HAI")
+def f():
+    pass
+
+# ok: Correct order
```
0xDC0DE, line 12: Same here

python/flask/correctness/flask-route-decorator.py
```diff
@@ -0,0 +1,50 @@
...
+@blueprint.route("HAI")
+@another.func("HAI")
+def f():
+    pass
+
+# ok: "func" not "route"
```
0xDC0DE, line 18: Same here

python/flask/correctness/flask-route-decorator.py
```diff
@@ -0,0 +1,50 @@
...
+@another.func("HAI")
+@blueprint.func("HAI")
+def f():
+    pass
+
+# ok: Correct order with app
```
0xDC0DE, line 24: Same here

python/flask/correctness/flask-route-decorator.py
```diff
@@ -0,0 +1,50 @@
...
+@app.route("HAI")
+@another.func("HAI")
+def f():
+    pass
+
+# ok: 2 routes are independent
```
0xDC0DE, line 30: Same here

0xDC0DE (changes requested):

PR #3246, opened 2023-12-14
Add test case for PHP short-hand echo syntax

appills:
Added `<?= ?>` example for PHP to demonstrate that semgrep properly captures the short-hand echo as a sink


0xDC0DE (approved):

p4p3r:
@appills I'm seeing a bunch of test failures in CI but I'm not sure I understand why they're happening. Could you please update your branch with the latest semgrep-rules:develop, maybe that's going to be enough to fix the failures :-)

PR #3250, opened 2023-12-21
Delete contrib directory

LewisArdern:
This is not maintained by, if people want to use most of these rules they can be found via https://semgrep.dev/p/nodejsscan which is the more updated version from the original author


p4p3r (approved):
LGTM

PR #3265, opened 2024-01-05
update to rails unsafe reflection methods

normprovost:
👋 Hey team!

I believe I've identified a false positive here.

This rule was intended to match when user input was provided as an argument to the `.tap` method. Here's the relevant brakeman rule that this was related to: https://github.com/presidentbeef/brakeman/blob/main/lib/brakeman/checks/check_unsafe_reflection_methods.rb

What I found in practice was that this would match whenever object the `.tap` method was called ON as well. I ran into this with a bit of code that looked like this

```ruby
  my_id = params[:id]
  my_new_object = MyObject.new(my_id)
  my_new_object.tap { |x| x.anything_here = 'asdf' }
```

In this case, the user input is not being passed in to the `tap` method, so I'm pretty sure there's no risk of unsafe reflection? I couldn't think of a way at least 🙃

### DIFF
--- a/ruby/rails/security/brakeman/check-unsafe-reflection-methods.yaml
+++ b/ruby/rails/security/brakeman/check-unsafe-reflection-methods.yaml
@@ -22,9 +22,11 @@ rules:
       - pattern-inside: |
           $X. ... .to_proc
       - pattern-inside: |
-          $Y.method(...)
+          $Y.method(<... $X ...>)
       - pattern-inside: |
-          $Y.tap(...)
+          $Y.tap(<... $X ...>)
+      - pattern-inside: |
+          $Y.tap { |$Z| <... $X ...> }
   message: Found user-controllable input to a reflection method. This may allow a user to alter program
     behavior and potentially execute arbitrary instructions in the context of the process. Do not provide
     arbitrary user input to `tap`, `method`, or `to_proc`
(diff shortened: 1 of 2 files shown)


ruby/rails/security/brakeman/check-unsafe-reflection-methods.yaml
```diff
@@ -22,9 +22,11 @@ rules:
       - pattern-inside: |
           $X. ... .to_proc
       - pattern-inside: |
-          $Y.method(...)
+          $Y.method(<... $X ...>)
```
kurt-r2c: Deep ellipsis is considered harmful by our engine team - it's supported but the performance is poor due to the recursive behavior under the hood.

Consider a `focus-metavariable` approach like the following: https://semgrep.dev/playground/s/pKLNe

kurt-r2c (changes requested):
I think the changes themselves are good, however the nature of the *implementation* needs some fixes to avoid poor performance.

ruby/rails/security/brakeman/check-unsafe-reflection-methods.yaml
normprovost: thanks! fixed!

kurt-r2c (approved):

PR #3267, opened 2024-01-08
Add fixes to python cryptography rules

0xDC0DE:
Related to RULES-6381

### DIFF
--- a/python/cryptography/security/insecure-cipher-mode-ecb.yaml
+++ b/python/cryptography/security/insecure-cipher-mode-ecb.yaml
@@ -4,7 +4,7 @@ rules:
     ECB (Electronic Code Book) is the simplest mode of operation for block ciphers. 
     Each block of data is encrypted in the same way. 
     This means identical plaintext blocks will always result in identical ciphertext blocks, which can leave significant patterns in the output.
-    Use a different, cryptographically strong mode instead, such as CBC.
+    Use a different, cryptographically strong mode instead, such as GCM.
   metadata:
     source-rule-url: https://github.com/PyCQA/bandit/blob/d5f8fa0d89d7b11442fc6ec80ca42953974354c8/bandit/blacklists/calls.py#L101
     cwe:
@@ -30,4 +30,4 @@ rules:
   languages:
   - python
   pattern: cryptography.hazmat.primitives.ciphers.modes.ECB($IV)
-  fix: cryptography.hazmat.primitives.ciphers.modes.CBC($IV)
+  fix: cryptography.hazmat.primitives.ciphers.modes.GCM($IV)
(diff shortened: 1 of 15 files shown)


python/cryptography/security/insecure-cipher-mode-ecb.yaml
```diff
@@ -30,3 +29,5 @@ rules:
   severity: WARNING
   languages:
   - python
+  pattern: cryptography.hazmat.primitives.ciphers.modes.ECB($IV)
+  fix: cryptography.hazmat.primitives.ciphers.modes.CBC($IV)
```
neuroo: We shouldn't recommend CBC since it has its own weaknesses as well (cf. all the padding oracles attacks). GCM is the recommended way nowadays. But then, you should move to using AHEAD.

python/cryptography/security/insecure-hash-algorithms-md5.yaml
```diff
@@ -32,3 +31,10 @@ rules:
   severity: WARNING
   languages:
   - python
+  patterns:
+    - pattern: cryptography.hazmat.primitives.hashes.$MD5()
+    - metavariable-regex:
+        metavariable: $MD5
+        regex: ^(MD5)$
+    - focus-metavariable: $MD5
+  fix: SHA256
```
neuroo, line 40: Any reason why you recommend SHA256? We had SHA3 family since, but also the blake family which IIRC has a better SIMD utilization

python/cryptography/security/insufficient-dsa-key-size.yaml
```diff
@@ -8,6 +8,9 @@ rules:
   - metavariable-comparison:
       metavariable: $SIZE
       comparison: $SIZE < 2048
+  - focus-metavariable: $SIZE
+  fix: |
```
neuroo, line 12: I  understand we want a min key size, but I would think we should also get the folks over to ECDSA. Why not have the fix actually move them to it?

python/cryptography/security/insufficient-rsa-key-size.yaml
```diff
@@ -8,6 +8,9 @@ rules:
   - metavariable-comparison:
       metavariable: $SIZE
       comparison: $SIZE < 2048
+  - focus-metavariable: $SIZE
+  fix : |
+      2048
```
neuroo, line 13: Same as above.

neuroo (changes requested):

python/cryptography/security/insecure-hash-algorithms-md5.yaml
0xDC0DE: The documentation of the `cryptography` package mentions this:

> SHA-3 is the most recent NIST secure hash algorithm standard. Despite the larger number SHA-3 is not considered to be better than SHA-2. Instead, it uses a significantly different internal structure so that if an attack appears against SHA-2 it is unlikely to apply to SHA-3. SHA-3 is significantly slower than SHA-2 so at this time most users should choose SHA-2.

neuroo (approved):
lgtm!

PR #3275, opened 2024-01-12
separate open redirect from tainted-url-host

hex0punk:
rule  go.lang.security.injection.tainted-url-host.tainted-url-host its well crafted, but it catches a lot of open redirects and marks them as SSRF, often TP findings, but simple redirects none the less. I am proposing instead adding a separate rule for open redirects, which typically would have a lesser impact that an SSRF.  While I am sure there are some edge cases that Open Redirects can lead to SSRF in Golang, I have not seen a good example either in the web or my work in security (consulting or otherwise).


LewisArdern:
@hex0punk can you remove it from the previous rule also?

hex0punk:
> @hex0punk can you remove it from the previous rule also?

Done. Note `go fmt` updated the tabs so it looks like there are more changes than there were

LewisArdern:
One nitpick can you also change the metadata to match the previous rule, you changed them both to low
```
      impact: MEDIUM
      likelihood: MEDIUM
``` 

etc that make sense for the non vuln description

hex0punk:
> One nitpick can you also change the metadata to match the previous rule, you changed them both to low
> 
> ```
>       impact: MEDIUM
>       likelihood: MEDIUM
> ```
> 
> etc that make sense for the non vuln description

Yep, done.

LewisArdern:
@hex0punk can u merge in the latest develop and we can get this merged

LewisArdern (approved):

PR #3283, opened 2024-01-17
Update confidence of C rules

philipturnbull:


c/lang/security/double-free.yaml
```diff
@@ -35,7 +35,7 @@ rules:
     category: security
     technology:
     - c
-    confidence: MEDIUM
+    confidence: LOW
```
neuroo, line 38: What is the rationale to move these rules from MEDIUM to LOW confidence? FP rate?

neuroo (approved):

kurt-r2c (approved):

PR #3288, opened 2024-01-24
feat(autofix): fix fix indentation

brandonspark:
This PR fixes the indentation of an autofixed file, after the changes in this PR: https://github.com/semgrep/semgrep/actions/runs/7587358673/job/20667705690?pr=9488


kurt-r2c (approved):
LGTM  once prereqs merged & tests pass

brandonspark:
@kurt-r2c so, it turns out the fix is a bit more intensive than i thought, and i had to edit the PR. several YAML rules had to be reworked because with the new aligning behavior, its not straightforward to produce a fix where successive lines are actually appearing _before_ the first. all the current existing rules are able to still be written and maintain semantic correctness, but they will produce slightly different YAML.

aryx (approved):

PR #3296, opened 2024-01-31
Add more negative patterns to exclude cases of reading from a file for Yaml bad deserialization

navhits:
For cases where `YAML.load(File.read(...))` was used, the rule flagged these as potential issues. However, when the same was written as,

```rb
filename = File.read(...)
obj = YAML.load(filename)
```

It was allowed. Either it is a problem with how the parser works, or the rule needs an additional negative pattern.

Likewise, the same is true for cases like `YAML.load($MOD.$METHOD(File.read(...)))`. Here, the file could be passed to a template renderer like `ERB`, which, in this case, should be fine.

```rb
template = ERB.new(File.read("test.yml"))
obj = YAML::load(template.result)
```

These 2 FP conditions are covered. Also, `File.read` takes a file name as a string. It may not be explicitly needed to make the pattern expect a string.


navhits:
@0xDC0DE would you please review this? I don't have the request review option.

0xDC0DE (approved):
Looks good to me, thanks for contributing, Naveen!

navhits:
I missed updating the test file for autofix test case to pass. Added that now.

0xDC0DE (approved):

PR #3307, opened 2024-02-16
Add PHP base-convert-loses-precision

Sjord:
The function base_convert uses 64-bit numbers internally, and does not correctly convert large numbers. It is not suitable for random tokens such as those used for session tokens or CSRF tokens.

### DIFF
--- a/php/lang/security/base-convert-loses-precision.php
+++ b/php/lang/security/base-convert-loses-precision.php
@@ -3,78 +3,78 @@
 $command = $secret = $counter = $uid = $totalColors = $i = $row = $value = $numeric = $string = '1';
 $getRandomNumber = 'rand';
 
-// ok: base-convert
+// ok: base-convert-loses-precision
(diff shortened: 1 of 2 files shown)


0xDC0DE (approved):

php/lang/security/base-convert-loses-precision.php
```diff
@@ -0,0 +1,80 @@
+<?php
+
+$command = $secret = $counter = $uid = $totalColors = $i = $row = $value = $numeric = $string = '1';
+$getRandomNumber = 'rand';
+
+// ok: base-convert
```
p4p3r: @Sjord all you annotations are using the wrong rule id: they should use `base-convert-loses-precision`.
That's why the CI check is failing :)

Sjord: Thanks. I fixed it. I renamed the rule to give it a more specific name, but apparently did not replace all occurrences.

PR #3327, opened 2024-03-11
updated message for Entrypoint docker security rule

un-stable:
This PR updates the description for the containsecurity rule in Semgrep's documentation for missing-user-entrypoint rule. The previous description lacked clarity and failed to adequately convey the security implications of not specifying an ENTRYPOINT in a Dockerfile. It seems it had repeated message from previous rule

The updated description emphasizes the importance of defining an ENTRYPOINT to mitigate potential security risks, particularly the elevated privileges that a program may inadvertently run with if not properly configured. By providing more comprehensive guidance, developers using Semgrep will gain a better understanding of the security implications and best practices associated with Dockerfile configurations.

Changes:

- Updated message at dockerfile.security.missing-user-entrypoint.missing-user-entrypoint

### DIFF
--- a/dockerfile/security/missing-user-entrypoint.yaml
+++ b/dockerfile/security/missing-user-entrypoint.yaml
@@ -9,9 +9,7 @@ rules:
   fix: |
     USER non-root
     ENTRYPOINT $...VARS
-  message: By not specifying a USER, a program in the container may run as 'root'. This is a security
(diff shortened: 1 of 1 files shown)


inkz:
Hi @un-stable 
The rule that you pointing out is intended to find absence of USER before setting ENTRYPOINT. While your description is about ENTRYPOINT being missed, which is not what this rule is expect to highlight.

Closing this PR. Thank you!

PR #3329, opened 2024-03-12
bash: Fix false positives on arithmetic expressions

### LINKED ISSUE #3328: unquoted-variable-expansion-in-command: False positive on arithmetic expressions "$((...))"
The pattern here:
https://github.com/semgrep/semgrep-rules/blob/93b5b0ddd064839705c6f8583bb3fc040374b8d9/bash/lang/correctness/unquoted-expansion.yaml#L48-L52

Gives false positives on arithmetic expressions like `$((2+2))` or `$((foo++))`.

For fix, see: #3329

AfroThundr3007730:
The bash rule `unquoted-variable-expansion-in-command` triggered a false positive on arithmetic expressions.

This fixes the issue  and adds new test cases.

Fixes: #3328


inkz (changes requested):
@AfroThundr3007730  Can you please rebase this commit to `develop` branch?

PR #3335, opened 2024-03-13
Escape hostnames and add fonts.gstatic.com to missing-integrity

allanbreyes:
This change:
* Cleans up the regexes to be more precise (escaping `.` in hostnames
* Adds `fonts.gstatic.com` to the safe hosts list


inkz (approved):
LGTM! Thank you!

PR #3344, opened 2024-03-26
removing rule that detects just parameter names

6f6d6172:
Hi folks,

It's me deleting another rule. This time it's a rule that basically flags specific parameter names as a high severity finding. I get what it's going for but this rule isn't really useful, especially when the guidance is "don't use those parameters". This rule can't tell if you're using the parameters safely or not, and the guidance might cause someone who was doing something safely to do something more dangerous (e.g. using `params[:account_id]` instead of `params.permit(...)` and assigning it manually to a model).

Regarding the "spirit" of what this rule is trying to detect, I don't think it's possible without doing some sort of control flow analysis. While it's a different tool entirely, I've had success [using this Rubocop rule](https://github.com/Betterment/betterlint/#bettermentauthorizationincontroller) I wrote a couple years ago.


0xDC0DE:
Closing this as the CLA was never signed and this is a really old PR.

PR #3349, opened 2024-04-10
remove redundant rules for HTML templates

inkz:
for: https://linear.app/semgrep/issue/CODE-6898/pdefault-redundant-rules-creating-triplicate-findings-for-html

Ideally would be to merge this one after this PR is done: https://github.com/semgrep/semgrep-rule-packs/pull/428


p4p3r:
@inkz `javascript.express.security.audit.xss.ejs.var-in-href.var-in-href` is another "duplicate", and the generic rule should match also `*.ejs` files.

inkz:
@p4p3r as far as I remember it uses regex and is intended to work on a different syntax. so the generic rule is not gonna work on it.
Mustache `{{ foobar }}`
EJS `<%= foobar %>`

p4p3r (approved):
lgtm

PR #3353, opened 2024-04-15
[PHP] Improve Echoed Requests

shellz-n-stuff:
## Description
This PR aims two improve the echoed requests PHP injection rule by doing the followings:
1. Splitting out Print and Echo into seperate rules to allow for Fixes/Autofixes
2. Adding additional sanitisation functions to reduce the number of false positives associated with the rules
3. Remove a false negative associated with "empty" (more on this below)

## False Negative
The empty function by itself isn't really a sanitiser as it returns a [boolean output rather than an empty string](https://www.php.net/manual/en/function.empty.php). In theory we could add a sanitiser similar to:
`pattern: if(empty($...VARS))`

However, this is likely to be just as noisy/false negative heavy. There may be a smart way to check this with metavariable comparisons. If someone wants to try further improve this rule. 

## Notes
I couldn't find any way to use fix-regex to keep these rules combined but if would love to simplify this if you have any suggestions.

### DIFF
--- a/php/lang/security/injection/echoed-request.php
+++ b/php/lang/security/injection/echoed-request.php
@@ -1,18 +1,18 @@
 <?php
 
 // example key-value: name=%3Cscript%3Econfirm%28%29%3C%2Fscript%3E
-function dangerousPrintUsage() {
+function dangerousEchoUsage() {
     $name = $_REQUEST['name'];
     // ruleid: echoed-request
-    print("Hello : $name");
+    echo "Hello : $name";
     // ruleid: echoed-request
-    print("Hello : " . $name);
+    echo "Hello : " . $name;
 }
 
-function safePrintUsage() {
+function safeEchoUsage() {
     $name = $_REQUEST['name'];
     // ok: echoed-request
-    print("Hello : " . htmlentities($name));
+    echo "Hello : " . htmlentities($name);
 }
 
 function doSmth() {
--- a/php/lang/security/injection/echoed-request.php
+++ b/php/lang/security/injection/echoed-request.php
@@ -44,6 +44,22 @@ function doSmth5() {
     echo "Hello ".trim($_POST['name']);
 }
 
+function doSmth6() {
+   $VAR = $_GET['someval']
+    if(isset($VAR)){ 
+        // ruleid: echoed-request
+        echo $VAR; 
+    }
+}
+
+function doSmth7() {
+    $VAR = $_GET['someval']
+     if(empty($VAR)){ 
+         // ruleid: echoed-request
+         echo $VAR; 
+     }
+ }
+
 function doOK1() {
     // ok: echoed-request
     echo "Hello ".htmlentities($_POST['name'])." !";
@@ -84,4 +100,14 @@ function doOK7() {
     echo $safevar;
 }
 
+function doOK8() {
+    // ok: echoed-request
+    echo "Hello ".isset($_POST['name'])." !";
+}
+
+function doOK9() {
+    $safevar = empty($_GET['name']);
+    // ok: echoed-request
+    echo "Hello $safevar !";
+}
 
--- a/php/lang/security/injection/echoed-request.php
+++ b/php/lang/security/injection/echoed-request.php
@@ -63,11 +63,11 @@ function doOK3() {
 
 function doOK4() {
     // ok: echoed-request
-    echo "Hello ".isset($_POST['name'])." !";
+    echo "Hello ".e($_POST['name'])." !";
 }
 
 function doOK5() {
-    $safevar = empty($_GET['name']);
+    $safevar = esc_attr($_GET['name']);
     // ok: echoed-request
     echo "Hello $safevar !";
 }
--- a/php/lang/security/injection/echoed-request.php
+++ b/php/lang/security/injection/echoed-request.php
@@ -45,7 +45,7 @@ function doSmth5() {
 }
 
 function doSmth6() {
-   $VAR = $_GET['someval']
+   $VAR = $_GET['someval'];
     if(isset($VAR)){ 
         // ruleid: echoed-request
         echo $VAR; 
@@ -53,7 +53,7 @@ function doSmth6() {
 }
 
 function doSmth7() {
-    $VAR = $_GET['someval']
+    $VAR = $_GET['someval'];
      if(empty($VAR)){ 
          // ruleid: echoed-request
          echo $VAR; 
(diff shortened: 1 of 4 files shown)


shellz-n-stuff:
Arguably isset should also be removed as a sanitisation function.

kurt-r2c:
@shellz-n-stuff I suspect `isset` is a sanitizer here because the return type is a boolean and standard propagation would consider the return value tainted (which is false, as it's a boolean not the user input)

shellz-n-stuff:
> @shellz-n-stuff I suspect `isset` is a sanitizer here because the return type is a boolean and standard propagation would consider the return value tainted (which is false, as it's a boolean not the user input)

@kurt-r2c,

I'll update this rule later today to remove isset. I think a pattern like the below would be a fairly common false negative. I can't think of many cases where this would meaningfully increase false positives either

`if(isset($VAR)){ echo $VAR; }`

I'll also sort out the test failures.

shellz-n-stuff:
@kurt-r2c updated 🙏

kurt-r2c (changes requested):
@shellz-n-stuff removal of `isset` here will cause false positives as the taint must flow *through* the function to the sink to be considered sanitized - the sanitizer is not by side effect here.

without `isset` this will (incorrectly) flag `echo isset($VAR)`
https://semgrep.dev/playground/s/JDold

the correct behavior is:
https://semgrep.dev/playground/s/5rz6W

note that even with the sanitizer we are still flagging line 13 because an *unsanitized* reference of `$VAR` makes it into the sink

Similarly, `empty(...)` should not be removed.

Please revert these changes and this should be good to merge.

shellz-n-stuff:
> @shellz-n-stuff removal of `isset` here will cause false positives as the taint must flow _through_ the function to the sink to be considered sanitized - the sanitizer is not by side effect here.
> 
> without `isset` this will (incorrectly) flag `echo isset($VAR)` https://semgrep.dev/playground/s/JDold
> 
> the correct behavior is: https://semgrep.dev/playground/s/5rz6W
> 
> note that even with the sanitizer we are still flagging line 13 because an _unsanitized_ reference of `$VAR` makes it into the sink
> 
> Similarly, `empty(...)` should not be removed.
> 
> Please revert these changes and this should be good to merge.

Howdy Kurt,

Happy to make the change, but I think in terms of optimising the rule having this as a valid sanitiser is far more likely to increase false negatives than false positives. 

It strikes me as a fairly common coding convention to check a variable isn't null before echoing it out e.g.:
```php
if(isset($VAR)){ 
  echo "your name is" + $VAR; 
}
```

Whereas the pattern you described where you are echoing out a function that returns a boolean seems less common/likely. Obviously I'm happy to make the change if you disagree as I firmly believe you're in a better position to assess the noise/data quality of these rules ❤️

kurt-r2c:
@shellz-n-stuff totally understand your concern - you'll note that in the correct behavior with the `isset` sanitizer (https://semgrep.dev/playground/s/5rz6W) line 9 is flagged (as it should be) which exercises the `if(isset(...)){...}` pattern you are concerned about.

```php
<?php

function minimalexample() {
    $VAR = $_GET['someval']
    if(isset($VAR)){ 
        // ok: echoed-request
        echo isset($VAR);
        // ruleid: echoed-request
        echo $VAR; 
        //ok: echoed-request
        echo isset($VAR) . "something";
        // ruleid: echoed-request
        echo isset($VAR) . $VAR;
    }

}

```

shellz-n-stuff:
> @shellz-n-stuff totally understand your concern - you'll note that in the correct behavior with the `isset` sanitizer (https://semgrep.dev/playground/s/5rz6W) line 9 is flagged (as it should be) which exercises the `if(isset(...)){...}` pattern you are concerned about.
> 
> ```
> <?php
> 
> function minimalexample() {
>     $VAR = $_GET['someval']
>     if(isset($VAR)){ 
>         // ok: echoed-request
>         echo isset($VAR);
>         // ruleid: echoed-request
>         echo $VAR; 
>         //ok: echoed-request
>         echo isset($VAR) . "something";
>         // ruleid: echoed-request
>         echo isset($VAR) . $VAR;
>     }
> 
> }
> ```

Wow, so semgrep's taint tracker mode tracks assignment. That's absolutely unreal. I'll fix this now, I'm super interested in how y'all got that to work in pass by value vs pass by reference.

Apologies for my misunderstanding!

kurt-r2c (approved):

kurt-r2c:
String interpolation tests are failing - investigating. Not sure why the removal of the parentheses cause this test to start failing.

php/lang/security/injection/echoed-request.php
```diff
@@ -44,6 +44,22 @@ function doSmth5() {
     echo "Hello ".trim($_POST['name']);
 }
 
+function doSmth6() {
+   $VAR = $_GET['someval']
```
kurt-r2c: typo, missing semicolon - missed this a couple times, causing a parse error 
```suggestion
   $VAR = $_GET['someval'];
```

php/lang/security/injection/printed-request.php
```diff
@@ -0,0 +1,112 @@
...
+function doSmth6() {
+    $VAR = $_GET['someval']
+     if(isset($VAR)){ 
+         // ruleid: printed-request
+         print($VAR); 
+     }
+ }
+ 
+ function doSmth7() {
+     $VAR = $_GET['someval']
```
kurt-r2c: typo, missing semicolon
```suggestion
     $VAR = $_GET['someval'];
```

php/lang/security/injection/printed-request.php
```diff
@@ -0,0 +1,112 @@
...
+    print("Hello ".htmlentities($_POST['name'])." !".$_POST['lastname']);
+}
+
+function doSmth5() {
+     // ruleid: printed-request
+    print("Hello ".trim($_POST['name']));
+}
+
+function doSmth6() {
+    $VAR = $_GET['someval']
```
kurt-r2c: typo, missing semicolon
```suggestion
    $VAR = $_GET['someval'];
```

php/lang/security/injection/echoed-request.php
```diff
@@ -44,6 +44,22 @@ function doSmth5() {
...
+    if(isset($VAR)){ 
+        // ruleid: echoed-request
+        echo $VAR; 
+    }
+}
+
+function doSmth7() {
+    $VAR = $_GET['someval']
```
kurt-r2c: typo, missing semicolon
```suggestion
    $VAR = $_GET['someval'];
```

kurt-r2c (dismissed):

shellz-n-stuff:
Oh good catch. Thank you @kurt-r2c, my apologies!

kurt-r2c (approved):

PR #3355, opened 2024-04-16
[Security] Added edge cases for tainted pickle deserialization

theinfosecguy:
- Added `pattern-not` conditions to exclude safe usage patterns and minimize false positives.
- Added new test cases to cover a wider range of scenarios, including both vulnerable and safe usage patterns.

The patterns like `pickle.load(pickle.dumps(...))` are added to `pattern-not` to avoid false positives in situations where the data being deserialized is explicitly serialized using pickle.dumps() within the same codebase.


0xDC0DE (approved):

0xDC0DE:
Looks like a good update to me! Thanks for your contribution!

python/aws-lambda/security/tainted-pickle-deserialization.py
```diff
@@ -29,3 +29,27 @@ def lambda_handler(event, context):
...
+  dill.loads(dill.dumps(event['safe_data']))
+
+  # ruleid: tainted-pickle-deserialization
+  pickle.load(pickle.dumps(event['exploit_code']))
+
+  # ruleid: tainted-pickle-deserialization
+  _pickle.loads(_pickle.dumps(event['exploit_code']))
+
+  # ruleid: tainted-pickle-deserialization
+  cPickle.load(cPickle.dumps(event['exploit_code']))
```
0xDC0DE, line 52: Looks like these are incorrectly annotated. If I understand correctly, you don't want these to match. This should be fixed before we can merge these changes.

0xDC0DE (changes requested):

0xDC0DE:
Closing due to inactivity

PR #3360, opened 2024-04-19
Upload dockerd socket mount detection rule and test file

r0075h3ll:
This PR adds a new rule, along with a test file, for dockerd unix socket file i.e. `docker.sock` mount detection. 

#### Description

Docker works in client-server architecture. In a normal setup, `docker` command issued in the CLI interacts with the docker server(`dockerd`) using a Remote API or Unix Socket File(`docker.sock`) - the latter is often encountered. However, in a scenario where an attacker breaks into the container that has `/var/run/docker.sock` mounted, it can be leveraged to execute arbitrary commands on the host machine i.e. to talk back to the host machine thereby providing a way for escaping the container.

#### Example

Here's a dummy Dockerfile that mounts docker.sock file:
```Dockerfile
FROM docker:latest

WORKDIR /app

VOLUME /var/run/docker.sock:/var/run/docker.sock

CMD ["docker", "images"]
```

Upon running Semgrep with the provided rule, this is how the output looks like: 
image

Thanks.


ievans:
thanks for the contribution! can you add a positive and negative test case annotation to the code? cf https://semgrep.dev/docs/writing-rules/testing-rules/

r0075h3ll:
@ievans 

Done :)

0xDC0DE (approved):

0xDC0DE:
This looks good to me, thanks for the contribution!

r0075h3ll:
Hey @0xDC0DE, can we merge?

java/lang/security/audit/xxe/documentbuilderfactory-disallow-doctype-decl-missing.yaml
```diff
@@ -94,62 +97,64 @@ rules:
       - by-side-effect: true
         pattern-either:
           - patterns:
-            - pattern-either:
```
p4p3r: I think you inadvertently pushed changes to this rule which is unrelated to your PR. Could you please revert?

p4p3r (changes requested):

java/lang/security/audit/xxe/documentbuilderfactory-disallow-doctype-decl-missing.yaml
r0075h3ll: My bad. I reverted the changes.

p4p3r (approved):
lgtm

dockerfile/security/dockerd-socket-mount.yaml
```diff
@@ -0,0 +1,36 @@
+rules:
+  - id: dockerfile-dockerd-socket-mount
+    message: >-
+        The Dockerfile(image) mounts docker.sock to the container which may allow an attacker already inside of the container
+        to escape container and execute arbitrary commands on the host machine.
+    languages:
+      - dockerfile
+      - yaml
```
aryx, line 8: @0xDC0DE why we have yaml listed as a language here?

PR #3369, opened 2024-05-02
Update allow-privilege-escalation-no-securitycontext.yaml

mterron:
Remove random $NAME that breaks the Fix

Before:
`Autofix ▶ securityContext:allowPrivilegeEscalation: falsename`

After
`Autofix ▶ securityContext:allowPrivilegeEscalation: false`


inkz:
`$NAME` - in this case is used for identifying where to apply the fix. If it is done incorrectly please make sure to use the newest version of Semgrep. If the problem remains please open an issue at https://github.com/semgrep/semgrep-rules/issues and provide a code snippet that is incorrectly fixed.

PR #3398, opened 2024-06-11
Update rails rule with skip_before_action

p4p3r:
Include skip_before_action, which is used in Rails 5 and onwards and behaves similarly/the same as skip_before_filter did previously.


armchairlinguist (approved):
LGTM, though I don't know if my review will unblock!

PR #3402, opened 2024-06-14
Allow markup-ing literal strings

xmo-odoo:
Literal strings in the application should be safe (similar to static markup in template files), and the normal way to create dynamic markup code side: create a properly marked up `Markup`, then `Markup.format` user-defined content into it.

### DIFF
--- a/python/flask/security/xss/audit/explicit-unescape-with-markup.py
+++ b/python/flask/security/xss/audit/explicit-unescape-with-markup.py
@@ -14,11 +14,12 @@ def markup_test():
 
     playlist = request.args.get('p')
     if playlist:
-        playlist = '"{0}"'.format(playlist.replace('\"', '\\\"').strip())
+        # ok: explicit-unescape-with-markup
+        playlist = Markup('"{0}"').format(playlist.replace('\"', '\\\"').strip())
     else:
         playlist = '""'
     # ruleid: explicit-unescape-with-markup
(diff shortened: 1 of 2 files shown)


xmo-odoo:
Will need to squash the PR in order to use the PR's description as commit message: had a typo ("and the**n** normal way")

0xDC0DE (approved):

xmo-odoo:
CI seems unhappy, even though 3.9.2 does seem available in the python-versions and [#3404](https://github.com/semgrep/semgrep-rules/pull/3404) seems to have fetched it fine.

Possibly a temporary network issue?

xmo-odoo:
Aaah I think I see what the issue is, I forked off of and targeted the `main` branch, not sure how (since I just used github's base UI).

`develop` seems to be the default and correct branch to target. I retargeted the PR and rebased, and updated the commit message to fix the typo while at it. Hopefully it works better like this.

xmo-odoo:
Right, yep, turns out using github's interface led me to not test the rule and thus forget half of it.

Also wow does the test suite not joke around. But since I was running it I added a few test cases to make sure the change actually works, probably can't hurt. Hopefully I got it right this time.

xmo-odoo:
Thanks :fireworks:

p4p3r:
Thanks for your contribution! Your changes will be available in Semgrep’s registry in a bit.

PR #3409, opened 2024-06-20
Exclude tests files in golang

hoeg:
Writing tests for HTTP handlers where we are calling localhost will get matched by the `http-customized-request` rule.

In golang all test files are postfixed with `_test.go` (https://go.dev/doc/tutorial/add-a-test). We suggest that we exclude those files as they are not part of the production code shipped with go build. Alternatively we can check the hostname directly and ensure it is not `localhost` and `127.0.0.1`.

### DIFF
--- a/problem-based-packs/insecure-transport/go-stdlib/http-customized-request.yaml
+++ b/problem-based-packs/insecure-transport/go-stdlib/http-customized-request.yaml
@@ -1,5 +1,8 @@
 rules:
 - id: http-customized-request
+  paths:
+    exclude:
+      - "**/*_test.go"
   message: >-
     Checks for requests sent via http.NewRequest to http:// URLS. This is dangerous because
     the server is attempting to connect to a website that does not encrypt traffic with TLS. Instead, send


inkz:
We suggest excluding the tests and other files that should not be scanned in the CLI,
```
       --exclude=PATTERN
           Skip any file or directory whose path that matches PATTERN.
           '--exclude=*.py' will ignore the following: 'foo.py',
           'src/foo.py', 'foo.py/bar.sh'. '--exclude=tests' will ignore
           'tests/foo.py' as well as 'a/b/tests/c/foo.py'. Multiple
           '--exclude' options may be specified. PATTERN is a glob-style
           pattern that uses the same syntax as gitignore and semgrepignore,
           which is documented at
           https://git-scm.com/docs/gitignore#_pattern_format 
```

link to the docs: https://semgrep.dev/docs/cli-reference

in your case it can be something like:
`semgrep -c problem-based-packs/insecure-transport/go-stdlib/http-customized-request.yam --exclude="**/*_test.go" dir_to_scan/`

PR #3427, opened 2024-07-10
Add rule to check for insecure load via joblib

theinfosecguy:
Added rule to find the unsafe usage of `joblib.load` when loading untrusted data. Loading untrusted data with `joblib.load` can lead to arbitrary code execution through deserialization vulnerabilities.

Also added `pattern-sanitizers` to avoid false positives due to `safe_mode=True`.


0xDC0DE (approved):

inkz (changes requested):
Can you please update the rule and remove taint mode from it. Do smth like this:
```yaml
patterns:
- pattern: joblib.load($X,...)
- pattern-not: joblib.load($X, safe_mode=True, ...)
- focus-metavariable: $X
```

In your rule you are using "$UNTRUSTED_DATA" as a source of taint, which means literally everything and is going to be very inefficient, I believe this is why the PR is failing in CI.
For such cases, where you do not know what the source is exactly, we have a good old "search" mode rules :+1:

0xDC0DE:
Closing this PR as it is stale

PR #3434, opened 2024-07-18
Fix python.lang.correctness.baseclass-attribute-override

IagoAbal:
The rule depended on Semgrep unifying two identifiers, corresponding to methods in two different classes, that were not the same but just had the same "string" name.


akuhlens (approved):
🙈

PR #3438, opened 2024-07-26
Fixed the dockerfile.security.missing-user rule

### LINKED ISSUE #3436: dockerfile.security.missing-user has a false positive related to HEALTHCHECK CMD
**Describe the bug**
dockerfile.security.missing-user has a false positive related to HEALTHCHECK CMD. It triggers for Dockerfiles that do not have a CMD directive and only use an ENTRYPOINT directive.

**To Reproduce**
```
FROM some.registry/java/jre-17-corretto-amazonlinux-2023:latest

WORKDIR /service
EXPOSE 18252 18253

HEALTHCHECK --interval=30s --timeout=3s \
  CMD curl -sSfI --retry 0 --tcp-nodelay http://localhost:18253/actuator/health/liveness || exit 1

USER user

COPY ./scripts/entrypoint.sh /service/bin/
COPY build-artifact /service/bin/build-artifact.jar
ENTRYPOINT ["/service/bin/entrypoint.sh"]
```

**Expected behavior**
This rule should not trigger because a non-root user is specified. 

**Priority**
How important is this to you?

**Additional Context**
I don't think that this rule should trigger in the first place as this isn't the CMD directive, it's part of the HEALTHCHECK. I believe the intent of this rule is to prevent services to run as root so as to minimize exposure if the service is compromised. I'm not sure how you take advantage of the HEALTHCHECK calling curl with root permissions to cause problems.

There is another missing-user rule for ENTRYPOINT directive. This is odd because the CMD and ENTRYPOINT directives can both be used in a Dockerfile and this is valid. There is even a[ documented example](https://docs.docker.com/reference/dockerfile/#exec-form-entrypoint-example) as such. I would think that this should be one rule given this fact.

Lastly, my understanding is that any ENTRYPOINT or CMD directive uses the last declared user in the Dockerfile. I don't believe the USER directive is required to be declared before the ENTRYPOINT or CMD directive. If I understand both of these rules correctly, they assume that the USER directive must be declared before the ENTRYPOINT and CMD directive.

saghaulor:
- The previous version of this rule would have false positives on
```
HEALTHCHECK ... \
  CMD ...
ENTRYPOINT ...
```
and
```
HEALTHCHECK ... \
CMD ...
ENTRYPOINT ...
```
  It doesn't really make sense to flag on the CMD sub-directive of the HEALTHCHECK
  directive since there's very little chance that the application could be compromised
  via the HEALTHCHECK and then gain root access. This false positive creates a lot of noise
  and therefore we're addressing it.
- There was a separate rule for ENTRYPOINT, which doesn't really make sense,
  since CMD and ENTRYPOINT can be used in the same Dockerfile,
  as per https://docs.docker.com/reference/dockerfile/#exec-form-entrypoint-example
  Therefore, the rule was removed
- Fixed the bug that will create two findings for a Dockerfile like this
```
FROM busybox
ENTRYPOINT ["some-command"]
CMD ["--some-arg"]
```
- The autofix arguments have changed because technically it doesn't matter
  where in the Dockerfile the USER directive is specified, insofar as the
  last specified USER is non-root. Previously, the autofix would attempt
  to add the USER directive above the CMD or ENTRYPOINT directives.
  However, since either or both of these can appear, we're not going to
  specify the CMD or ENTRYPOINT directive in the fix.
- Cleaned up some of the test files to remove invalid syntax like calling CMD twice
- Fixes #3436


p4p3r:
Hello @saghaulor and thanks for your contribution!
The rules were indeed split to allow to add a fix. Your fix will not work because the way it works is that Semgrep replaces the whole match whatever is specified by the fix key: in your case it will replace, for instance, `ENTRYPOINT "semgrep --config p/xss"` with `USER non-root`!

I would suggest to keep the rules separated but rewrite the one about the CMD to match it only when ENTRYPOINT is not present:
* missing-user: https://semgrep.dev/playground/s/kxOpK
* missing-user-entrypoint: https://semgrep.dev/playground/s/bwjze

As you can notice, in the `missin-user` rule, I'm using a slightly different (and more straightforward) way to make sure I don't match the HEALTCHECK command. This is not going to be working right now, but it will as soon as a new Semgrep release is published next week (this is the relevant change that I'm waiting for: https://github.com/semgrep/semgrep/commit/c697cdf772760110ad084138219042952d10dfcf).
This should allow us to reduce FPs without regexes :-)

p4p3r:
Also, the way Semgrep's testsuite works is that it will match rule file name and test file name. In your PR, `dockerfile/security/missing-user-entrypoint-cmd.dockerfile` doesn't have a matching rule file (that would be `dockerfile/security/missing-user-entrypoint-cmd.yaml`), so it will be ignored.

saghaulor:
@p4p3r thanks for the context and the engine fix, I definitely wasn't excited about using regexes for the solution. I just tested it out and it's not longer triggering false positives on the `HEALTHCHECK ... CMD` directives. It's also not triggering the missing-user rule for Dockerfiles with ENTRYPOINT and CMD.

Do you agree that these are valid dockerfiles?

```dockerfile
FROM busybox

COPY thing /usr/bin/thing

ENTRYPOINT ["/usr/bin/thing"]

USER non-root
```

```dockerfile
FROM busybox

COPY thing /usr/bin/thing

CMD ["/usr/bin/thing"]

USER non-root
```

My understanding is that the ENTRYPOINT and/or CMD will be executed as user non-root as it is the last declared user, and if so, there's still a bug in the rule. If my understanding is correct, then that means that there will be false positives related to the above examples.

Also, have y'all considered restructuring how your test suite applies to rules? It seems that currently it will necessitate creating new rules for edge cases, or creating test files that are not valid to include edge cases so that you don't have to introduce new rules.

Overall, I think most of the false positives have been addressed, but the above examples will still mean that at least one false positive exists (assuming that I have understood dockerfiles correctly).

saghaulor:
I've opened a new PR that addresses the bug related to a `USER` directive not being recognized after a `CMD` or `ENTRYPOINT` directive. I felt like it was better to do it that way so that we can keep the context of this conversation but not make the PR confusing, since this conversation was about the HEALTHCHECK issue and the new PR is about the aforementioned bug.

PR #3440, opened 2024-07-31
Fix CSharp SQLI rule

LewisArdern:
Reported by user where taint tracking was 'tainting' due to a string entering `SqlConnection` first argument, and then a static string was flagging as a finding. 

Community slack context https://semgrep.slack.com/archives/C031LV9H6P9/p1722445805060099


inkz (approved):
Nice!

PR #3455, opened 2024-08-19
fix(naming-regressions): more updates to semgrep-rules 

yosefAlsuhaibani:
See https://github.com/semgrep/semgrep-proprietary/pull/2105/#issue-2487634317


python/sqlalchemy/security/sqlalchemy-execute-raw-query.py
```diff
@@ -198,6 +198,7 @@
 # ok: sqlalchemy-execute-raw-query
 engine = create_engine('postgresql://user@localhost/database')
 query = select(literal_column("users.fullname", String) + ', ' + literal_column("addresses.email_address").label("title")).where(and_(literal_column("users.id") == literal_column("addresses.user_id"), text("users.name BETWEEN 'm' AND 'z'"), text("(addresses.email_address LIKE :x OR addresses.email_address LIKE :y)"))).select_from(table('users')).select_from(table('addresses'))
+# deepruleid: sqlalchemy-execute-raw-query
 conn.execute(query, {"x":"%@aol.com", "y":"%@msn.com"}).fetchall()
```
yosefAlsuhaibani, line 202: I am not sure if this is true; I'll have to inspect this later this week

python/pycryptodome/security/insecure-cipher-algorithm-blowfish.py
```diff
@@ -25,10 +25,12 @@
 bs = pycrypto_blowfish.block_size
 # ruleid:insecure-cipher-algorithm-blowfish
 cipher = pycrypto_blowfish.new(key, pycrypto_blowfish.MODE_CBC, iv)
+# deepruleid:insecure-cipher-algorithm-blowfish
 msg = iv + cipher.encrypt(plaintext + padding)
 bs = pycryptodomex_blowfish.block_size
 # ruleid:insecure-cipher-algorithm-blowfish
 cipher = pycryptodomex_blowfish.new(key, pycryptodomex_blowfish.MODE_CBC, iv)
+# deepruleid:insecure-cipher-algorithm-blowfish
 msg = iv + cipher.encrypt(plaintext + padding)
```
yosefAlsuhaibani, line 35: NOTE: all of these pycryptodome findings were introduced as result of uncommenting the line that wipes id_info in generic v generic; I am not 100% sure if these findings are expected. Will need to check if this is indeed the right behavior.

python/pycryptodome/security/insecure-cipher-algorithm-rc2.py
```diff
@@ -19,9 +19,11 @@
 iv = Random.new().read(pycrypto_arc2.block_size)
 # ruleid:insecure-cipher-algorithm-rc2
 cipher = pycrypto_arc2.new(key, pycrypto_arc2.MODE_CFB, iv)
+# deepruleid:insecure-cipher-algorithm-rc2
 msg = iv + cipher.encrypt(b'Attack at dawn')
```
yosefAlsuhaibani, line 23: Why `deepruleid:`? symbolic propagation propagates the expression `pycrypto_arc2.new(...)` to cipher, resulting in two results: one on the def of cipher and the other is on its use. This should be the case for all of the pycryptodome rules

java/lang/security/audit/xss/no-direct-response-writer.java
```diff
@@ -191,7 +191,7 @@ public void doPost(HttpServletRequest request, HttpServletResponse response)
 
         response.setHeader("X-XSS-Protection", "0");
         Object[] obj = {"a", bar};
-        // ruleid: no-direct-response-writer
+        // ruleid: deepok: no-direct-response-writer
```
yosefAlsuhaibani, line 215: From iago:
```ocaml
  (* Java/tainting:
     * This is a test that comes from the OWASP Benchmark v1.2.
     * Here DeepSemgrep doesn't report lines 56, 128, and 195.
     * But these are actually false positives! This benchmark tries
     * to confuse analyzers into reporting these false positives.
     * It does this in two ways, 1) by using a third-function
     * `doSomething` that receives tainted data, even though it
     * returns safe data; and 2) by putting both safe and unsafe
     * data into a `HashMap`, but ultimately only returning the
     * safe data. FOSS Semgrep falls into the first trap.
     * DeepSemgrep does inter-procedural analysis so it is only
     * affected by the second trap, but it seems to not fall
     * into it because we are lacking a `pattern-propagators` spec
     * for `HashMap`s. If we told DeepSemgrep that `HashMap`s
     * store/propagate taint, then it should report the same
     * false positives.
  *)
```

python/pycryptodome/security/insecure-cipher-algorithm-blowfish.py
```diff
@@ -25,10 +25,12 @@
 bs = pycrypto_blowfish.block_size
 # ruleid:insecure-cipher-algorithm-blowfish
 cipher = pycrypto_blowfish.new(key, pycrypto_blowfish.MODE_CBC, iv)
+# deepruleid:insecure-cipher-algorithm-blowfish
 msg = iv + cipher.encrypt(plaintext + padding)
```
kopecs, line 29: Why is symbolic prop on for this rule?
https://github.com/semgrep/semgrep-rules/blob/de3f55438bd280eafedcae0ae4daae1ce61d89c0/python/pycryptodome/security/insecure-cipher-algorithm-blowfish.yaml#L30-L31

python/pycryptodome/security/insecure-cipher-algorithm-des.py
```diff
@@ -23,11 +23,13 @@
 ctr = Counter.new(pycrypto_des.block_size*8/2, prefix=nonce)
 # ruleid:insecure-cipher-algorithm-des
 cipher = pycrypto_des.new(key, pycrypto_des.MODE_CTR, counter=ctr)
+# deepruleid:insecure-cipher-algorithm-des
 msg = nonce + cipher.encrypt(plaintext)
```
kopecs, line 27: Similar.

python/pycryptodome/security/insecure-cipher-algorithm-rc2.py
kopecs: Similar as above

python/sqlalchemy/security/sqlalchemy-execute-raw-query.py
kopecs: Why does this only match with deep?

kopecs (approved):
The changes seem fine but I am not sure why the crypto rules have symbolic prop at all. Can we investigate why before merging?

java/lang/security/audit/xss/no-direct-response-writer.java
kopecs: This should probably be a comment in the test here if you're removing it from the overrides in pro. I could see this coming back depending on how sensitive we are to taint in HashMaps and it would be useful to have that context.

python/sqlalchemy/security/sqlalchemy-execute-raw-query.py
yosefAlsuhaibani: I believe `query` gets sym prop'd (i.e `query = select(...) + ...`) so then you'd get `conn.execute(query, ...)` which might match to this part of the rule?

https://github.com/semgrep/semgrep-rules/blob/fb248f4c8ecd0474594180c1187fba068932cf84/python/sqlalchemy/security/sqlalchemy-execute-raw-query.yaml#L43-L48

PR #3456, opened 2024-08-20
Fix rule extension: yml -> yaml

IagoAbal:
Semgrep Pro's test engine expects `.yaml`. Maybe we need to revisit that, but anyways we should be consistent, and if we use `.yaml` for every rule already then better to stick to that convention.


IagoAbal:
The rule must have been merged previously despite the "validate-r2c-registry-metadata" check failing, or perhaps it didn't exist, because I have not changed the rule at all.

yosefAlsuhaibani:
re: `.yaml` vs `.yml`: in looking at naming regressions https://github.com/semgrep/semgrep-proprietary/pull/1979; I added the ability for `Test.ml` to accept .yml/src pairs (as well as accepting files that have multiple extensions like `build-gradle-password-hardcoded.build.gradle.kts`); is this even worthwhile doing? or should expect that we rename and stick to the convention of .yaml only?

IagoAbal:
> re: `.yaml` vs `.yml`: in looking at naming regressions [semgrep/semgrep-proprietary#1979](https://github.com/semgrep/semgrep-proprietary/pull/1979); I added the ability for `Test.ml` to accept .yml/src pairs (as well as accepting files that have multiple extensions like `build-gradle-password-hardcoded.build.gradle.kts`); is this even worthwhile doing? or should expect that we rename and stick to the convention of .yaml only?

We need some discussion around this, at this very moment I think it's better to keep the convention and use `.yaml` and if the test suite enforces this then it's good. The `.yml` extension could be used for YAML target files perhaps, although I think we have another convention, `.target.yaml` ? Anyways, right now I am quite sure it's better to just fix the extension, and have the discussion later.

inkz (dismissed):
There is a a rule for `build-gradle-password-hardcoded.build.gradle.kts` test here:
 https://github.com/semgrep/semgrep-rules/blob/develop/kotlin/gradle/security/build-gradle-password-hardcoded.yaml#L2
So lets please not delete it :pray:

IagoAbal:
> There is a a rule for `build-gradle-password-hardcoded.build.gradle.kts` test here: https://github.com/semgrep/semgrep-rules/blob/develop/kotlin/gradle/security/build-gradle-password-hardcoded.yaml#L2 So lets please not delete it 🙏

Aha! OK I'll fix that, but we need to improve reviews or something, it's not great that this happens, I'm not sure if that rule was being tested because the target file doesn't match what we usually expect.

IagoAbal:
> > There is a a rule for `build-gradle-password-hardcoded.build.gradle.kts` test here: https://github.com/semgrep/semgrep-rules/blob/develop/kotlin/gradle/security/build-gradle-password-hardcoded.yaml#L2 So lets please not delete it 🙏
> 
> Aha! OK I'll fix that, but we need to improve reviews or something, it's not great that this happens, I'm not sure if that rule was being tested because the target file doesn't match what we usually expect.

Hmm, damn, there is an `include` that requires that kind of name.

LewisArdern (approved):

IagoAbal:
OK I just keep the fix to the rule extension. We'll just XFAIL `build-gradle-password-hardcoded` for now. We need to have a single "test engine", there is way too much duplication in that area.

PR #3457, opened 2024-08-21
python: Future-proof use-ftp-tls for stdlib libdefs

nmote:
Explanation in the rule

### DIFF
--- a/python/lang/security/audit/insecure-transport/ftplib/use-ftp-tls.yaml
+++ b/python/lang/security/audit/insecure-transport/ftplib/use-ftp-tls.yaml
@@ -1,6 +1,18 @@
 rules:
 - id: use-ftp-tls
-  pattern: ftplib.FTP(...)
+  patterns:
+    - pattern: ftplib.FTP(...)
+    # With stdlib libdefs, Semgrep knows that `ftplib.FTP_TLS` is a subclass of
+    # `ftplib.FTP`, and therefore the pattern `ftplib.FTP` matches when we
+    # encounter `ftplib.FTP_TLS` too.
+    #
+    # Therefore, we explicitly exclude `FTP_TLS`.
+    #
+    # Currently libdefs are only available with the interfile engine, and since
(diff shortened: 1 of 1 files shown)


yosefAlsuhaibani (approved):
Yeah this looks good: appreciate the in depth comment!

nmote:
Thanks! I'll wait a bit to merge in case someone from the Code team wants to weigh in.

inkz:
@nmote `ftplib.FTP(...)` pattern should always match only `ftplib.FTP`, adding exclusion for FTP_TLS would make sense if we had something like this:
```
  patterns:
    - pattern: ftplib.$WHATEVER(...)
    - pattern-not: ftplib.FTP_TLS(...)
```

I recommend adding test case with FTP_TLS first, then see how the rule performs and if updates needed add them

kopecs:
@inkz I think Nat is saying that in ftplib `FTP_TLS` subclasses `FTP`, and we have the pattern `A` match `B` when `A` is a subclass of `B`. Previously we didn't know that `FTP_TLS` inherited from `FTP`, but we now do/will know with interfile.

nmote:
@inkz In fact there is such a test case, and it failed on the branch I'm using to develop Python libdefs, for the reason that @kopecs explained. That's why I put up this PR.

nmote:
This is the test case that fails when this rule is run using libdefs: https://github.com/semgrep/semgrep-rules/blob/a326e6f005b08bc7823d5483a3392a1cafca913e/python/lang/security/audit/insecure-transport/ftplib/use-ftp-tls.py#L9-L10

kopecs (approved):

inkz:
Interesting! thanks for clarification @nmote @kopecs 
is information about `ftplib` classes hardcoded in the engine?

nmote:
It will be for the interfile engine, just like we currently hardcode information about the Java standard library.

inkz:
@nmote Cool! thanks

PR #3460, opened 2024-08-26
fix: no pro finding for raw-html-concat.js:18

kopecs:
We should not expect a pro finding here since the tainted `newContent` is not in scope, so this actually throws `ReferenceError`. See also discussion on SAF-1013.

Cf. a version written with `var`, where hoisting would mean that `newContent` would be in scope, and the test should generate a finding.

### DIFF
--- a/javascript/browser/security/raw-html-concat.js
+++ b/javascript/browser/security/raw-html-concat.js
@@ -14,7 +14,7 @@ $(function ($) {
                 
                 var x = `<div align="left">${content}</div>`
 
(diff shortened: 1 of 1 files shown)


neuroo (approved):

kopecs:
Hmm looks like we may only allow `deepok` second. https://github.com/semgrep/semgrep-proprietary/blob/740675e6e29744a1146bc842ef45fe05b68808de/OSS/src/osemgrep/cli_test/Test_subcommand.ml#L176-L185

PR #3464, opened 2024-09-02
PHP tainted-callable

Sjord:
A callable is the name of a function, or an array with a class/object and a method. Basing these on user input makes it possible to call arbitrary functions.


0xDC0DE (approved):
Looks good to me. Thanks for the contribution!

PR #3467, opened 2024-09-05
chore: put ruleid annotation alone on its own line for tainted-sql-string.py

aryx:
This is the only file doing that, so let's be consistent.
It also helps osemgrep test which does not handle this case.

This was mentioned in
https://linear.app/semgrep/issue/SAF-1529/same-line-annotations-fail-when-running-semgrep-test-but-work-with

test plan:
make test


p4p3r (approved):
Lgtm

PR #3468, opened 2024-09-05
PHP tainted exec

Sjord:
When user input is passed to a function that executes a shell command, without escaping.


0xDC0DE (approved):

0xDC0DE:
Hi @Sjord !

This looks good to me. Thanks for yet another contribution! 🔥

php/lang/security/injection/tainted-exec.yaml
```diff
@@ -0,0 +1,51 @@
+rules:
+- id: tainted-exec
+  languages:
+  - php
+  severity: WARNING
+  message: <-
```
0xDC0DE:
```suggestion
  message: >-
```

We usually go for this operator. It might change how multiline strings are displayed in the CLI. I'm not sure anymore what the exact reason is. I think there's a CI check for this as well.

0xDC0DE: This recommendation is included in the docs as well:
https://semgrep.dev/docs/contributing/contributing-to-semgrep-rules-repository#rule-messages

PR #3474, opened 2024-09-19
remove fingerprints/fingerprints.yaml

aryx:
No idea what this file is, but it's annoying because
we have to skip it in many scripts because it does not
contain regular rules and target test files.
Let's just remove it to simplify things.

test plan:
wait for green CI checks


p4p3r (approved):
lgtm

PR #3481, opened 2024-09-26
run osemgrep validate --pro

aryx:
test plan:
osemgrep validate --pro .


aryx:
takes 12s (against 1m11 with pysemgrep).

aryx:
(but the pysemgrep run the validation for each language dir, so it calls the registry for the metachecking rules that many times).

p4p3r (approved):
lgtm

PR #3487, opened 2024-10-14
add sanitizer and update message of dangerous subprocess rule

### LINKED ISSUE #3485: Once triggered, `dangerous-subprocess-use-tainted-env-args` cannot be addressed
**Describe the bug**

The subject rule can be triggered in some valid circumstances. But the suggested recipe is misleading or does not work at all.

1. The rule description mentioning using 'shlex.escape()' is wrong because `shlex.escape` is not a function (I suggest to replace it with `shlex.quote()`).
2. Using `shlex.quote` is not accepted by the rule. As a result, I ended up ignoring it. I am not sure if currently there is any reasonable way to address the rule once triggered.

**To Reproduce**

Try adding `shlex.quote` to the rule test cases and observe nothing changed in the static analysis report.

**Expected behavior**

The recipe to fix the triggering rule in the rule description works in practice.

**Priority**
How important is this to you?

0xDC0DE:
Fixes https://github.com/semgrep/semgrep-rules/issues/3485

### DIFF
--- a/python/lang/security/audit/dangerous-subprocess-use-tainted-env-args.yaml
+++ b/python/lang/security/audit/dangerous-subprocess-use-tainted-env-args.yaml
@@ -3,6 +3,8 @@ rules:
   mode: taint
   options:
     symbolic_propagation: true
+  pattern-sanitizers:
+    - pattern: shlex.quote(...)
   pattern-sources:
   - patterns:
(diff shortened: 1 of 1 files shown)


neuroo (approved):
🐑

PR #3490, opened 2024-10-18
Fix for osemgrep test --pro with DeepScan too (often deepruleid: -> deeptodoruleid:)

aryx:
Mostly some deepruleid: -> deeptodoruleid: as unfortunately
the engine is still not good enough to find them

test plan:
osemgrep-pro test --pro .


aryx:
engine changes here: https://github.com/semgrep/semgrep-proprietary/pull/2424

0xDC0DE (approved):

PR #3493, opened 2024-10-21
Remove scripts/run-test to simplify, call just osemgrep test

aryx:
It has been almost a month that we run both osemgrep test
and pysemgrep --test and no complaints, so let's remove
the use of pysemgrep --test so we can then remove the
corresponding python code in pysemgrep.

test plan:
make validate
make test-only
wait for green CI checks


aryx:
Can I get approval on this one? It's blocking https://github.com/semgrep/semgrep-rules-proprietary/pull/674 which will allow us to finally prevent tests that do not pass CoreScan/ProScan/DeepScan to be merged in.

0xDC0DE (approved):

PR #3499, opened 2024-10-25
Allow OWASP Top 10 references from Kubernetes and LLM Top 10

berney:
Currently the rule only allows references to the web application OWASP Top 10.
The change allows referencing https://owasp.org/www-project-kubernetes-top-ten/ and https://owasp.org/www-project-top-10-for-large-language-model-applications/

There's more Top 10s, but they can probably be added as rules are written and updated.


0xDC0DE (approved):
Looks great. Thanks for your contribution!

PR #3501, opened 2024-10-25
Filter exceptions from tainted-sql-string

berney:
Original rule will think string interpolation in throwing an exception is a SQL injection.
Which makes this rule very noisy.


0xDC0DE (approved):

0xDC0DE:
Hey @berney , this looks like a great update! Thanks for your contribution.

PR #3506, opened 2024-11-04
Add flask-cors-misconfiguration rule

lfama:
Hi!

I've created a simple rule that matches [flask_cors ](https://flask-cors.readthedocs.io/en/latest/index.html ) misconfiguration. 

Basically when setting `origin="*"` and `support_credentials=True` flask_cors dinamically reflects the Origin of each request in the `Access-Control-Allow-Origin` header which together with `Access-Control-Allow-Credential` set to true can be dangerous.

### DIFF
--- a/python/flask/security/audit/flask-cors-misconfiguration.yaml
+++ b/python/flask/security/audit/flask-cors-misconfiguration.yaml
@@ -1,0 +1,36 @@
+rules:
+  - id: flask-cors-misconfiguration
+    message: >-
+      Setting 'support_credentials=True' together with 'origin="*"' is a CORS
+      misconfiguration that can allow third party origins to read sensitive
+      data. Using this configuration, flask_cors will dynamically reflects the
+      Origin of each request in the Access-Control-Allow-Origin header, allowing
+      all origins and allowing cookies and credentials to be sent along with
+      request. It is recommended to specify allowed origins instead of using "*"
+      when setting 'support_credentials=True'.
+    languages:
+      - python
+    severity: WARNING
+    patterns:
+      - pattern-either:
+          - pattern: |
+              @cross_origin(..., origins="*", supports_credentials=True, ...)
+          - pattern: |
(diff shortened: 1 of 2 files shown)


python/flask/security/audit/flask-cors-misconfiguration.yaml
```diff
@@ -0,0 +1,36 @@
...
+      when setting 'support_credentials=True'.
+    languages:
+      - python
+    severity: WARNING
+    patterns:
+      - pattern-either:
+          - pattern: |
+              @cross_origin(..., origins="*", supports_credentials=True, ...)
+          - pattern: |
+              CORS(..., supports_credentials=True, origins="*", ...)
```
0xDC0DE, line 19:
```suggestion
          - pattern: |
              @cross_origin(..., origins="*", ..., supports_credentials=True, ...)
          - pattern: |
              CORS(..., supports_credentials=True, ..., origins="*", ...)

```

0xDC0DE: I'm assuming this annotation has other options that can potentially be set in between these two

0xDC0DE:
Hi! @lfama Thanks for your contribution, this rule looks good! I've made a suggestion to improve your rule slightly. If you address this and the CI checks pass, we can merge this. 🥇

python/flask/security/audit/flask-cors-misconfiguration.yaml
lfama: Hi @0xDC0DE,

thank you for your feedback.
I've tried to keep the rule as simple as possible (by using the minimum amount of ellipsis operators) because I've noticed that semgrep is already able to handle cases where there are other options in between. Please take a look here --> https://semgrep.dev/playground/s/wdE9Y

0xDC0DE: Ooh that's great! Thanks for checking.

0xDC0DE (approved):

PR #3511, opened 2024-11-05
[run-as-non-root-security-context-pod-level] runAsNonRoot does not necessarily comes directly after securityContext

SebGondron:
This rule will create false positives if `runAsNonRoot: true` is not placed directly after `securityContext`. I added the appropriate ellipsis.


inkz (approved):

inkz:
@SebGondron thank you!

PR #3512, opened 2024-11-05
Fix broken link in system-wildcard-detected.yaml

adhintz:
The previous URL is a good original source, but it has been down for years. Replaced it with a gist posting that has the original content.


0xDC0DE (approved):

0xDC0DE:
That seems like a reasonable update to me. I am, however, uncertain of linking to user-controlled content like this. It seems to me that this user could change the contents of this file to something that would not be desirable to show to users of Semgrep. Could we potentially find another resource to link instead?

0xDC0DE:
@adhintz Are you able to update this PR or should I close it?

PR #3517, opened 2024-11-17
added a rule to detect usage of uuid version 1 in python

mohamed-osama-aboelkheir:
Added a Semgrep rule to detect UUID version 1 in Python, which could lead to predictable UUIDs based on system information (e.g., MAC address, timestamp). This may lead to security risks such as the sandwich attack e.g. https://www.landh.tech/blog/20230811-sandwich-attack/.

### DIFF
--- a/python/lang/security/insecure-uuid-version.yaml
+++ b/python/lang/security/insecure-uuid-version.yaml
@@ -3,8 +3,10 @@ rules:
     patterns:
       - pattern: uuid.uuid1(...)
     message: |
-      Using UUID version 1 for UUID generation can lead to predictable UUIDs based on system information (e.g., MAC address, timestamp). This may lead to security risks such as the sandwich attack e.g. https://www.landh.tech/blog/20230811-sandwich-attack/. Consider using `uuid.uuid4()` instead for better randomness and security.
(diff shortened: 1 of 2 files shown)


python/lang/security/insecure-uuid-version.yaml
```diff
@@ -0,0 +1,31 @@
+rules:
+  - id: insecure-uuid-version
+    patterns:
+      - pattern: uuid.uuid1(...)
+    message: |
+      Using UUID version 1 for UUID generation can lead to predictable UUIDs based on system information (e.g., MAC address, timestamp). This may lead to security risks such as the sandwich attack e.g. https://www.landh.tech/blog/20230811-sandwich-attack/. Consider using `uuid.uuid4()` instead for better randomness and security.
+    metadata:
+      cwe:
```
0xDC0DE:
```suggestion
    message: |
      Using UUID version 1 for UUID generation can lead to predictable UUIDs based on system information (e.g., MAC address, timestamp). This may lead to security risks such as the sandwich attack. Consider using `uuid.uuid4()` instead for better randomness and security.
    metadata:
      references:
        - https://www.landh.tech/blog/20230811-sandwich-attack/
      cwe: 
```

Hi @mohamed-osama-aboelkheir . Thanks for your contribution, this rule looks solid. 
Before merging, can I ask you to remove the url from the message and add it to the metadata?

0xDC0DE (changes requested):

0xDC0DE (approved):

PR #3530, opened 2024-12-10
Only detect Sauce API token if its a V4 GUID

Sjord:
The regex already implicitly matched a GUID, but this makes it more strict. This solves false positives where another hexadecimal string follows "sauce", such as in yarn.lock files.


0xDC0DE (approved):
Looks good to me!

PR #3541, opened 2024-12-18
Also find weak crypto in PHP when hash(...) is called

Sjord:
MD5 and SHA1 are insecure hash functions. These have their own function names (`md5(...)`, `sha1(...)`) but can also be calculated using `hash('md5', ...)` and `hash('sha1', ...)`. Also find these instances and report them as weak crypto.

Also, write out all function names as function calls instead of matching them with a regular expression, for readability and performance reasons.


p4p3r (approved):
lgtm

PR #3542, opened 2024-12-18
Add rule for SHA224 hash in PHP, Java, Go, Python, Ruby

Sjord:
SHA224 and other hashes with 224 bits are currently on the edge of what is considered a secure hash function. The [Australian Cyber Security Center](https://www.cyber.gov.au/resources-business-and-government/essential-cyber-security/ism/cyber-security-guidelines/guidelines-cryptography) and [NIST](https://nvlpubs.nist.gov/nistpubs/SpecialPublications/NIST.SP.800-131Ar3.ipd.pdf) have indicated that SHA224 is going to be deprecated in the future.

This adds a rule to detect SHA224 and similar hashes. It is distinct from the existing MD5 and SHA1 rules, because those are currently a security problem, and SHA224 is more of a future compliance problem.

### DIFF
--- a/go/lang/security/audit/crypto/sha224-hash.go
+++ b/go/lang/security/audit/crypto/sha224-hash.go
@@ -1,0 +1,43 @@
+package main
+
+import (
+	"crypto/sha256"
+	"golang.org/x/crypto/sha3"
+	"fmt"
+	"io"
+	"log"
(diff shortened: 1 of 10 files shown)


p4p3r (approved):
lgtm

PR #3543, opened 2024-12-21
Improve error message for i18next key format rule

nathanielkohl83:
- Updated the error message in the i18next key format rule to provide clearer guidance on valid key formats.
- Clarified that the module and feature parts of the translation key must use only lowercase letters (a-z) and hyphens (-).
- Highlighted that only the subkey part may include uppercase letters, underscores (_), or periods (.).
- Replaced the misleading all-caps example with a more accurate and valid example: my-module.my-feature.my_key.
- These changes aim to reduce confusion and save time for developers when debugging key format errors.


0xDC0DE:
Hi, @nathanielkohl83 , thanks for contributing! This looks like a reasonable change.

You still need to:
- sign the CLA
- Address some failing rule lints

Happy to help you if you run into any problems.

0xDC0DE:
Closing this PR since it seems stale.

PR #3560, opened 2025-02-11
Add empty-with-boolean-expression for php

Sjord:
Detects when:

    empty($foo) && $bar

is accidentally written as

    empty($foo && $bar)

### DIFF
--- a/php/lang/correctness/empty-with-boolean-expression.yaml
+++ b/php/lang/correctness/empty-with-boolean-expression.yaml
@@ -1,0 +1,16 @@
+rules:
+- id: empty-with-boolean-expression
+  pattern-either:
+  - pattern: |
+      empty($A && $B)
+  - pattern: |
+      empty($A || $B)
+  message: >-
+    Calling `empty` on a boolean expression may be an indication that a parenthesis is misplaced.
+  metadata:
+    category: correctness
+    technology:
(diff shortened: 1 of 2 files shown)


Sjord:
Inspired by [Standalone - Fix syntax error typo in conditional by colemanw · Pull Request #32041 · civicrm/civicrm-core](https://github.com/civicrm/civicrm-core/pull/32041)

php/lang/correctness/empty-with-boolean-expression.yaml
```diff
@@ -0,0 +1,16 @@
...
+      empty($A || $B)
+  message: >-
+    Calling `empty` on a boolean expression may be an indication that a parenthesis is misplaced.
+  metadata:
+    category: correctness
+    technology:
+    - php
+    confidence: LOW
+  languages: [php]
+  severity: WARNING
```
p4p3r, line 16: Maybe worth lowering severity to `INFO`?

Sjord: I didn't give much thought to what the severity should be, but this rule does seem to reliably find actual bugs, in my experience, so I think `WARNING` is accurate.

p4p3r (approved):
Lgtm

PR #3565, opened 2025-02-18
Tweak javascript.browser.security.open-redirect for perf

IagoAbal:
Avoiding those `pattern: $PROP` makes the rule a bit faster overall, and it should help the rule scale better on extreme cases.


0xDC0DE (approved):
Thanks, Iago!

PR #3572, opened 2025-03-01
Generically check for loose equal check on hexadecimal strings

Sjord:
Instead of only checking for comparisons containing `md5` or `md5_file`, make a taint mode rule that checks comparisons for all hexadecimal strings.

This PR changes the `md5-loose-equality` rule. However, after the change that rule name is not accurate anymore. Does it make sense to edit that rule, or should this be a new and separate rule?


0xDC0DE (approved):
Thanks for your contribution. This looks good to me! 🔥

PR #3575, opened 2025-03-06
update tests for the change in code-8114

ihji:


IagoAbal (approved):
We should probably update this rule TBH, this feature `"=~/http:\/\/.../"` is deprecated and it seems to be a rule that was written before const-prop was a thing in Semgrep. But your update seems fine!

python/requests/security/no-auth-over-http.py
```diff
@@ -32,6 +32,10 @@ def test3():
     good_url = "https://www.github.com"
     r = requests.get(good_url, auth=('user', 'pass'))
 
+def test4():
+    # ruleid:no-auth-over-http
+    r = requests.get(bad_url, auth=('user', 'pass'))
```
IagoAbal, line 37: Seems like this is not found: https://github.com/semgrep/semgrep-rules/actions/runs/13937946989/job/39059770849?pr=3575

> [57.78][ERROR]: test failed for rule id no-auth-over-http on target python/requests/security/no-auth-over-http.py (with deep engine) (missing findings lines 37.)

ihji: fixed by the latest pro-develop image

PR #3577, opened 2025-03-10
Optimizing Variable Initialization to Save Gas

mingbaile:
Hi, We are a research group specializing in programming languages and software engineering. We recently studied inconsistent state update vulnerabilities in Solidity and are developing a tool to detect such issues.

After testing our prototype on your repository, we found some state variables that have never been updated, suggesting they might be missing updates or could be optimized for gas efficiency. Would you find our results useful?

These modifiers save gas by embedding values directly into the bytecode instead of using storage. The key difference:
constant: Set at compile time, no runtime cost.
immutable: Assigned in the constructor, slightly more gas usage.
([More details](https://docs.soliditylang.org/en/latest/contracts.html))

Potential Optimizations: In your repository, the following state variables might need attention. If they are meant to remain unchanged, using constant or immutable can optimize gas:

[erc721-arbitrary-transferfrom.sol](https://github.com/semgrep/semgrep-rules/blob/develop/solidity/security/erc721-arbitrary-transferfrom.sol)
State variable: maxSupply, maxFree
```solidity
uint public maxSupply = 1000;
uint public maxFree = 200;
```

Thanks for your time! Does this modification seem appropriate to you?

### DIFF
--- a/solidity/security/erc721-arbitrary-transferfrom.sol
+++ b/solidity/security/erc721-arbitrary-transferfrom.sol
@@ -1498,8 +1498,8 @@ contract Distortion is ERC721A, Ownable, ReentrancyGuard {
 
 
   bool public isClaimActive = false;
-  uint public maxSupply = 1000;
-  uint public maxFree = 200;
+  uint public immutable maxSupply = 1000;
+  uint public constant maxFree = 200;
   uint256 public price = 0.05 ether;    
   address payable private immutable payoutAddress;
 


mingbaile:
I have just signed it. Thank you

inkz:
`solidity/security/erc721-arbitrary-transferfrom.sol` is a test file for SAST tool, not an actual smart contract.

PR #3578, opened 2025-03-11
Add metadata: source to metadata-schema.yaml.schm

candrews:
`metadata: source` is used in some rules to indicate the source of that rule.

Since the field is used by rule and supported by semgrep, it should be defined in the schema.

semgrep maps `metadata: source ` to SARIF `helpUri`: https://github.com/semgrep/semgrep/pull/3329


0xDC0DE:
Let me double check if we have another source of truth for this schema before we merge this. 👍

0xDC0DE (approved):
Ok, looks like this is the place to maintain this. The engine does not care about the metadata. ✅ 
Good to merge.

PR #3584, opened 2025-03-15
Improve `insecure-load-balancer-tls-version` AWS rule

reedloden:
Add support for `ELBSecurityPolicy-TLS13-1-2-Res-2021-06`, which is the most secure TLS 1.2+ policy that AWS ALBs offer. Recommend it as the preferred fix for rule failures.

`ELBSecurityPolicy-TLS13-1-2-Res-2021-06` removes the following (insecure) cipher suites as compared to `ELBSecurityPolicy-TLS13-1-2-2021-06` because they are CBC-based instead of GCM-based:

- ECDHE-ECDSA-AES128-SHA256
- ECDHE-RSA-AES128-SHA256
- ECDHE-ECDSA-AES256-SHA384
- ECDHE-RSA-AES256-SHA384

https://docs.aws.amazon.com/elasticloadbalancing/latest/application/describe-ssl-policies.html#tls-security-policies

Also, improve/clean-up related test.

### DIFF
--- a/terraform/aws/security/insecure-load-balancer-tls-version.tf
+++ b/terraform/aws/security/insecure-load-balancer-tls-version.tf
@@ -137,6 +137,48 @@ resource "aws_alb_listener" "https_fs_1_2" {
   }
 }
 
+resource "aws_alb_listener" "tls_1_3_1_2" {
+  load_balancer_arn = var.aws_lb_arn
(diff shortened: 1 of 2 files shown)


inkz (approved):

inkz:
@reedloden thank you :pray:

PR #3590, opened 2025-03-25
Fix jwt issue

LewisArdern:
With the focus-metavariable it was leading to matching strings inside non string code blocks.


p4p3r (approved):
lgtm

PR #3602, opened 2025-04-01
Reduce noise in github action run-shell-injection rule

0xDC0DE:


neuroo (approved):
🥑

Piccirello:
I know this was removed to reduce noise, but I actually think this is a valuable rule to have. We've had at least one confirmed instance where a previous step's output could have been used to perform shell injection. Storing the step's output in an environment variable before re-use would have prevented this.

PR #3606, opened 2025-04-04
Reduce express tainted-sql-string confidence

0xDC0DE:


javascript/express/security/injection/tainted-sql-string.yaml
```diff
@@ -25,11 +25,11 @@ rules:
     - vuln
     likelihood: HIGH
     impact: MEDIUM
-    confidence: MEDIUM
+    confidence: LOW
   languages:
   - javascript
   - typescript
-  severity: ERROR
+  severity: WARNING
```
p4p3r: Why reducing severity? High likelihood x Medium impact should give Error.
(the severity is independent from the confidence)

0xDC0DE: The rule generates lots of noise, the goal is to reduce the severity.

p4p3r (approved):
lgtm

PR #3611, opened 2025-04-10
fix: remove additional brace.

yosefAlsuhaibani:
The non tree-sitter based parser would be able to recover just fine while parsing this file, probably because it wouldn't report a missing brace and instead just stealthily remove it. We are in the process of switching c/cpp to only using the tree-sitter parser.


p4p3r:
@yosefAlsuhaibani can you try merging develop (or rebasing) again? The failures in rules-test-develop are related to something that got merged last week but you shouldn't be having them.

yosefAlsuhaibani:
@p4p3r just did!

p4p3r (approved):

PR #3619, opened 2025-04-23
a fix of issue3436

sarr423:
Add a filter to avoid false positive in #3436, along with update of corresponding test case.


0xDC0DE (approved):

kurt-r2c:
There is a new rule lint for the CWE for this rule - it is not in scope for these changes and I will open a PR to bring this rule up to code.

PR #3622, opened 2025-04-24
Update dockerfile.security.missing-user

kurt-r2c:
Due to improved checks both in the engine and CI, some older rules are now failing CI due to unmappable or discouraged CWE associations.

This PR updates the CWE mapping to a mappable CWE that is relevant to the rule: https://cwe.mitre.org/data/definitions/250.html


p4p3r (approved):
lgtm

PR #3629, opened 2025-05-13
Docker audit rule for `pip install` with `--extra-index-url`

### LINKED ISSUE #3032: [Rule] Dependency confusion
Consider the following snippet in [this Dockerfile](https://github.com/ofbennett/sentiment-analysis-app/blob/94362ae3e638daeec29e09065549fd4078af8a1a/api/Dockerfile#L16-L18):

```
RUN pip install model_lstm==${API_MODEL_LSTM_VERSION} \
    --index-url https://${FURY_TOKEN}@pypi.fury.io/${FURY_USERNAME}/ \
    --extra-index-url https://pypi.org/simple/
```

This installs the package `model_lstm`. It configures two sources for packages, pypi.fury.io and pypi.org. Presumably `model_lstm` is meant to be loaded from the private pypi.fury.io, since it's not in PyPI, the public package index. However, anyone can add a package called `model_lstm` to the public PyPI repository, and if the version number is high enough, it will be installed when building this docker image. So this allows an attacker to install their own code instead of the code from the private repo.

I think semgrep can find vulnerabilities like this, for example by searching for `--extra-index-url` in Dockerfiles, where the URL specifies credentials.

The above example is with Python's pip, but dependency confusion is possible with most dependency tools.

0xDC0DE:
Fixes https://github.com/semgrep/semgrep-rules/issues/3032

### DIFF
--- a/dockerfile/audit/dockerfile-pip-extra-index-url.dockerfile
+++ b/dockerfile/audit/dockerfile-pip-extra-index-url.dockerfile
@@ -1,0 +1,14 @@
+FROM python:3.10.1-alpine3.15@sha256:4be65b406f7402b5c4fd5df7173d2fd7ea3fdaa74d9c43b6ebd896197a45c448
+
+RUN pip install model_lstm==${API_MODEL_LSTM_VERSION} \
+    --index-url https://${FURY_TOKEN}@pypi.fury.io/${FURY_USERNAME}/ \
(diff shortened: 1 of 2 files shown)


LewisArdern (approved):

p4p3r (approved):
lgtm

PR #3635, opened 2025-06-02
fix(tests): query-set-extra

tean-lai:
Updated some tests that previously considered constant strings to not be constant, leading to some incorrect matches.

Also added a negative test case.


LewisArdern (approved):
lgtm gamers

PR #3649, opened 2025-07-03
[REPLACE] New Published Rules - righettod.java-pattern-matches-from-string-parameter.

righettod:
Hello,

This PR replace the PR #3648 in order to be able to fix the warning and error raised by the CI validation workflow as well as any of your feedback 😉

This rule, for java language, is intended to detect and raise a notification (INFO level) when a regular expression is build from a parameter of a method. The goal is to allow to detect such situation and perform a manual control that a validation is in place against the regex used.


righettod:
Hi,

Any news about this PR?

Thanks for your feedback.

righettod:
Hi @inkz ,

I synced the fork.

inkz (approved):

inkz:
Thank you ✅

PR #3658, opened 2025-07-14
Update semgrep-rules-test-develop.yml

p4p3r:


armchairlinguist (approved):
This appears correct to me based on https://hub.docker.com/r/semgrep/semgrep-nightly/tags?name=develop, thank you!

PR #3659, opened 2025-07-16
Fix csharp-sqli sanitizers

jrmnl:
Hi,

Currently, the `csharp-sqli` pattern-sanitizers do not work correctly.

See existing examples of sanitizer usage:
https://github.com/semgrep/semgrep-rules/blob/develop/csharp/lang/security/sqli/csharp-sqli.cs#L178-L179

Also added additional methods:
https://learn.microsoft.com/en-us/dotnet/api/system.data.sqlclient.sqlparametercollection.addrange
https://learn.microsoft.com/en-us/dotnet/api/system.data.sqlclient.sqlparametercollection.addwithvalue

before fix
1_before

after fix
2_after

### DIFF
--- a/csharp/lang/security/sqli/csharp-sqli.yaml
+++ b/csharp/lang/security/sqli/csharp-sqli.yaml
@@ -28,9 +28,13 @@ rules:
     pattern-sanitizers:
       - pattern-either:
           - pattern: |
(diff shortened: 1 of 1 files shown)


kurt-r2c (approved):

kurt-r2c:
LGTM

PR #3670, opened 2025-08-05
Fix typo in express-jwt-not-revoked.yaml `asvs: control_id:`

myersg86:
Simple typofix: `Insecue` > `Insecure`


0xDC0DE (approved):
Awesome, good find and thanks for the fix!

PR #3676, opened 2025-08-18
Use different scriptlet JSP to test no-scriptlets rule

### LINKED ISSUE #3673: Microsoft Defender triggers on JSP webshell used in test
https://github.com/semgrep/semgrep-rules/blob/develop/java/lang/security/audit/xss/jsp/no-scriptlets.jsp

no-scriptlets.jsp contains a webshell, and Microsoft Defender gets angry about that. This is of course not really a bug in the semgrep rules, but still mildly annoying when working on these rules.

Perhaps a better alternative would be something like https://github.com/nirmaldhara/Jsp-hello-world/blob/master/WebContent/index.jsp, which also contains various variations on `<%` syntax.

Sjord:
Fixes #3673

Change test from webshell to hello world, because
- it tests declarations, scriptlets and expressions,
- it is not a webshell which triggers virus scanners.


Sjord:
@kurt-r2c Could you take a look at this PR?

Sjord:
This is ready for review.

kurt-r2c (approved):

PR #3680, opened 2025-09-02
Fix false positive for unquoted-attribute-var

Sjord:
If the expression itself contains the same quotes as that surround it, this rule would give a false positive.


Sjord:
@minusworld Could you take a look at this PR?

Sjord:
Can this be reviewed please?

0xDC0DE (approved):
Thanks for your contribution.

PR #3706, opened 2025-10-24
New Rules Proposal: Detect exposure to log injection in java.

righettod:
Hello,

This rule, for java language, is intended to detect and raise a warning when an unsanitized String method parameter is used as an argument to a logger call. The goal is to allow to detect such situation and perform a manual control that a validation is in place.

💡 To limit the false positives, I updated the rule to only trigger for methods that are exposed as a "web service". I added annotations used by Spring Web, JAX-RS and JAX-WS frameworks.

image

### DIFF
--- a/java/lang/security/audit/java-logging-from-string-parameter.java
+++ b/java/lang/security/audit/java-logging-from-string-parameter.java
@@ -1,0 +1,37 @@
+package com.test;
+
+import org.slf4j.LoggerFactory;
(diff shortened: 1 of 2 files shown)


righettod:
Hi,
Any news about this PR?
Thanks for your feedback 😉

righettod:
Hi,
Any news about this PR?
Thanks for your feedback 😉

PR #3707, opened 2025-10-30
New Rules Proposal: Detect usage of SHA1PRNG in java. 

righettod:
Hello,

This rule, for java language, is intended to detect and raise a warning when **SHA1PRNG**, a pseudo random number generator algorithm which is considered insecure, is used.

image

image


righettod:
Hi,
Any news about this PR?
Thanks for your feedback 😉

righettod:
Hi,
Any news about this PR?
Thanks for your feedback 😉

PR #3708, opened 2025-10-31
New Rules Proposal: Detect usage in java of an XML canonicalization method that excludes XML comments.

righettod:
Hello,

This rule, for java language, is intended to detect and inform when an XML canonicalization method that excludes XML comments from the computation of Digest and Signature is used.

image


righettod:
Hi,
Any news about this PR?
Thanks for your feedback 😉

righettod:
Hi,
Any news about this PR?
Thanks for your feedback 😉

PR #3744, opened 2026-01-18
New Published Rules - preprocess SSTI in thymeleaf

EvtDanya:
Hello!

This PR introduces a new semgrep rule detecting dangerous Thymeleaf expressions that use the preprocessing marker `__${...}__`

This rule is based on security researches:
- https://modzero.com/en/blog/spring_boot_ssti/
- https://www.acunetix.com/blog/web-security-zone/exploiting-ssti-in-thymeleaf/ 

image

image

### DIFF
--- a/java/spring/security/injection/thymeleaf-preprocess-ssti.html
+++ b/java/spring/security/injection/thymeleaf-preprocess-ssti.html
@@ -5,10 +5,11 @@
     <title>SSTI DEMO</title>
   </head>
   <body>
-    <!-- True positive: preprocessing pattern with SSTI -->
+    <!-- ruleid: spring-thymeleaf-preprocess-ssti -->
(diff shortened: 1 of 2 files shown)


EvtDanya:
I added support of another types of expressions 

```
To attempt an SSTI in Thymeleaf, we first must understand expressions that appear in Thymeleaf attributes. Thymeleaf expressions can have the following types:

${...}: Variable expressions – in practice, these are OGNL or Spring EL expressions.
*{...}: Selection expressions – similar to variable expressions but used for specific purposes.
#{...}: Message (i18n) expressions – used for internationalization.
@{...}: Link (URL) expressions – used to set correct URLs/paths in the application.
~{...}: Fragment expressions – they let you reuse parts of templates.
```

PR #3748, opened 2026-02-11
New Rules Proposal: Detect usage of RemoteIPHeader directive in an apache configuration.

righettod:
Hello,

This generic rule for apache is intended to detect when the directive [RemoteIPHeader](https://httpd.apache.org/docs/current/mod/mod_remoteip.html#remoteipheader) is used. The goal is for inform about the potential risks of unexpected DNS queries if the value of the header is not checked via for example a [rewrite rule](https://httpd.apache.org/docs/current/mod/mod_rewrite.html).

image


righettod:
Hi,
Any news about this PR?
Thanks for your feedback 😉

PR #3749, opened 2026-02-11
Update warning message in unsafe-deserialization-interface.yaml

inkz:
Removed reference to github.com/ravisastryk/go-safeinput for automatic protection in the warning message about unsafe deserialization.


Fidget-Grep (approved):
I'm probably not qualified to approve this but it all makes sense to me.

PR #3756, opened 2026-02-27
fix: Remove duplicate names from test files

jmgrosen:
Several Java and Go test files for rules had duplicate names where they weren't actually allowed by the language. This caused issues for naming in Semgrep, which assumes that this should be impossible.

This commit replaces those duplicate names with distinct names, or adds parameters to make method signatures non-overlapping.


neuroo (approved):
Thank you @jmgrosen

PR #3761, opened 2026-03-04
Detect shell injection inside GitHub 'script' step

Piccirello:
When used with e.g. `actions/github-script`, `script:` is effectively equivalent to `run:`.


Piccirello:
@0xDC0DE @inkz given you've both reviewed previous PRs of mine, would you mind taking a look at this?

Piccirello:
Could someone take a look at this, or share some info on why this won't be moving forward?

inkz:
@Piccirello there is a separate rule for `actions/github-script` case: https://github.com/semgrep/semgrep-rules/blob/develop/yaml/github-actions/security/github-script-injection.yaml

PR #3765, opened 2026-03-13
fix: Remove duplicate function names from test files

jmgrosen:
Disambiguate duplicate function/method names that lead to odd shadowing of results. All the actual matches are preserved.


neuroo (approved):
Thanks again @jmgrosen

PR #3769, opened 2026-03-17
fix(c): improve insecure-use-strtok-fn message

MarkLee131:
The current message says strtok "permanently erases the delimiter character", which is strtok's normal documented behavior, not a security issue.

So we updated the message to describe the **actual risks**:                                                                                 
  - Non-thread-safe internal static state (concurrent/nested calls corrupt parsing)
  - Input buffer mutation (secondary, but worth noting)                                                                                 
  - Recommends both `strtok_r()` (POSIX) and `strtok_s()` (C11) as alternatives                                                         

Found while scanning [Vim](https://github.com/vim/vim) (~400k LOC) with `--config auto` — all 21 strtok findings were false positives. The inaccurate message made triage harder since the stated risk wasn't the real concern.

### DIFF
--- a/c/lang/security/insecure-use-strtok-fn.c
+++ b/c/lang/security/insecure-use-strtok-fn.c
@@ -1,4 +1,5 @@
 #include <stdio.h>
+#include <string.h>
 
 int DST_BUFFER_SIZE = 120;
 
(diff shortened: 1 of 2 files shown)


MarkLee131:
Bump on this... 
Is anyone able to take a look? Just a message-text change, no rule logic touched.                                                                                                                                           
@LewisArdern @0xDC0DE if either of you has a moment. :)

PR #3771, opened 2026-03-17
fix(c): reduce false positives in double-free and use-after-free rules

### LINKED ISSUE #3770: C memory safety rules (double-free, use-after-free) missing common exclusion patterns cause false positives
**Describe the bug**
Three C memory safety rules produce false positives on common, safe C patterns:

  1. **`double-free`**: Does not recognize `return` between two `free()` calls as separating mutually exclusive paths. Also only
  excludes `malloc()` reallocation but misses `calloc()`, `realloc()`, and `strdup()`.

  2. **`use-after-free`** and **`function-use-after-free`**: Same missing `calloc`/`realloc`/`strdup` exclusions.

**To Reproduce**

```c
  // False positive: double-free flags this, but return makes paths mutually exclusive
  int cleanup(int error) {
      char *buf = malloc(100);
      if (!error) {
          free(buf);
          return 0;
      }
      free(buf);  // <- flagged as double-free
      return -1;
  }

  // False positive: strdup reallocation not recognized
  char *var = strdup("hello");
  free(var);
  var = strdup("world");
  free(var);  // <- flagged as double-free

  // Same issue with calloc/realloc for use-after-free rules
  char *p = malloc(100);
  free(p);
  p = calloc(1, 200);
  p[0] = 'x';  // <- flagged as use-after-free
```

  Run`semgrep --config r/c.lang.security.double-free (or use-after-free / function-use-after-free) `
on the above code.

**Expected behavior**

These patterns should not be flagged:
  - Two free() calls separated by a return are mutually exclusive cleanup paths, not double-free
  - calloc/realloc/strdup between free and reuse is safe reallocation, same as malloc (which is already excluded)

**Priority**

**Desktop (please complete the following information):**
  - OS: macOS (Darwin 25.3.0)
  - CLI tool
  - Version: semgrep 1.132.0
  
**Additional context**

  Found while scanning Vim (~400k lines of C) with semgrep --config auto. The double-free rule's return gap caused false positives on
  Vim's common goto-based error handling pattern where success and error paths each call free() but are separated by return. The missing
   calloc/realloc/strdup exclusions affected all three rules.

MarkLee131:
This PR fixes https://github.com/semgrep/semgrep-rules/issues/3770,  adding the missing exclusions to three C memory safety rules: 
  - **double-free**: Add `return` exclusion between two `free()` calls (mutually exclusive cleanup paths), and recognize
  `calloc`/`realloc`/`strdup` as reallocation (previously only `malloc`)
  - **use-after-free**: Add `calloc`/`realloc`/`strdup` as recognized reallocation patterns
  - **function-use-after-free**: Same `calloc`/`realloc`/`strdup` additions

---
  -  `semgrep --test c/lang/security/` — 12/12 pass
  - All existing test cases still match; new `ok:` cases added for each exclusion


dannytheway:
I have accidentally opened a PR with the same fixes, I can close it if yours gets merged, but can I suggest a few changes?

Detection for functions other than malloc can be implemented more elegantly as:
```
- pattern-not:
          patterns:
            - pattern-inside: |
                free($VAR);
                ...
                $VAR = $ALLOC(...);
                ...
            - metavariable-pattern:
                metavariable: $ALLOC
                pattern-either:
                  - pattern: malloc
                  - pattern: calloc
                  - pattern: strdup
```
rather than a series of `- pattern-not-inside` and it should work fine.

Also the original rule will not flag `*ptr=something;` as vulnerable, so I suggest to add a `- pattern: (*$VAR)`.

PR  is #3795, check [here](https://github.com/semgrep/semgrep-rules/pull/3795/changes/9acda378fe7f18e6821df9cce0f80f24efde9846) for details.

MarkLee131:
Hi @dannytheway, adopted both in 0dab1b83. The `$ALLOC` refactor cuts ~15 duplicated lines per rule, threw in `reallocarray` and `strndup` to the list while at it. `(*$VAR)` is in the access patterns with new tests for write/read deref. Tests still 12/12 on `semgrep --test c/lang/security/`.

PR #3780, opened 2026-03-25
feat(netsuite): add best practices SAST governance rules

joshOrigami:
### **Description:**

This PR introduces the first foundational SAST ruleset for **Oracle NetSuite (SuiteScript)** to the Semgrep registry. 

NetSuite runs on a proprietary JavaScript API (SuiteScript 2.x) executed via a GraalVM backend. Because it is a multi-tenant cloud ERP, it enforces strict "Governance Limits" (usage units) on database operations. Standard JavaScript linting cannot catch these platform-specific governance traps, which often lead to fatal `SSS_USAGE_LIMIT_EXCEEDED` errors in production. 

With the rapid adoption of AI-generated code (Copilot, Claude), this problem is escalating. AI models frequently write naive "Happy Path" JavaScript that works syntactically but violently violates NetSuite's database governance. This ruleset acts as a gatekeeper to prevent architectural anti-patterns from reaching production.

**Rules Included in this PR:**
1. `netsuite-no-record-load-save-in-loop`: 
   * **The Risk:** Loading or saving a NetSuite record inside a `for/while/for each` loop creates an $O(n^2)$ performance bottleneck and rapidly burns through script governance units.
   * **The Fix:** Forces developers to rethink the architecture and use bulk-fetch mechanisms like `N/search` or `N/query` outside the loop.
2. `netsuite-unhandled-aftersubmit-try-catch`: 
   * **The Risk:** NetSuite User Event `afterSubmit` triggers execute *after* the 'database' commit but before the user's UI resolves. An unhandled exception here can cause data inconsistency or block the user from saving the record entirely.
   * **The Fix:** Enforces a root-level `try/catch` block for mission-critical entry points.
3. `netsuite-no-console-log`: 
   * **The Risk:** Using standard browser `console.log` in backend SuiteScript is an audit and execution risk. 
   * **The Fix:** Enforces the use of NetSuite's native telemetry module (`N/log`).

**Testing:**
* Added comprehensive `_test.js` files covering `ruleid` hits and `ok` passes for all patterns.
* Passed local `semgrep --validate` checks.

**About the Author:**
My name is Joshua Meiri, and I specialize in AI-Native NetSuite Engineering. I am the creator of the **ABCD (Always Be Continuously Deliberate)** methodology, which transitions NetSuite teams from legacy, UI-based scripting into modern, testable, MVC-driven DevOps lifecycles. I have over 20 years of experience in CRM and ERP systems and the creator of [origami lens](http://origamilens.com/) of NetSuite metadata exploration. 

I built these rules because the ecosystem currently lacks a standardized, automated defense against AI-generated hallucinations and legacy messy code. By contributing this to the official registry, I hope to establish a true "financial primitive" of platform-wide safety for NetSuite developers globally. 

More rules to come!

### DIFF
--- a/javascript/contributions/netsuite/best-practices.yaml
+++ b/javascript/contributions/netsuite/best-practices.yaml
@@ -121,26 +121,6 @@ rules:
           - pattern: $SEARCH.create(...)
           - pattern: $QUERY.run(...)
 
-  - id: netsuite-query-injection
-    languages: [javascript, typescript]
-    severity: WARNING
-    metadata:
-      technology: [netsuite]
-      type: Security
-      category: Injection
-      author: Joshua Meiri (Origami Precision, LLC)
-      references: [https://www.origamiprecision.com/]
-      origami-rationale: Concatenating variables into query strings can lead to SuiteQL injection. Using parameter binding with `params` ensures that variables are safely escaped, preventing malicious data extraction or manipulation.
-    message: "**Injection Risk:** Potential SuiteQL injection via string concatenation. Use the `params` property in `query.runSuiteQL()` to safely bind variables."
-    patterns:
-      - pattern-either:
-          - pattern: |
-              query.runSuiteQL({ query: ... + $VAR + ..., ... })
-          - pattern: |
-              query.runSuiteQL({ query: ... + $VAR, ... })
-          - pattern: |
-              query.runSuiteQL({ query: $VAR + ..., ... })
-
   - id: netsuite-enforce-sri-cdn
     languages: [javascript, typescript, html]
     severity: WARNING
--- a/javascript/contributions/netsuite/best-practices.yaml
+++ b/javascript/contributions/netsuite/best-practices.yaml
@@ -13,6 +13,7 @@ rules:
       risk: Console
       license: commons
       author: Joshua Meiri (Origami Precision, LLC)
+      origami-rationale: console.log is transient and doesn't write to the Execution Log. Using it in production leaves zero audit trail for troubleshooting.
       technology:
         - netsuite
       references:
@@ -33,6 +34,7 @@ rules:
       risk: Blow-up
       license: commons
       author: Joshua Meiri (Origami Precision, LLC)
+      origami-rationale: Standard O(n) operations become O(n^2) governance killers when inside loops.
       references:
         - https://www.origamiprecision.com/
       technology:
@@ -60,6 +62,7 @@ rules:
       risk: Try/Catch
       license: commons
       author: Joshua Meiri (Origami Precision, LLC)
+      origami-rationale: Prevents silent crashes that leave integrations or related records in an inconsistent state.
       technology:
         - netsuite
       source_rule_url:
@@ -69,6 +72,7 @@ rules:
     message: "**NetSuite Reliability Risk:** Root try-catch missing in afterSubmit. 
              While NetSuite handles uncaught exceptions, explicit error handling allows for better logging, notification of admins, and graceful recovery of side-effects
              (like integrations or related record updates) that occur post-save."
+    # Removed $(SOMETHING) at the end, makes it trully function level 
     patterns:
       - pattern-inside: |
           function afterSubmit($CONTEXT) {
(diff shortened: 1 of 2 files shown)


javascript/contributions/netsuite/best-practices.yaml
```diff
@@ -0,0 +1,85 @@
...
+      license: commons
+      author: Joshua Meiri (Origami Precision, LLC)
+      technology:
+        - netsuite
+      source_rule_url:
+        - https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/try...catch
+      references:
+        - https://www.origamiprecision.com/
+    message: "**NetSuite Reliability Risk:** afterSubmit function lacks a root
+      `try/catch` block. Unhandled errors here will block the record from
```
stephen-lemp: The save will still occur, though the error will be raised to the user without a try-catch. I'm not sure that failing silently is a best practice here. NetSuite already wraps afterSubmit and like functions with a try-catch to ensure it doesn't crash the server, and they properly raise the error to the user.

javascript/contributions/netsuite/best-practices.yaml
```diff
@@ -0,0 +1,85 @@
+rules:
+  - id: netsuite-no-console-log
```
stephen-lemp, line 2: The no-console-log rule needs to be scoped to server-side scripts only.
It should be acceptable in client scripts.

javascript/contributions/netsuite/best-practices.yaml
```diff
@@ -0,0 +1,85 @@
...
+      references:
+        - https://www.origamiprecision.com/
+
+  - id: netsuite-no-record-load-save-in-loop
+    languages:
+      - javascript
+      - typescript
+    message: >
+      **NetSuite Governance Trap:** Database call (`load/save`) detected inside
+      a loop. This leads to O(n^2) performance and will cause
```
stephen-lemp: This will not always cause SSS_USAGE_LIMIT_EXCEEDED, particularly if user includes a limit. For example, in a Suitelet the user might check number of records to load/save. If the Suitelet can cover it with 1000 governance units, it would process, otherwise it would offload to Map/Reduce. This is a valid pattern.

javascript/contributions/netsuite/best-practices.yaml
```diff
@@ -0,0 +1,85 @@
+rules:
+  - id: netsuite-no-console-log
+    languages:
+      - javascript
+    severity: ERROR
```
stephen-lemp: I'd recommend considering this a WARNING/Medium severity

stephen-lemp (changes requested):
Some review from a fellow SuiteScript developer. Thanks for the contribution here @joshOrigami!

javascript/contributions/netsuite/best-practices.yaml
joshOrigami: @stephen-lemp on this one I disagree, as releasing (client side) code into Production where you are potentially logging "stuff" to console is an Anti-Pattern. Agree this should be a warning.

javascript/contributions/netsuite/best-practices.yaml
joshOrigami: agree, will change.

javascript/contributions/netsuite/best-practices.yaml
joshOrigami: Great catch, Stephen. You’re spot on regarding the post-commit nature of afterSubmit—the save is done, and NetSuite’s native handler does prevent a total server-hang.

My intent with this rule isn't to encourage 'silent failures' (swallowing errors), but to enforce deliberate handling. I see too many scripts that crash and leave external systems (integrations, custom records) in an inconsistent state because the dev relied on the generic NS error page instead of a structured catch that logs proper context or triggers a fallback.

I’ll refine the rule message to emphasize 'Structured Error Handling' rather than just 'Avoid Crashing.' Appreciate the sharp eye!

javascript/contributions/netsuite/best-practices.yaml
joshOrigami: Spot on, Stephen. The 'Suitelet-to-M/R offload' is a classic high-perf pattern. I’m updating the message to acknowledge that usage-limit checks and offloading are the valid exceptions here.

My goal with the initial ruleset is to catch the 90% of cases where someone is just 'loop-loading' without a safety net. I’ll refine the severity and the wording so it reads more as a performance audit than a hard block. Thanks for keeping the edge cases in mind!

+ will add '- pattern-not-inside: ' for map/reduce usage.

PR #3791, opened 2026-03-27
Add per-language supply chain attack prevention via default cooldowns

pid1:
Require cooldowns or minimum release ages before updating to newly published package versions.

This covers:

- Dependabot
- Renovate
- pnpm
- uv
- bun

Additionally, detect unpinned dependencies in Python scripts using uv’s inline script metadata format.


generic/uv/security/uv-pyproject-missing-dependency-cooldown.yaml
```diff
@@ -0,0 +1,43 @@
+rules:
+  - id: uv-pyproject-missing-dependency-cooldown
+    patterns:
+      - pattern-regex: (?s)(\[tool\.uv\].*)
+      - pattern-not-regex: exclude-newer
```
leifdreizler, line 5: won't this miss instances where there is no `[tool.uv]` block?

pid1: I was intentionally scoping this to where uv was in use.

leifdreizler: I don't think `[tool.uv]` is required

generic/dependabot/security/dependabot-missing-cooldown.yaml
```diff
@@ -0,0 +1,48 @@
+rules:
+  - id: dependabot-missing-cooldown
+    patterns:
+      - pattern-regex: (?s)(version\s*:.*)
+      - pattern-not-regex: cooldown
```
gbennett-squarespace: This is going to match the whole file if we find cooldown, which doesn't work if you have multiple update entries (for example both NPM and PyPi).  If one has cooldown and the other doesn't, then it'll be a problem.

Using yaml mode instead of generic is necessary (otherwise it doesn't understanding the nesting properly), and updated the pattern to:
```
    patterns:
      - pattern-inside: |
          updates:
            ...
      - pattern: |
          - package-ecosystem: $ECOSYSTEM
            ...
      - pattern-not: |
          - package-ecosystem: $ECOSYSTEM
            ...
            cooldown:
              ...
            ...
```

If you want a test file:
```
version: 2
updates:
  # ok: dependabot-missing-cooldown
  - package-ecosystem: "npm"
    directory: "/"
    schedule:
      interval: "weekly"
    cooldown:
      default-days: 7

---
version: 2
updates:
  # ok: dependabot-missing-cooldown
  - package-ecosystem: "pip"
    cooldown:
      default-days: 7
    directory: "/"
    schedule:
      interval: "weekly"
  # ok: dependabot-missing-cooldown
  - package-ecosystem: "docker"
    directory: "/"
    groups:
      dependencies:
        patterns:
          - "*"
    schedule:
      interval: "weekly"
    cooldown:
      default-days: 7

---
version: 2
updates:
  # ok: dependabot-missing-cooldown
  - package-ecosystem: "pip"
    cooldown:
      default-days: 7
    directory: "/"
    schedule:
      interval: "weekly"
  # ruleid: dependabot-missing-cooldown
  - package-ecosystem: "docker"
    directory: "/"
    schedule:
      interval: "weekly"

---
version: 2
updates:
  # ruleid: dependabot-missing-cooldown
  - package-ecosystem: "pip"
    directory: "/"
    schedule:
      interval: "weekly"

---
version: 2
updates:
  # ok: dependabot-missing-cooldown
  - package-ecosystem: "pip"
    cooldown:
      default-days: 7
    directory: "/"
    schedule:
      interval: "weekly"

---
version: 2
updates:
  # ok: dependabot-missing-cooldown
  - package-ecosystem: "pip"
    directory: "/"
    cooldown:
      default-days: 7
    schedule:
      interval: "weekly"

---
version: 2
updates:
  # ok: dependabot-missing-cooldown
  - package-ecosystem: "pip"
    directory: "/"
    schedule:
      interval: "weekly"
    cooldown:
      default-days: 7
```

I think you'll need to name your test file `dependabot-missing-cooldown.test.yaml` to get it to work correctly.

diptendur2c:
pnpm minimumReleaseAge was introduced in v10, would it not be FP for < 10 ? May be we can edit the message?

generic/uv/security/uv-pyproject-missing-dependency-cooldown.yaml
0xDC0DE: Wouldn't the current rule flag on all python projects, even those where `uv` is not used? The message and id indicate that the rule is looking for cases where `uv` is used.

0xDC0DE:
You currently don't have a lot of `ok` testcases. We usually expect at least one for each rule, could you take a look at adding these?

0xDC0DE:
Overall, very useful contribution and well authored rules. Metadata and messages are great, path scoping is good.
🥇 Thanks for contributing, let's see if we can resolve the open comments and merge.

generic/bun/security/bunfig-missing-minimum-release-age.yaml
```diff
@@ -0,0 +1,43 @@
+rules:
+  - id: bunfig-missing-minimum-release-age
+    patterns:
+      - pattern-regex: (?ms)(^[\[a-zA-Z].*)
+      - pattern-not-regex: minimumReleaseAge
+    message: >-
```
0xDC0DE:
```suggestion
    message: |
```

Multiline messages are displayed better in CLI output when this operator is used, can you update the rules to use this instead?

0xDC0DE (changes requested):

pid1:
@diptendur2c That was explicitly removed as per [this feedback](https://github.com/semgrep/semgrep-rules/pull/3791#discussion_r3012842042).

pid1:
@diptendur2c updated the `pnpm` phrasing and @0xDC0DE addressed your feedback RE `ok` cases.

0xDC0DE:
Thanks for this contribution, Jonathan. You were ahead of the curve in identifying cooldown rules as important, and the metadata and messaging across your rules are well done.

Since your PR was opened, however, we received an overlapping PR in #3805 that covers much of the same ground. After comparing both, I've decided to move forward with #3805 as the base because it uses language-aware parsers (`yaml`, `json`) instead of `generic` for structured config files, and validates that cooldown values meet a minimum threshold rather than only checking presence/absence (e.g., `minimumReleaseAge: 0` would pass the generic + `pattern-not-regex` approach but is caught by the `metavariable-comparison` approach).

That said, your PR has two rules that #3805 does not cover:

`npmrc-missing-min-release-age` 
`uv-script-unpinned-dependency`
We'd like to incorporate both of these. Would you be open to submitting them as a separate, smaller PR once #3805 is merged? You can ping me for review.

Thank you again for the contribution and for driving the discussion on this topic.

pid1:
Closing in favor of a smaller, focused follow-up PR per @0xDC0DE's request: #3861.

Summary of what happened with this PR:

- #3805 was merged and covers six of the eight rules originally proposed here with stronger implementations (language-aware parsers, threshold validation).
- The two rules @0xDC0DE flagged as still wanted were `uv-script-unpinned-dependency` and `npmrc-missing-min-release-age`.
- `uv-script-unpinned-dependency` is being submitted as #3861, with the suggested `message: |` style applied.
- `npmrc-missing-min-release-age` is being dropped: #3805's `npm-missing-minimum-release-age` already targets `.npmrc` and additionally validates the threshold value, so the rule from this PR would only weaken coverage if added.

Thanks again @0xDC0DE for the careful review and the path forward.

PR #3797, opened 2026-03-30
fix(kotlin): exclude ephemeral port detection from unencrypted-socket rule

0xDC0DE:
## Summary

Fixes `kotlin.lang.security.unencrypted-socket.unencrypted-socket` false positives on ephemeral port detection patterns.

**Problem**: The rule flags `ServerSocket(0)` used in port-detection helpers (`ServerSocket(0)` → `localPort` → `close()`), which is a common benign pattern that does not transmit any cleartext data. Reported by Airbnb.

**Root cause**: The rule uses a bare `pattern-either` matching any `ServerSocket(...)` or `Socket(...)` construction without considering usage context.

**Fix**: Restructured the rule to use `patterns` with two `pattern-not-inside` clauses that exclude `ServerSocket(0)` inside functions returning `Int` that close the socket — the canonical shape of port-detection helpers. `ServerSocket(0)` used with `accept()` or in non-`Int`-returning functions is still flagged.

## Test plan

- [x] `semgrep --validate` passes
- [x] `semgrep --test` passes (13/13 tests)
- [x] New test cases cover:
  - `ok:` ServerSocket(0) → localPort → close() in Int-returning function
  - `ok:` Same pattern with different ordering of intermediate statements
  - `ruleid:` ServerSocket(0) followed by accept() (true positive retained)
- [x] All existing test cases continue to pass (no regressions)

Closes SRC-12442

Made with [Cursor](https://cursor.com)

### DIFF
--- a/kotlin/lang/security/unencrypted-socket.kt
+++ b/kotlin/lang/security/unencrypted-socket.kt
@@ -72,3 +72,32 @@ public class UnencryptedServerSocket {
     }
 
 }
+
+public class EphemeralPortDetection {
+
+    fun getOpenPort(): Int {
+        // ok: unencrypted-socket
+        val ss = ServerSocket(0)
+        val port = ss.localPort
+        ss.reuseAddress = true
+        ss.close()
+        return port
+    }
+
+    fun getAvailablePort(): Int {
+        // ok: unencrypted-socket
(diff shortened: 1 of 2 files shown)


p4p3r (approved):
lgtm

PR #3798, opened 2026-03-30
fix(java,kotlin): use case-insensitive matching for crypto algorithm names

0xDC0DE:
## Summary

- Java's cryptographic API treats standard algorithm names as case-insensitive per [Oracle's JCA documentation](https://docs.oracle.com/en/java/javase/17/docs/specs/security/standard-names.html). Rules using exact string matching or case-sensitive regex patterns missed lowercase/mixed-case variants, causing false negatives.
- Updated 18 rules across Java and Kotlin to use `(?i)` case-insensitive matching via `metavariable-regex` or inline regex patterns.
- Added mixed-case test cases to all affected rules to verify detection of case variants.

### Rules updated

**Java** (14 rules):
- `use-of-rc4`, `use-of-rc2`, `use-of-blowfish` — converted exact string patterns to `metavariable-regex` with `(?i)`
- `use-of-aes-ecb` — converted inline regex to `metavariable-regex` with `(?i)`
- `des-is-deprecated`, `desede-is-deprecated` — added `(?i)` to inline regex patterns
- `use-of-sha1`, `ecb-cipher` — added `(?i)` to existing `metavariable-regex`
- `blowfish-insufficient-key-size`, `weak-rsa` — converted exact strings to `metavariable-regex` with `(?i)`
- `md5-used-as-password` — added `(?i)` to inline regex in taint source
- `use-of-default-aes` — converted all `"AES"` literals to `(?i)` inline regex
- `cbc-padding-oracle` — added `(?i)` to inline regex
- `rsa-no-padding` — added `(?i)` and simplified manual `[Nn][Oo][Nn][Ee]` to `None`

**Kotlin** (4 rules):
- `use-of-md5` — converted exact `"MD5"` to `(?i)` inline regex
- `use-of-sha1`, `ecb-cipher` — added `(?i)` to `metavariable-regex`
- `weak-rsa` — converted exact `"RSA"` to `(?i)` inline regex

## Test plan

- [x] `semgrep --test` passes for all 18 modified rules
- [x] `semgrep --validate` passes for all modified rules (40 rules validated)
- [x] New mixed-case test cases added to all affected rules
- [x] Existing test cases continue to pass (no regressions)
- [x] Fix tests pass for `des-is-deprecated` and `cbc-padding-oracle`

Ref: https://github.com/semgrep/semgrep/issues/11408

Closes SRC-12443

Made with [Cursor](https://cursor.com)

### DIFF
--- a/java/lang/security/audit/blowfish-insufficient-key-size.java
+++ b/java/lang/security/audit/blowfish-insufficient-key-size.java
@@ -17,6 +17,12 @@ public class Cls {
         keyGen.init(128);
     }
 
+    public void unsafeKeySize_lowercase() {
+        // ruleid: blowfish-insufficient-key-size
+        KeyGenerator keyGen = KeyGenerator.getInstance("blowfish");
+        keyGen.init(64);
+    }
+
     public void superSafeKeySize() {
         // ok: blowfish-insufficient-key-size
         KeyGenerator keyGen = KeyGenerator.getInstance("Blowfish");
--- a/java/lang/security/audit/blowfish-insufficient-key-size.yaml
+++ b/java/lang/security/audit/blowfish-insufficient-key-size.yaml
@@ -31,9 +31,12 @@ rules:
   - java
   patterns:
   - pattern: |
-      $KEYGEN = KeyGenerator.getInstance("Blowfish");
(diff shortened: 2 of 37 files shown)


p4p3r (approved):
lgtm

PR #3849, opened 2026-05-20
Add oauth-state-unsigned-base64 rule for unsigned OAuth state forgery

nedu-m:
## Summary

Adds a new rule under `javascript/lang/security/` that detects OAuth `state` parameters being deserialized from base64-encoded JSON without an adjacent HMAC verification.

When an OAuth flow uses `state` to round-trip identifiers (such as `organization_id` or `tenant_id`) across the authorization redirect, the `state` must be integrity-protected. Otherwise an attacker who knows or guesses a victim's identifiers can construct a `state` value from scratch and inject it into the callback — binding their own third-party account into the victim's tenant.

This pattern is most commonly observed in Next.js / Node SaaS apps that hand-roll OAuth handlers (rather than going through `passport` / `next-auth` / `iron-session`, which sign `state` by default).

## Real-world context

This exact pattern was identified in a 2026 vulnerability disclosure against an AI SEO SaaS where:

1. `/api/integrations/notion/auth` accepted an `organization_id` in the body and returned a Notion OAuth URL with `state = base64(JSON.stringify({ organization_id, ... }))` — no HMAC.
2. `/api/integrations/notion/callback` decoded `state` directly via `JSON.parse(Buffer.from(state, "base64").toString())` and used the embedded `organization_id` to persist a Notion access token — with no session check and no signature verification.

Result: an unauthenticated attacker could bind their own Notion workspace into any victim org by constructing the `state` themselves.

Full writeup: https://shippedwithbugs.com/writeups/outrank

## Detection

The rule matches:

- `JSON.parse(Buffer.from($STATE, "base64").toString(...))` (Node)
- `JSON.parse(atob($STATE))` (browser / edge runtime)

and is suppressed when the enclosing scope contains a sibling `createHmac(...)` or `timingSafeEqual(...)` call — i.e. the developer is verifying the signature before deserializing.

## Files

- `javascript/lang/security/oauth-state-unsigned-base64.yaml` — the rule
- `javascript/lang/security/oauth-state-unsigned-base64.js` — two vulnerable cases and one safe case

## Notes

- Severity set to `ERROR` because exploitation is unauthenticated and the consequence is cross-tenant account binding.
- Placed under `javascript/lang/security/` since the pattern is framework-agnostic (any Node/edge handler that base64-decodes `state` without HMAC).
- Happy to refine the `pattern-not-inside` shape if maintainers prefer a tighter scope (e.g. JWT-verify or named state-validation helpers).

References:
- RFC 6749 §10.12 — CSRF in OAuth `state`: https://datatracker.ietf.org/doc/html/rfc6749#section-10.12
- Writeup with real-world impact analysis: https://shippedwithbugs.com/writeups/outrank

### DIFF
--- a/javascript/lang/security/oauth-state-unsigned-base64.js
+++ b/javascript/lang/security/oauth-state-unsigned-base64.js
@@ -1,0 +1,33 @@
+const crypto = require("node:crypto");
+
+// --- vulnerable ---
+
+function callbackBad(req) {
+  const state = req.query.state;
+  // ruleid: oauth-state-unsigned-base64
+  const decoded = JSON.parse(Buffer.from(state, "base64").toString("utf-8"));
+  return decoded.organization_id;
+}
+
+function callbackBadBrowser(req) {
+  const state = req.query.state;
+  // ruleid: oauth-state-unsigned-base64
+  const decoded = JSON.parse(atob(state));
+  return decoded.organization_id;
+}
+
+// --- safe: HMAC-verified before deserialization ---
+
+const SECRET = process.env.OAUTH_STATE_SECRET;
+
+function callbackGood(req) {
+  const raw = req.query.state;
+  const [body, sig] = raw.split(".");
+  const expected = crypto.createHmac("sha256", SECRET).update(body).digest("hex");
+  if (!crypto.timingSafeEqual(Buffer.from(sig), Buffer.from(expected))) {
+    return null;
+  }
+  // ok: oauth-state-unsigned-base64
+  const decoded = JSON.parse(Buffer.from(body, "base64").toString("utf-8"));
+  return decoded.organization_id;
+}
--- a/javascript/lang/security/oauth-state-unsigned-base64.js
+++ b/javascript/lang/security/oauth-state-unsigned-base64.js
@@ -1,33 +1,76 @@
 const crypto = require("node:crypto");
 
-// --- vulnerable ---
+// --- vulnerable: state from query, deserialized without verification ---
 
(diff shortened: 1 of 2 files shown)


nedu-m:
Thanks for the Codex review — both critiques were correct. Pushed a revision that converts the rule to taint mode:

**Addressing P2 (scope too broad):** Sources are now constrained to OAuth `state` inputs only — `req.query.state`, `req.body.state`, bracket-access variants, and `searchParams.get("state")` for Next.js / edge-runtime route handlers. Generic base64-JSON parsing of unrelated values (config blobs, etc.) no longer triggers.

**Addressing P1 (suppression too loose):** Sanitizers (`timingSafeEqual`, `jwt.verify`) now only suppress findings when the *same tainted state value* flows through them. The `callbackBadWithUnrelatedTimingSafeEqual` test case in the fixtures specifically exercises this: an unrelated `timingSafeEqual` on an API key in the same function no longer suppresses the OAuth state finding.

Expanded fixtures:
- Direct source (`req.query.state`) and indirect (via local variable)
- Next.js App Router route-handler form (`searchParams.get`)
- Browser/edge `atob` variant
- False-positive guard for unrelated `timingSafeEqual` (must fire)
- False-positive guard for generic base64-JSON parsing (must not fire)
- Safe case with HMAC verification on the state itself
- Safe case with JWT-encoded state verified via `jwt.verify`

Happy to refine further — e.g. extending sources to cover other common state-passing patterns, or tightening sanitizer matching with `pattern-inside`. Let me know.

PR #3860, opened 2026-06-02
Poetry and Ruby cooldown

gbennett-squarespace:
Continuation from:
https://github.com/semgrep/semgrep-rules/pull/3805

### Poetry

Adds Poetry cooldown.  Poetry doesn't seem to look at `pyproject.toml`. I haven't actually fully tested that, but according to their documentation and looking where the setting is populated when you use `poetry config`, `poetry.toml` and `config.toml` are the spots to check.

This one's pretty simple as Poetry only allows an `int`, no `7 day` shenanigans.

### Ruby Bundler

Pretty straight forward.  Bundler doesn't have a separate field for setting exceptions, so have to go the Renovate route of allowing an in place exception.  

### Renovate

Also updated the Renovate rule to allow to set to `false` to disable minimum package age. Renovate doesn't have another field to set exclusions for packages, so the only mechanism is on the rule itself.

### DIFF
--- a/package_managers/poetry/poetry-missing-solver-min-release-age.test.toml
+++ b/package_managers/poetry/poetry-missing-solver-min-release-age.test.toml
@@ -1,0 +1,63 @@
+---
+# ruleid: poetry-missing-solver-min-release-age
+[solver]
+
+---
+# ruleid: poetry-missing-solver-min-release-age
+[solver]
+min-release-age = 0
+
+---
+# ruleid: poetry-missing-solver-min-release-age
+[solver]
+min-release-age = 3
+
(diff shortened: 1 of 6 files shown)


0xDC0DE (approved):
Looks good to me!

gbennett-squarespace:
@0xDC0DE Added Ruby if you want to re-review.  Blame Leif for this one. :)

0xDC0DE (approved):

0xDC0DE:
Sorry for the delayed review! This looks good to me! Thanks for adding @gbennett-squarespace !

PR #3871, opened 2026-06-05
Add llm-output-to-sql taint rule for Python

KezoSec:
## What this adds
A new taint-mode rule, `llm-output-to-sql-python`, that flags LLM API
response data flowing into raw SQL execution (CWE-89). It's the SQL
sibling of the existing `ai/ai-best-practices/llm-output-to-exec` rule.

## Why
Text-to-SQL and agentic LLM features increasingly let model output reach
database queries directly. Combined with prompt injection or RAG context
poisoning (OWASP LLM01), a manipulated response can inject SQL. There's a
rule for LLM-output-to-exec, but no SQL equivalent yet.

## How it works
- Mode: taint
- Sources: OpenAI / Anthropic / Gemini chat & completion calls, plus
  LangChain `.invoke(...)`
- Sinks: `cursor.execute` / `executemany`, `sqlalchemy.text(...)`,
  Django `.objects.raw(...)`
- Parameterized queries are not flagged: the focus metavariable is the
  query string, so passing tainted data as a bound parameter stays clean.

## Tests
`semgrep --test` passes — 4 true-positive cases fire, 2 safe cases
(parameterized + hardcoded) stay clean.

## Open questions for maintainers
1. Placement: I put it next to `llm-output-to-exec` under
   `ai/ai-best-practices/`. Happy to move it if that isn't the right
   home for community-contributed rules.
2. The `.invoke(...)` source is intentionally broad and could over-fire
   outside LLM contexts, glad to tighten it to specific client classes
   if you'd prefer.

Open to feedback.

### DIFF
--- a/ai/ai-best-practices/llm-output-to-sql/llm-output-to-sql-python.py
+++ b/ai/ai-best-practices/llm-output-to-sql/llm-output-to-sql-python.py
@@ -1,0 +1,54 @@
+from openai import OpenAI
+from langchain_openai import ChatOpenAI
+import sqlalchemy
+
+client = OpenAI()
+llm = ChatOpenAI()
+
+
+def vulnerable_fstring(cursor):
+    r = client.chat.completions.create(
+        model="gpt-4", messages=[{"role": "user", "content": "table name"}]
+    )
+    table = r.choices[0].message.content
+    # ruleid: llm-output-to-sql-python
+    cursor.execute(f"SELECT * FROM {table}")
+
+
+def vulnerable_sqlalchemy(conn):
+    r = client.chat.completions.create(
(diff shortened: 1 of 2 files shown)


KezoSec:
Quick note on the bot findings above: they're all on the rule's test
file (llm-output-to-sql-python.py). Those flagged lines are the
intentional vulnerable examples the new rule is meant to detect, each
is marked with a `# ruleid:` comment, so they're fixtures, not real
code.

I deliberately didn't add `# nosemgrep` to silence them, since that
would also suppress the rule's own test assertions on those same lines.
Happy to handle the fixtures however fits your convention for rule test
files (e.g. excluding them from the dogfood scan), just let me know
what you'd prefer.

PR #3983, opened 2026-06-28
fix(php): extend symfony-non-literal-redirect with RedirectResponse and setTargetUrl patterns

SkajeDev:
## Summary

Extends the existing `php.symfony.security.audit.symfony-non-literal-redirect` rule 
with missing sink patterns that were not covered by the original rule.

## Problem

The original rule only covered `$this->redirect()` but missed two common patterns 
for open redirect vulnerabilities in Symfony:

- `new RedirectResponse($url)` — direct instantiation without controller helper
- `$response->setTargetUrl($url)` — modifying redirect target after instantiation

Both patterns are equally dangerous and widely used in Symfony applications.

### symfony-non-literal-redirect.yaml
- Added `new RedirectResponse(...)` as a sink with literal string exclusion
- Added `$RESPONSE->setTargetUrl(...)` as a sink with literal string exclusion
- Added `pattern-inside` guards for Symfony imports to reduce FP rate
- Added `vulnerability_class: Open Redirect` to metadata

### symfony-non-literal-redirect.php
- Added `test3` — `ruleid` for `new RedirectResponse($url)`
- Added `test4` — `ruleid` for `setTargetUrl($url)`
- Added `okTest4` — `ok` for `new RedirectResponse("literal")`
- Added `okTest5` — `ok` for `setTargetUrl("literal")`
- Preserved all original test cases unchanged

## Test Plan

- [x] semgrep --validate passes
- [x] semgrep --test passes (3/3 tests)

### DIFF
--- a/csharp/dotnet/security/audit/xpath-injection.cs
+++ b/csharp/dotnet/security/audit/xpath-injection.cs
@@ -1,4 +1,4 @@
-public List<Knowledge> Search(string input)
+public List<Knowledge> Search1(string input)
 {
     List<Knowledge> searchResult = new List<Knowledge>();
     var webRoot = _env.WebRootPath;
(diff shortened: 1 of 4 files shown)


SkajeDev:
This branch is based on a separate PR #3982 (csharp xpath-injection improvements). 
Only the symfony-related files are part of this PR.

PR #4022, opened 2026-07-27
fix(c): describe the real hazard in insecure-use-strtok-fn

### LINKED ISSUE #3772: insecure-use-strtok-fn: rule message describes normal behavior instead of actual security risks
**Describe the bug**

The current message of rule [insecure-use-strtok-fn](https://github.com/MarkLee131/semgrep-rules/blob/develop/c/lang/security/insecure-use-strtok-fn.yaml) states:

  > Avoid using 'strtok()'. This function directly modifies the first argument buffer, permanently erasing the delimiter character.

  Modifying the input buffer is `strtok`'s normal, documented behavior, not a security issue. Callers who use `strtok` expect this. The message does not mention the actual security risks:

  1. **Non-thread-safe internal static state**: concurrent or nested `strtok()` calls on different strings silently corrupt each
  other's parsing position
  2. **No `strtok_s()` recommendation**: only suggests `strtok_r()` (POSIX), missing the C11 Annex K alternative

  This makes triage harder: reviewers must independently determine that the stated risk is not a real concern, rather than being guided
  toward the actual threat.

  **To Reproduce**

  1. Run `semgrep --config auto` on any C codebase using `strtok()`
  2. Read the finding message
  3. Note the message describes intended behavior, not a vulnerability

  **Expected behavior**

  The message should describe the real security risks (thread safety, nested call corruption) and recommend both `strtok_r()` and
  `strtok_s()`.

  **Priority**

  - [] P2: annoying but not blocking me

  **Additional context**

Found while scanning [Vim](https://github.com/vim/vim) (~400k LOC). All 21 strtok findings were false positives in a single-threaded application. The inaccurate message added unnecessary triage burden.

Eljees:
Fixes #3772

## Problem

The `insecure-use-strtok-fn` message describes `strtok()`'s documented behavior as if it were the hazard:

> Avoid using 'strtok()'. This function directly modifies the first argument buffer, permanently erasing the delimiter character. Use 'strtok_r()' instead.

Modifying the input in place is exactly what `strtok()` is specified to do, and it is not why the rule recommends `strtok_r()`. As #3772 points out, the message leaves a reader unsure what the actual risk is — and it doesn't explain what makes the suggested replacement better, since `strtok_r()` modifies the buffer in the same way.

## Change

Message only — the pattern, metadata and severity are untouched.

The rewrite leads with the real reason `strtok_r()` exists: `strtok()` keeps the parser position in a hidden static buffer, so it is neither reentrant nor thread-safe and interleaved or concurrent calls clobber each other's state. In-place modification is kept as a secondary note, and `strtok_s()` (C11 Annex K) is mentioned next to `strtok_r()` for portability.

No test changes are needed since matching behavior is unchanged.

### DIFF
--- a/c/lang/security/insecure-use-strtok-fn.yaml
+++ b/c/lang/security/insecure-use-strtok-fn.yaml
@@ -2,9 +2,12 @@ rules:
 - id: insecure-use-strtok-fn
   pattern: strtok(...)
   message: >-
-    Avoid using 'strtok()'. This function directly modifies the first argument buffer,
-    permanently erasing the
-    delimiter character. Use 'strtok_r()' instead.
+    Avoid 'strtok()': it is not reentrant or thread-safe, since it keeps the parser
+    position in a hidden static buffer, so interleaved or concurrent calls (even on
+    unrelated strings) clobber each other's state. It also modifies its input in
+    place, overwriting each delimiter with a NUL byte. Use the reentrant
+    'strtok_r()' (POSIX) or 'strtok_s()' (C11 Annex K), which keep the scan state in
+    a caller-provided context.
   metadata:
     cwe:
     - 'CWE-676: Use of Potentially Dangerous Function'
--- a/c/lang/security/insecure-use-strtok-fn.yaml
+++ b/c/lang/security/insecure-use-strtok-fn.yaml
@@ -2,12 +2,12 @@ rules:
 - id: insecure-use-strtok-fn
   pattern: strtok(...)
   message: >-
-    Avoid 'strtok()': it is not reentrant or thread-safe, since it keeps the parser
-    position in a hidden static buffer, so interleaved or concurrent calls (even on
-    unrelated strings) clobber each other's state. It also modifies its input in
-    place, overwriting each delimiter with a NUL byte. Use the reentrant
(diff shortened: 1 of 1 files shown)


Eljees:
All three are right, and on a PR that is specifically about message accuracy they matter. Fixed in d2e8e14:

- "hidden static buffer" -> "hidden static state". The saved scan position is a pointer into the caller's string, not a separate buffer; the input is the only buffer, and it is modified in place.
- "not reentrant or thread-safe" -> "not reentrant, and the C standard does not require it to be thread-safe". Implementations that keep the state in TLS are safe across threads; the portable problem is that nothing guarantees it, while interleaved calls in one thread still share state.
- "each delimiter" -> "each token-ending delimiter". Leading, trailing and repeated delimiters are skipped without being modified - tokenizing `",,a,,b,,"` leaves several commas intact.

`semgrep --test` on `c/lang/security/insecure-use-strtok-fn.c` still passes and `--validate` reports no errors; the pattern is unchanged, so matching behaviour is identical.

Eljees:
Ping on this one and #4023, #4024, #4025, #4026, #4027, #4029 - all open since 27-28 July without a review. They are independent of each other, so they can be handled or closed one by one.

Issue #3831, opened 2026-04-28
FP and FNs for documentbuilderfactory-disallow-doctype-decl-missing:


sarr424:
Hi, I found some pattern-coverage gaps in the Java XXE rule documentbuilderfactory-disallow-doctype-decl-missing.

Case 1 (FP): pattern-sanitizers only recognize setFeature(...). Needs setAttribute support:
- patterns:
    - pattern: $FACTORY.setAttribute($ATTR, "")
    - metavariable-regex:
        metavariable: $ATTR
        regex: ACCESS_EXTERNAL_DTD|ACCESS_EXTERNAL_SCHEMA
Cases 2 & 3 (FN): pattern-sources requires $FACTORY = DocumentBuilderFactory.newInstance(). 
Needs a supplementary source for chained calls:
- pattern: DocumentBuilderFactory.newInstance().newDocumentBuilder()
- pattern: $X = DocumentBuilderFactory.newInstance().newDocumentBuilder()

```java
package example;
import javax.xml.parsers.DocumentBuilderFactory;
import javax.xml.parsers.DocumentBuilder;
import javax.xml.parsers.ParserConfigurationException;
import javax.xml.XMLConstants;
import org.xml.sax.SAXException;
import java.io.IOException;
class AllCases {
    // ── Case 1: FP · setAttribute-based defense ──
    // Should NOT be flagged (ACCESS_EXTERNAL_DTD="" blocks external entities).
    // Semgrep flags it anyway → FALSE POSITIVE
    public void case1_setAttribute() throws ParserConfigurationException {
        DocumentBuilderFactory dbf = DocumentBuilderFactory.newInstance();
        dbf.setFeature(XMLConstants.FEATURE_SECURE_PROCESSING, true);
        dbf.setAttribute(XMLConstants.ACCESS_EXTERNAL_DTD, "");
        dbf.setAttribute(XMLConstants.ACCESS_EXTERNAL_SCHEMA, "");
        dbf.newDocumentBuilder();
    }
    // ── Case 2: FN · full chained call ──
    // Should be flagged (no XXE protection at all).
    // Semgrep misses it → FALSE NEGATIVE
    public void case2_fullChained() throws ParserConfigurationException, SAXException, IOException {
        DocumentBuilderFactory.newInstance().newDocumentBuilder().parse("system.xml");
    }
    // ── Case 3: FN · inline chained assignment ──
    // Should be flagged (no XXE protection at all).
[...]

PR #4024, opened 2026-07-28
fix(java): treat ACCESS_EXTERNAL_DTD as an XXE sanitizer

Eljees:
Addresses case 1 of #3831 (the false positive).

## Problem

The rule only recognizes `setFeature(...)` as a defense, so a factory hardened through the JAXP properties is still reported:

```java
DocumentBuilderFactory dbf = DocumentBuilderFactory.newInstance();
dbf.setFeature(XMLConstants.FEATURE_SECURE_PROCESSING, true);
dbf.setAttribute(XMLConstants.ACCESS_EXTERNAL_DTD, "");
dbf.setAttribute(XMLConstants.ACCESS_EXTERNAL_SCHEMA, "");
dbf.newDocumentBuilder();   // flagged before this change
```

Setting `ACCESS_EXTERNAL_DTD` to the empty string stops the parser from resolving external DTDs and entities — the same protection the rule already accepts from the `external-general-entities` + `external-parameter-entities` features, and the mitigation the OWASP XXE cheat sheet recommends when DOCTYPE declarations cannot be disabled outright.

## Change

One sanitizer alternative:

```yaml
- patterns:
    - pattern: $FACTORY.setAttribute($ATTR, "")
    - metavariable-regex:
        metavariable: $ATTR
        regex: ^(.*\.)?ACCESS_EXTERNAL_DTD$
```

Two notes on scope:

- The issue suggests matching `ACCESS_EXTERNAL_DTD|ACCESS_EXTERNAL_SCHEMA`. I deliberately match only `ACCESS_EXTERNAL_DTD`: `ACCESS_EXTERNAL_SCHEMA` restricts schema resolution and does not stop external entity resolution, so accepting it on its own would turn the false positive into a false negative. Code that sets both (as above) is still sanitized, via the DTD property.
- The regex is anchored so it works whether the constant is written qualified (`XMLConstants.ACCESS_EXTERNAL_DTD`) or imported statically.

Cases 2 and 3 of the issue (false negatives on chained `DocumentBuilderFactory.newInstance().newDocumentBuilder()`) are not addressed here — new sources interact with the rule's autofix, so they are better handled in a separate change.

## Tests

Added the reproducer as an `ok` case in the test file, mirrored into `.fixed.java`.

```
semgrep --test --config .../documentbuilderfactory-disallow-doctype-decl-missing.yaml .../documentbuilderfactory-disallow-doctype-decl-missing.java
1/1: ✓ All tests passed
1/1: ✓ All fix tests passed
```

### DIFF
--- a/java/lang/security/audit/xxe/documentbuilderfactory-disallow-doctype-decl-missing.fixed.java
+++ b/java/lang/security/audit/xxe/documentbuilderfactory-disallow-doctype-decl-missing.fixed.java
@@ -3,6 +3,7 @@ package example;
 import javax.xml.parsers.DocumentBuilderFactory;
 import javax.xml.parsers.DocumentBuilder;
 import javax.xml.parsers.ParserConfigurationException;
+import javax.xml.XMLConstants;
 
 
 class GoodDocumentBuilderFactory {
@@ -166,4 +167,15 @@ class GoodDocumentBuilderFactoryCtr3 {
         dbf.setFeature("http://xml.org/sax/features/external-parameter-entities", false);
     }
 
+}
+
+class GoodDocumentBuilderFactoryAccessExternalDtd {
+    public void blockExternalDtd() throws ParserConfigurationException {
+        DocumentBuilderFactory dbf = DocumentBuilderFactory.newInstance();
+        dbf.setFeature(XMLConstants.FEATURE_SECURE_PROCESSING, true);
+        dbf.setAttribute(XMLConstants.ACCESS_EXTERNAL_DTD, "");
+        dbf.setAttribute(XMLConstants.ACCESS_EXTERNAL_SCHEMA, "");
+        //ok:documentbuilderfactory-disallow-doctype-decl-missing
+        dbf.newDocumentBuilder();
+    }
 }
--- a/java/lang/security/audit/xxe/documentbuilderfactory-disallow-doctype-decl-missing.fixed.java
+++ b/java/lang/security/audit/xxe/documentbuilderfactory-disallow-doctype-decl-missing.fixed.java
@@ -179,3 +179,13 @@ class GoodDocumentBuilderFactoryAccessExternalDtd {
         dbf.newDocumentBuilder();
     }
 }
+
+class BadDocumentBuilderFactoryUnrelatedConstant {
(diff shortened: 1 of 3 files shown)


Eljees:
Good catch on the first point — fixed in 102f6ec.

The regex accepted any expression ending in `ACCESS_EXTERNAL_DTD`, so an unrelated application constant with the same trailing name would have suppressed the finding. It's now anchored to the JAXP symbol:

```yaml
regex: ^(javax\.xml\.)?XMLConstants\.ACCESS_EXTERNAL_DTD$|^ACCESS_EXTERNAL_DTD$
```

(the bare alternative keeps static imports working). I added a test case using a look-alike `MyOwnConfig.ACCESS_EXTERNAL_DTD` that must still be reported — it fails against the previous regex and passes now.

On the other three: they're all real, and all pre-existing shapes of the same limitation rather than something this PR introduces.

- **Static initializers** and **hardening helpers**: the rule already misses `setFeature` in both of those positions today; the sanitizer I added inherits that scope, it doesn't narrow it. Worth fixing, but for `setFeature` and `setAttribute` together rather than only for the property I'm adding.
- **Re-taint when the property is restored to a permissive value**: agreed this is unsound, but it applies equally to every existing sanitizer in this rule (`setFeature(..., true)` followed by `setFeature(..., false)` has the same problem). That needs `by-side-effect` re-tainting on the whole rule, which is a bigger change than this PR should carry.

Happy to open a follow-up for the taint-scope items if that's useful.

Eljees:
Ping — open since 28 July, no review yet. Checks are green on `102f6ec`.

This addresses case 1 of #3831: a factory hardened through the JAXP properties (`setAttribute(XMLConstants.ACCESS_EXTERNAL_DTD, "")`) is still reported, because the rule only recognises `setFeature(...)` as a defence. The regex is anchored to the JAXP symbol, so a look-alike application constant with the same trailing name cannot suppress a finding — there is a test using `MyOwnConfig.ACCESS_EXTERNAL_DTD` that must still be reported.

The other three points raised in review are pre-existing shapes of the same limitation rather than something this PR introduces; detail is in the thread. Happy to rework it, or to close it if the change is not wanted in this shape.

PR #4037, opened 2026-08-08
fix(docker-compose): support files without version

### LINKED ISSUE #4035: p/docker-compose rules fail when obsolete version attribute is omitted
**Describe the bug**

The Semgrep Community Docker Compose rules appear to depend on the obsolete top-level Docker Compose `version` attribute.

When `version` is omitted from a valid modern Docker Compose file, at least the following rules fail to report findings:

```text
yaml.docker-compose.security.no-new-privileges.no-new-privileges

yaml.docker-compose.security.writable-filesystem-service.writable-filesystem-service
```

When the same Compose file is given:

```yaml
version: '3.8'
```

both findings are reported as expected.

Modern Docker Compose treats the top-level `version` attribute as obsolete and ignores it, so valid Compose files without `version` should still be analyzed correctly.

**To Reproduce**

Use this Docker Compose file without a top-level `version` attribute:

```yaml
services:
  www:
    container_name: www
    image: docker.io/webpwnized/dragonfly:www

    ports:
      - 127.0.0.1:80:80
      - 127.0.0.1:8888:80

    networks:
      - httpnet

networks:
  httpnet:
```

This service intentionally omits both:

```yaml
security_opt:
  - no-new-privileges:true
```

and:

```yaml
read_only: true
```

Run the Docker Compose Community ruleset:

```bash
docker run --rm \
  -v "$(pwd)/.build:/mnt/src:ro" \
  docker.io/semgrep/semgrep:latest \
  semgrep scan \
    --config p/docker-compose \
    /mnt/src \
    --error
```

Semgrep reports no findings for the missing `no-new-privileges` and writable root filesystem conditions.

The individual `no-new-privileges` rule also fails:

```bash
docker run --rm \
  -v "$(pwd)/.build:/mnt/src:ro" \
  docker.io/semgrep/semgrep:latest \
  semgrep scan \
    --config r/yaml.docker-compose.security.no-new-privileges.no-new-privileges \
    /mnt/src \
    --error
```

Result:

```text
Findings: 0
Rules run: 1
Targets scanned: 1
Parsed lines: ~100.0%
```

Now add only this line at the top of the same file:

```yaml
version: '3.8'
```

The file becomes:

```yaml
version: '3.8'

services:
  www:
    container_name: www
    image: docker.io/webpwnized/dragonfly:www

    ports:
      - 127.0.0.1:80:80
      - 127.0.0.1:8888:80

    networks:
      - httpnet

networks:
  httpnet:
```

Run the same Semgrep scans again.

The missing `no-new-privileges` rule now fires:

```text
yaml.docker-compose.security.no-new-privileges.no-new-privileges

Service 'www' allows for privilege escalation via setuid or setgid binaries.
Add 'no-new-privileges:true' in 'security_opt' to prevent this.
```

The writable root filesystem rule also fires:

```text
yaml.docker-compose.security.writable-filesystem-service.writable-filesystem-service

Service 'www' is running with a writable root filesystem. This may allow malicious applications to
download and run additional payloads, or modify container files. If an application inside a
container has to save something temporarily consider using a tmpfs. Add 'read_only: true' to this
service to prevent this.
```

The only relevant change is the addition of:

```yaml
version: '3.8'
```

**Expected behavior**

These Docker Compose security rules should detect the vulnerable service configuration regardless of whether the obsolete top-level `version` attribute is present.

Both of these should be equivalent scan targets:

```yaml
version: '3.8'

services:
  www:
    image: docker.io/webpwnized/dragonfly:www
```

and:

```yaml
services:
  www:
    image: docker.io/webpwnized/dragonfly:www
```

The second form is valid modern Docker Compose syntax and should not cause these security rules to stop matching.

**Screenshots**

I can provide screenshots of both scan results if useful.

**What is the priority of the bug to you?**

- [ ] P0: blocking your adoption of Semgrep or workflow
- [x] P1: important to fix or quite annoying
- [ ] P2: regular bug that should get fixed

**Environment**

Semgrep is being run using the official Docker image:

```text
docker.io/semgrep/semgrep:latest
```

The source is mounted read-only into the scanner container:

```bash
-v "$(pwd)/.build:/mnt/src:ro"
```

The affected rules are:

```text
yaml.docker-compose.security.no-new-privileges.no-new-privileges

yaml.docker-compose.security.writable-filesystem-service.writable-filesystem-service
```

The rules are exercised through:

```text
p/docker-compose
```

and, for isolated testing:

```text
r/yaml.docker-compose.security.no-new-privileges.no-new-privileges
```

Semgrep version can be obtained with:

```bash
docker run --rm docker.io/semgrep/semgrep:latest semgrep --version
```

**Use case**

I am evaluating Semgrep Community Edition for a large-scale GitOps-based SAST architecture where Docker Compose and application security rules would run automatically in CI/CD pipelines.

The intent is to run broad Community rulesets in audit mode and promote selected high-confidence rules into blocking controls.

A structural dependency on the obsolete Docker Compose `version` attribute creates a significant false-negative risk because modern Compose files commonly omit that field. Fixing these rules would allow the same security controls to work correctly against current Docker Compose syntax without requiring organizations to retain an obsolete configuration attribute solely for SAST compatibility.

fyrsta7:
## Link to an issue

Fixes #4035.

## Summary

Modern Docker Compose files can omit the obsolete top-level `version` field, but six Docker Compose security rules currently require it before they can match. This causes false negatives for valid versionless Compose files.

This change:

- removes the `version` requirement from all six affected Docker Compose security rules;
- keeps findings scoped to `services`;
- excludes GitHub Actions service containers under `jobs.<job>.services`;
- adds regression coverage for versionless Compose files and GitHub Actions workflows.

The GitHub Actions exclusion preserves the false-positive protection introduced in #1241 while allowing current Compose syntax.

Two touched rules also update their discouraged `CWE-284` metadata to `CWE-250: Execution with Unnecessary Privileges`, which matches related container-confinement rules and satisfies the current changed-file lint.

## Testing

- `semgrep validate yaml/docker-compose/security`
- `semgrep test yaml/docker-compose/security` (6/6 passed)
- repository Semgrep rule and changed-file strict lints (23 rules, 0 findings)
- registry metadata schema validation
- pre-commit hooks for all changed files
- historical Semgrep Docker validation and tests with `returntocorp/semgrep:1.163.0`
- `git diff --check`


fyrsta7:
Hi maintainers, just a gentle follow-up on this PR. The CLA and Semgrep Cloud checks are passing, and the PR is currently conflict-free. When you have a chance, I would appreciate a review. Thank you!

